PR #13, opened 2015-08-01
Update eo.md

planteome-user:
Introducing a deliberate mistake for demonstration purposes


planteome-user:
Oops, sorry about that

PR #35, opened 2015-08-12
Update ms.md

germa:
Updated

### DIFF
--- a/ontology/ms.md
+++ b/ontology/ms.md
@@ -1,16 +1,23 @@
 ---
 layout: ontology_detail
 id: ms
+label: Mass spectrometry ontology
+title: Mass spectrometry ontology
 contact: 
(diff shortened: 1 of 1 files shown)


cmungall:
Sorry @germa, I made some changes en masse to the md files and that seems to have resulted in a conflict.  It's perhaps easiest if you resolve this, but if this is too difficult I will just close this PR and manually add your changes myself.

One thing that I will have to change is the product id, for our system this has to be `mi` because that's the id space that you use. You can have products of the form `ms.FMT` or `ms/anything/you/like/here`.

Alternatively you can register 'psims' as the ontology ID... but we strongly recommend that this is the same as the ID space http://obofoundry.github.io/id-policy.html

cmungall:
@germa, have a look at the current version and see if you need to make changes

PR #40, opened 2015-08-12
Update cmo.md

rgduser:


cmungall:
btw, that's a neat rgd logo you have there - shall I add it to your ontologies?

rgduser:
Thank you!  Yes, it would be great if you could add that to our ontologies.

PR #58, opened 2015-08-14
Update cmo.md

rgduser:
Add homepage, publication and browser links

### DIFF
--- a/ontology/cmo.md
+++ b/ontology/cmo.md
@@ -6,7 +6,9 @@ contact:
   label: Mary Shimoyama
 description: Morphological and physiological measurement records generated from clinical and model organism research and health programs.
 domain: clinical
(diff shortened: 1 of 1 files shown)


cmungall:
I'm not totally sure why the check failed. Perhaps some invisible character is being inserted?

If you can't figure it out on your end then I'll just merge and fix

mcourtot:
Seems to be a tab after "publications:"? Maria had similar issue earlier today (sorry on mobile or would look myself)
found character '\t' that cannot start any token
  in "<unicode string>", line 18, column 14:
    publications:

cmungall:
I added `util/tidy-whitespace.pl` to fix the tab issue (unfortunately I don't think this can be hooked into the web based editing).

PR #61, opened 2015-08-14
Update rs.md

rgduser:
Add homepage link, publication title and browser link


cmungall:
Redid your changes, let us know if you have questions

The multiple homepage thing is a bit odd at the moment. The idea is that every ontology has exactly one `homepage` - this corresponds to foaf:homepage which is declared as a functional property, i.e. can't have >1

The `page` key should really be a list, so that you can list as many secondary pages as you like

We may review all this in the future

PR #69, opened 2015-08-18
Adding stato logo

agbeltran:

PR #70, opened 2015-08-18
Update stato.md #69

agbeltran:


cmungall:
Ahhhhh that's not too good! Perhaps we should be dynamically resizing but I'd like to get advice from a CSS person before mucking around at that level... for now do you think you can generate a smaller image?

agbeltran:
Indeed, it is too big. I will look into re-sizing the image. Thanks! 
BTW, two contacts are not supported?

cmungall:
AFAICR a single (human) contact was mandated by the OBO Foundry a while ago. But I don't see why we shouldn't support secondary contacts, help mailing lists, etc - you can open a ticket for this

Issue #73, opened 2015-08-18
Support for multiple contacts - which include ORCIDs


agbeltran:
I think it would be useful to allow for multiple contacts for an ontology, including multiple people or mailing list(s).

cmungall:
+1 for single primary contact (continue to use 'contact' field) and any number of additional contacts

agbeltran:
It seems that #33 would require this, as well as STATO.

zhengj2007:
If I remember correctly, based on OBO Foundry policy, only one person is allowed for ontology contact. Sorry, I cannot find where it is documented.

scheuerm:
I agree with Chris M., +1 for single primary contact (continue to use 'contact' field) and any number of additional contacts

mbrochhausen:
I also agree with Chris': +1 for single primary contact (continue to use 'contact' field) and any number of additional contacts

cstoeckert:
+1 for single primary contact (continue to use 'contact' field) and any number of additional contacts.
The 'locus of authority' principle (#11) names a Contact Person as "a single person who is responsible for communications...". So need to be unambiguous about which contact serves that role.

batchelorc:
I'm with Mathias on this. And Chris by the looks of things.

mcourtot:
I'm on the fence. I could live with the 1 primary contact and additional, but I'm wondering if the need for additional doesn't stem from the old association with "contacts = credit". If we have a single contact and for example a mailing-list, do we really need additional contacts as well? Perhaps if the rational for the request could be expanded a bit more, maybe with use cases explaining why this is desirable?

The big argument in favour of having a single contact is to have one responsible person - having many contacts has the potential to open the door to no one taking care of the request ultimately (thinking someone else did)

agbeltran:
The request wasn't based on contacts = credit (as this would imply having potentially many contacts) but considering that there might be a few (two or three possibly) responsible people that could reply to requests rather than a single person. Thanks.

cmungall:
In general I would encourage ontologies to encourage the use of the tracker for most requests. For users who don't want to do this, don't want to get a login, etc, then a help email or helpdesk type URL is a good idea (we should have a proper field for this - what RDF property would this be?). The contact is primarily for OBO central to communicate with the ontology maintainers (maybe now it's easier to introspect this from github, but this may not always be the optimal route). Maybe we show it too prominently on the site

cmungall:
[...]

PR #97, opened 2015-09-04
Update mamo.md

niko-rodrigue:
Updating the contact information

### DIFF
--- a/ontology/mamo.md
+++ b/ontology/mamo.md
@@ -2,8 +2,8 @@
 layout: ontology_detail
 id: mamo
 contact:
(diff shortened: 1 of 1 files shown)


cmungall:
Thanks for the update!

Just a note to say that the intent for the contact is really meant to be a single human individual. However, we recognize that a help email address is generally the more useful metadata field, and a number of people have filled in contact with a help email, so this is all fine for now, but when this ticket is resolved, we may want to make global changes and ensure that distinct help and primary contact emails are distinguished: #73

PR #109, opened 2015-09-18
Update zfa.md

cerivs:
Point source at GitHub not Google code preversion


cmungall:
Looks like yaml wants a space after the `:`. I will merge then fix. Thanks @cerivs !

PR #133, opened 2015-10-02
Update omp.md

dsiegele:
Added missing information to metadata and updated source_url.

### DIFF
--- a/ontology/omp.md
+++ b/ontology/omp.md
@@ -1,11 +1,26 @@
 ---
 layout: ontology_detail
 id: omp
+contact:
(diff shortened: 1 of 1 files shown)


cmungall:
huh that's an uninformative error... will look later

cmungall:
OK, it looks like you provided 3 blank products. I will merge and fix

Also, for the obo2owl method, you need to provide a URL that resolves to (raw) obo, i will fix

cmungall:
do you want to do another PR for a license or shall I add?

PR #134, opened 2015-10-05
Update chebi.md

jannahastings:
I have updated the metadata to fix the capitalisation of the ontology title, and to make myself not the primary point of contact any longer, as after 9 years I am leaving the EBI.


cmungall:
All the best in what's next... Cambridge PhD?

jannahastings:
Yes! I’m going to become an expert on the metabolism of C. elegans… :-)

PR #160, opened 2015-11-03
Sort ontology list: Foundry, Library, Obsolete

jamesaoverton:
- update extract-metadata.py to sort YAML objects
- update ontology_table.html to change sort and add classes
- add custom.css to highlight Foundry and Obsolete rows
- modify default.html to use custom.css

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -7,20 +7,227 @@ imgurl: /images
 author:
   name: OBO Technical WG
 ontologies:
-- contact: {email: David Blackburn, label: david.c.blackburn@gmail.com}
-  homepage: http://github.com/seger/aao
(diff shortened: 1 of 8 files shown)


jamesaoverton:
For your consideration. We discussed this on the OFOC call this morning as a light-weight improvement to the table of ontologies.

jamesaoverton:
Note: in `extract-metadata.py` I'm just checking for the presence of `is_obsolete` and `in_foundry_order` keys, not the values. The former is boolean and the latter is always "1", but might be intended to have other values. I can make this smarter.

cmungall:
nice!

+1

remember to remake the yml after merging the PR

PR #184, opened 2015-11-30
Added link to purl entry in github. Fixes #181

### LINKED ISSUE #181: Add Edit purl metadata button on ontology_detail pages
Analogous to the edit main metadata button.

Adding the button should be trivial. But also need to integrate with main documentation

cmungall:
@jamesaoverton look OK?

### DIFF
--- a/_layouts/ontology_detail.html
+++ b/_layouts/ontology_detail.html
@@ -261,6 +261,15 @@ layout: default
                     class="btn btn-default">
               Edit</button>
           </a>
+          <a href="https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/{{page.id}}.yml">
+
+            <button type="button" 
+                    data-toggle="tooltip" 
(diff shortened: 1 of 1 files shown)


cmungall:
This is how it looks

screen shot 2015-11-30 at 12 39 02 pm

jamesaoverton:
The link is correct. In my `simpler` branch the filename uses the IDSPACE casing, but I'll worry about that later.

Maybe the button label should be plural: "PURLs"?

cmungall:
Yes, PURLs is better

Hmm, could get confusing if one is IDSPACE case

jamesaoverton:
The registry YAML files should have a field for the exact case-sensitive IDSPACE. Is this it? `preferredPrefix: FBdv`

In any case, I don't need that field right now, and there's an `idspace` field in the PURL YAML files that I've manually checked for most of our ontologies.

PR #187, opened 2015-12-10
Remove use of legacy URL from principle11 PR

cmungall:


rlwalls2008:
I did these in the wrong order (186 before 187). @cmungall, can you tell me if the final results are correct?

PR #207, opened 2016-01-28
add URL for tracker and creative commons license

dsiegele:
I added the URL for the OMP ontology tracker and a CC-BY license to the OMP listing on obofoundry.org


dsiegele:
Thank you!

PR #209, opened 2016-01-31
Update ico.md

yongqunh:
Update the ico.md information.
Yongqun Oliver He

### DIFF
--- a/ontology/ico.md
+++ b/ontology/ico.md
@@ -2,10 +2,33 @@
 layout: ontology_detail
 id: ico
 title: Informed Consent Ontology
+description: An ontology of clinical informed consents, which is aligned with OBO Foundry library ontologies including BFO, OBI, IAO, and OGMS.
+license:
+  url: http://creativecommons.org/licenses/by/3.0/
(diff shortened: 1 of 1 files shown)


yongqunh:
Cool! This was my first .md file update request. Thanks, Chris!
Oliver

cmungall:
I recommend short descriptions, for display in the main table, and keeping the description focused on the __content__ rather than __methods__.

E.g. rather than

> An ontology of clinical informed consents, which is aligned with OBO Foundry library ontologies including BFO, OBI, IAO, and OGMS

> An ontology clinical informed consent

Or

> An ontology that represents processes and information pertaining to obtaining informed consent in medical investigations

IMHO details about what ontologies are "aligned" (I think you mean imports a subset of) should go in other fields, or in the detailed formatted markdown information. Here you can provide more meaningful information: the ontology inherits from classes in BFO like A and B, it is complementary to OGMS in that it represents C, D and E etc.

For my proposed guidelines see #8

yongqunh:
Short description is indeed much better. Thanks!

PR #222, opened 2016-02-23
Reformat page for principle 001 open, see #5

jamesaoverton:
The principle pages were copied directly from the Google Code site.
Google Code used wiki syntax, but GitHib uses Markdown syntax.
This and other pages need to be updated to use Markdown.
I change the syntax and made minor style changes:

- soft-wrap all lines
- reformat lists as Markdown
- fix escaping in code blocks
- add/reformat links
- consistent capitalization and punctuation in headings

### DIFF
--- a/principles/fp-001-open.md
+++ b/principles/fp-001-open.md
@@ -1,168 +1,99 @@
 ---
 layout: principle
 id: fp-001-open
-title: open
+title: Open
 ---
 
-Principle Name
---------------
+## Principle Name
 
 open
 
-Summary
--------
+## Summary
 
-The ontology MUST be openly available to be used by all without any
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I can merge this myself, but I want to run it past the Editorial Working Group first.

Once concern that I have is the "Original Formulation". I linked to the history on the old wiki, but there might be a better option.

jamesaoverton:
Correction: the principle pages did not come from Google Code. They came from the old wiki, e.g. http://wiki.obofoundry.org/wiki/index.php/FP_001_open

cmungall:
+1

cmungall:
yes, converted from wiki via pandoc

PR #225, opened 2016-02-25
Reformat page for principle 002 format, see #5

jamesaoverton:
The principle pages were copied directly from toehold MediaWiki site.
MediaWiki syntax differs from GitHib's Markdown syntax.
I have revised the syntax and made minor style changes,
without making any substantive changes to the content:

- soft-wrap all lines
- reformat lists as Markdown
- add/reformat links
- consistent capitalization in headings

### DIFF
--- a/principles/fp-002-format.md
+++ b/principles/fp-002-format.md
@@ -1,90 +1,65 @@
 ---
 layout: principle
 id: fp-002-format
-title: format
+title: Format
 ---
 
-Principle Name
---------------
+## Principle name
 
 common format
 
-Summary
--------
+## Summary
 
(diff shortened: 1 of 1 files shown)


nataled:
+1

cmungall:
+1

We should probably recommend the release versions of OWL only use RDF/XML or RDF/turtle as that is consumable by a much wider range of tooling (I love some manchester internally). We should also have a policy on OWL/Zip when available. But all this can come later

jamesaoverton:
Thanks @cmungall. I filed that idea as issue #226.

alanruttenberg:
+1 modulo discussion in #226

Issue #227, opened 2016-03-02
Autogenerate copyable citation


mellybelly:
We will have something like this on the Citation information page:
The order of authors is not important, though you can use the primary contact listed on the OBO site as the first author as a default. Here is an example citation:

Mungall, C., Haendel, M., Dahdul, W., Ibrahim, N., Segerdell, E., Blackburn, D., Comte, A., Niknejad, A., and Decechi, A. (2016), Uberon Ontology, http://purl.obolibrary.org/obo/uberon/releases/2016-01-26/uberon.owl

It would be great if we could autogenerate a copyable citation of this nature, that includes all the creators in the file, with the "owner" as the first author, and the latest version IRI. 

Even better might be to have a doi with a landing page for such citations, these could be generated as part of the release process and appended in the above citation format.

@cmungall @jamesaoverton ideas for how to implement?

nlharris:
Is this done?
#220 seems to address most of this.
Ontology pages such as Uberon (http://www.obofoundry.org/ontology/uberon.html) have publications under "cite".
There's also a citation info page at http://www.obofoundry.org/docs/Citation.html
What else is needed?

nlharris:
can this be closed?

nlharris:
Closing; please open new issue(s) if needed

PR #228, opened 2016-03-02
Add citation policy

jamesaoverton:
Add a document on OBO citation policy to the site.


jamesaoverton:
This pull request is just for discussion, and is not yet ready to merge.

mellybelly:
+1 

See also #227 
I think if we had this implemented, it would be much easier for users to use best practices for citing the version IRI.

mellybelly:
Also, I think the publication citations should be in addition to the version IRI citation.

cmungall:
+1

jamesaoverton:
We talked about this on the call, and people (including me) found the JATS reference confusing.

@mellybelly: I did my best to incorporate your suggestion, but I'm still not convinced that we should recommend citing *both* the recommended publication and the ontology IRI for the project.

@alanruttenberg: I agree that dc:creator isn't universal. Do you have a better alternative in the short-term?

mellybelly:
On Mar 8, 2016, at 12:35 PM, James A. Overton <notifications@github.com<mailto:notifications@github.com>> wrote:


We talked about this on the call, and people (including me) found the JATS reference confusing.

@mellybelly<https://github.com/mellybelly>: I did my best to incorporate your suggestion, but I'm still not convinced that we should recommend citing both the recommended publication and the ontology IRI for the project.

disagree -  the former is for people, the latter for reproducibility/computational referencing

@alanruttenberg<https://github.com/alanruttenberg>: I agree that dc:creator isn't universal. Do you have a better alternative in the short-term?

—
Reply to this email directly or view it on GitHub<https://github.com/OBOFoundry/OBOFoundry.github.io/pull/228#issuecomment-193882528>.

Dr. Melissa Haendel

Associate Professor
Ontology Development Group, OHSU Library
www.ohsu.edu/library/ontology<http://www.ohsu.edu/library/ontology>
Department of Medical Informatics and Clinical Epidemiology
Oregon Health & Science University
haendel@ohsu.edu<mailto:haendel@ohsu.edu>
skype: melissa.haendel
503-407-5970
Appointments: Shanez De Silva desilva@ohsu.edu<mailto:desilva@ohsu.edu>

cmungall:
Should we have a new ticket for requirements for reproducibility? Even with versionIRIs reproducibility is not guaranteed as none of our infrastructure is properly funded. And as we move away from monolithic ontologies, there are particular challenges with imports here.

The best way to guarantee reproducibility is to deposit a version of the ontology in amongst the other files used in the analysis into a repository with  better guarantees on longevity, e.g. dryad/figshare. Even with larger ontologies there is no onerous space requirements here. I would recommend here that the imports closure is merged to guarantee immutability of the load operation.

There are also challenges in interfacing with modern reproducible bioinformatics workflows. For example, iPython/Jupyter is very popular and good for reproducibility of workflows. Yet the options for using ontologies in python are shockingly bad; I would bet that the majority of the time any ontology is used in a any kind of computational workflow that is a target for reproducibility, there is an ad-hoc obo parser used. Many ontologies do not even have versioned obo files.

mellybelly:
Perhaps its time to work on our principle here:
http://www.obofoundry.org/principles/fp-004-versioning.html

bpeters42:
Discussed on 3/22. Lynn, Jie, Chris, Simon, James + Bjoern. The consensus was that there is a stronger desire to keep this simple. That will ensure that people read and try to follow it. So decision to post as is at this point. We can add more complexity later.

jamesaoverton:
Reopening

PR #236, opened 2016-03-23
Update ms.md

germa:
Now points to the new location of psi-ms.obo at https://github.com/HUPO-PSI/psi-ms-CV/blob/master/psi-ms.obo


cmungall:
Can you update this to use the raw URL for the source? This one returns HTML

germa:
Now points to the raw URL

PR #259, opened 2016-05-17
Update to fix file downloads for VTO.

balhoff:


cmungall:
The redirects tag does nothing, you can make a separate PR here https://github.com/OBOFoundry/purl.obolibrary.org/

balhoff:
Ah right, thanks.

PR #268, opened 2016-06-07
Changed ChEBI issue tracker url

muthuvenkat:
I have changed the ChEBI tracker url. Please update it. 

Thanks,
Venkat

### DIFF
--- a/ontology/chebi.md
+++ b/ontology/chebi.md
@@ -24,7 +24,7 @@ browsers:
 products:
   - id: chebi.owl
   - id: chebi.obo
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I created this PR to update this PURL <http://purl.obolibrary.org/obo/chebi/tracker> to point to the new tracker. Perhaps you don't use that PURL, but I think it make sense to keep it updated -- please confirm.

muthuvenkat:
Thanks @jamesaoverton, please do point it to the new tracker (here @EBI-CHEBI).

PR #286, opened 2016-07-13
Update naming conventions

### LINKED ISSUE #282: Naming Conventions Link broken
At
http://www.obofoundry.org/principles/fp-012-naming-conventions.html 
the link
http://www.obofoundry.org/principles/Naming
does not resolve.
Where can naming conventions be found?

cmungall:
See #282 for details. We have lost the text that was originally linked (But in general the principles should be self-contained and not rely on external links).

My modifications here attempt to include a concise summary of naming conventions, attempting to pre-empt the most common problems (e.g. CamelCase)

Fixes #282

### DIFF
--- a/principles/fp-012-naming-conventions.md
+++ b/principles/fp-012-naming-conventions.md
@@ -12,9 +12,17 @@ naming conventions
 Details
 -------
 
-See: [Naming](Naming "wikilink")
+For full details, see this paper: <http://www.biomedcentral.com/1471-2105/10/125>
 
-And this paper: <http://www.biomedcentral.com/1471-2105/10/125>
+Briefly, some important things to remember:
+
+ * use rdfs:label for the primary label
+ * include exactly one rdfs:label for every declared entity (e.g. class, property)
+ * write labels, synonyms, etc as if writing in plain English text. ie use spaces to separate words, only capitalize proper names (e.g. Parkinson disease). Do not use CamelCase, do_not_use_underscores
+ * spell out abbreviations. Abbreviations can be included as a separate property.
+ * make the primary labels to be as unambiguous as possible. Remember, your ontology may be used in a different context than that for which it was originally intended. Remember also of course that the label should be unambiguous without looking at parent terms
+ * labels should be unique within an ontology
+ * use the IAO property 'obo foundry unique label' [http://purl.obolibrary.org/obo/IAO_0000589](http://purl.obolibrary.org/obo/IAO_0000589) to declare a pan-OBO unique label if required
 
 Date Accepted
 -------------


mellybelly:
+1
thought we had some review and text of this before but having hard time finding it.

mellybelly:
found it: https://docs.google.com/document/d/1nCCO88uOWgWp3z5AwyapcTP0T_C2ZmdPi7cWiO9_aDM/edit#heading=h.lmriq5ksqxfh


Principle 12: Naming conventions

New wording draft

Current details: http://obofoundry.org/wiki/index.php/Naming and http://www.biomedcentral.com/1471-2105/10/125.

Sez Melissa: Naming is complicated as per the referenced documentation. Do you think it is possible to provide a high level summary here so that we can use this for the purposes of review? Basically a short list of requirements and consistency checks.
We agreed.

Name: Naming conventions

Summary and Purpose: the names for classes in an ontology should be intelligible to scientists and be amenable to natural language processing.

Implementation:

[todo: put MUST and SHOULD in caps.]

Here is a list of the headline points from the above links.

Names of classes should as far as possible reflect usage in ordinary scientific body text. They are not titles. This means that words within them should only be capitalized if they are proper names (for example Golgi) multiword names should have the words separated by spaces (not underscores or CamelCase) non-deprecated class names should not contain information relevant only to the ontology, for example “protocol (definition incomplete)”. Reasons for allowing this for deprecated terms include to avoid clashes between classes which would otherwise have the same name, and to provide a readily-usable audit trail.

Short, commonly-used names should appear as they do in ordinary running text in the normal order of words and should not contain commas, hence “hoses for the use of firemen”, not “hoses, firemen, for the use of”. Commas may be used where they are part of, say, chemical names, or in long complicated terms (more than four or five words) such as in the Gene Ontology.

names should be singular: “plant cell” rather than “plant cells”. If you are referring to a collective or aggregate, then say so. Plurals may be added as synonyms.

Names of classes should be short and memorable, for example “wall of esophagus” rather than “the wall of the esophagus”.

Non-obvious abbreviations should be spelt out, for example “nuclear magnetic resonance” rather than “NMR”. DNA and laser are clear exceptions to this. Abbreviations should be added as synonyms.

MERGE WITH PROPOSED ‘UNIQUE LABEL’ PRINCIPLE: Names should be univocal within a given ontology, for example “nuclear magnetic resonance magnet” rather than “magnet”. 

Discussion point: what about names of relations?

cstoeckert:
We do have a new draft in the "Proposed new wording for OBO Foundry Principles" doc. I think that Chris has captured most of the points there with one exception:
names should be singular: “plant cell” rather than “plant cells”. If you are referring to a collective or aggregate, then say so. Plurals may be added as synonyms.
Otherwise, I think it is good for now.

mellybelly:
@cstoeckert is that the doc i linked above? I might not be aware of a new one as out of the loop

cstoeckert:
@mellybelly yes that's the doc.

rlwalls2008:
+1

nataled:
+1
The text that currently shows is from the original principles page, which is why it references that external doc. As I recall, that doc was rather long and had a lot of information in it that might have touched upon a number of principles. Shame to lose it. The text and link provided above represents an EWG work-in-progress. I'm fine with using it as is for now, or with using the quick list provided by Chris. Either way it will be updated in the near future.

cmungall:
Thanks for the link to the gdoc @mellybelly - I forgot about the existence of those.

As an aside, that gdoc has a revision history going back to 2013. I suggest we move towards very granular PRs directly on the md file, maximizing transparency, rapidly incorporating easy suggestions, having a voting process etc

mellybelly:
+1. I'll try to make some incremental progress based on earlier discussion and the ones I was assigned

On Jul 13, 2016, at 9:01 AM, Chris Mungall <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for the link to the gdoc @mellybelly<https://github.com/mellybelly> - I forgot about the existence of those.

As an aside, that gdoc has a revision history going back to 2013. I suggest we move towards very granular PRs directly on the md file, maximizing transparency, rapidly incorporating easy suggestions, having a voting process etc

PR #291, opened 2016-07-25
new foodon config file

Public-Health-Bioinformatics:
Should be sufficient for first launch

### DIFF
--- a/ontology/foodon.md
+++ b/ontology/foodon.md
@@ -1,0 +1,29 @@
+---
+layout: ontology_detail
+id: foodon
+contact:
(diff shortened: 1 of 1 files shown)


cmungall:
A few things we can correct later:

 * did you mean to assign a homepage to the .obo product specifically?
 * I would recommend something like "Food Ontology" for title

Public-Health-Bioinformatics:
ok, title change would be good; and I was copying ENVO form, but didn't clue in about YAML entry for subordinate home page entry - yes I will drop that soon.

PR #309, opened 2016-09-04
Update hp.md

drseb:
- Update email
 - Add monarch browser
 - Image link was broken

P.S.: Can we have multiple contact persons? If so, can you add Peter with peter.robinson@jax.org ?


cmungall:
See #73 for multiple contacts - this was for OBO but for other purposes you may want to show two or three people equally on the website

PR #349, opened 2016-12-02
Add Oral Health and Disease Ontology

alanruttenberg:
Around for a longish time but never registered.


jamesaoverton:
I'll merge the corresponding PURLs once this registry entry is merged: https://github.com/OBOFoundry/purl.obolibrary.org/pull/271

PR #354, opened 2016-12-08
fix bootstrap path, regenerate to include ohd

alanruttenberg:
_includes/themes/bootstrap-3/default.html  used path

{{ ASSET_PATH }}/bootstrap/css/bootstrap.min.css
(and 2 others)

but that was moved (by someone else). I changed it to current location

{{ ASSET_PATH }}/bootstrap-3/bootstrap/css/bootstrap.min.css
(and 2 others)

Not sure how it could have worked on obofoundry.org


cmungall:
@alanruttenberg any chance you could do another PR only committing the html and not the derived files? Otherwise I'll go ahead and do this manually when I get the chance

PR #363, opened 2017-01-04
spd now on github

cmungall:
@martinjramirez - type `+1` if you agree to these changes

cc @balhoff

### DIFF
--- a/ontology/spd.md
+++ b/ontology/spd.md
@@ -14,10 +14,10 @@ taxon:
   label: spiders
 title: Spider Ontology
 build:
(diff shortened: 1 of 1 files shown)


cmungall:
Note that we don't need to modify
https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/spd.yml

As you are using the central OBO berkeley build (see obofoundry.org/faq/what-is-the-build-field.html for more info).

At some stage you may wish to switch to producing the OWL yourself using ROBOT as part of the release, but no rush.

martinjramirez:
+1

cmungall:
Correct - this configuration drives the nightly job that deposits things at the standard purl. You can find out more: obofoundry.org/faq/what-is-the-build-field.html

Issue #341, opened 2016-12-01
Fix ontoneo


cmungall:
This is invalid:

```
build:
  source_url: 
  method: owl2obo
```

cc @FernandaFarinelli @mcourtot 

Either delete the entire build blob OR put the source URL in there.

You also need a PURL entry

FernandaFarinelli:
Hi Melanie @mcourtot ,

Just let me know if I need to do something.

cmungall:
Is this fixed yet?

FernandaFarinelli:
@mcourtot 
I submitted the Update to ontoneo.md (#369) right now.

cmungall:
See my comments in the PR, not sure what happened here

mcourtot:
Fixed.

PR #369, opened 2017-01-05
Update ontoneo.md

FernandaFarinelli:
Update ontoneo.md according to issue Fix ontoneo (#341).

### DIFF
--- a/ontology/ontoneo.md
+++ b/ontology/ontoneo.md
@@ -15,8 +15,7 @@ mailing_list: http://groups.google.com/group/ontoneo-discuss
 products: 
   - id: ontoneo.owl
 build:
-  source_url: 
-  method: owl2obo
+  source_url: http://purl.obolibrary.org/obo/ontoneo/ontoneo.owl
 tracker: https://github.com/ontoneo-project/Ontoneo/issues
 browsers:
   - label: BioPortal


cmungall:
-1

This isn't right.

Sorry, something is messed up here, I don't think it's your fault.

First of all we don't have an entry for ontoneo in our purl registry:

https://github.com/OBOFoundry/purl.obolibrary.org/tree/master/config

This still works:
http://purl.obolibrary.org/obo/ontoneo/ontoneo.owl
...but it seems to be relying on an entry in the deprecated OCLC system.

We should just get you an entry in https://github.com/OBOFoundry/purl.obolibrary.org/tree/master/config

You could make a PR to do this, but we don't have good docs for this yet.

After that is done, the correct purl which is actually http://purl.obolibrary.org/obo/ontoneo.owl will redirect directly to your raw github file. You will no longer have need of the build tag

mcourtot:
Yes - just back at office and still need to create PURLs for ontoneo and pdro. sorry about delay.

cmungall:
purl PR here: https://github.com/OBOFoundry/purl.obolibrary.org/pull/280

cmungall:
Looks like the upstream issue is fixed

PR #370, opened 2017-01-08
Fix source of so

cmungall:


mcourtot:
Hi @cmungall - does this mean I can add the SO import in the GO makefile again?

cmungall:
I think the SO ontologyIRI has been correct for a while now..

PR #375, opened 2017-02-03
Update wbphenotype.md

Gschindelman:
I received a request from Chris Mungall to add a reference to the wbphenotype page
http://obofoundry.org/ontology/wbphenotype.html
(Chris may be reading this edit request)

I'm not sure I did it correctly...
I forked and then I looked at other pages to see how they added a reference and came up with adding the following to the following lines:
21 publications:
22  - id: http://www.ncbi.nlm.nih.gov/pubmed/?term=21261995
23  title: "Worm Phenotype Ontology: integrating phenotype data within and beyond the C. elegans community."
How this was correct.
Thanks
Gary


cmungall:
Almost, the title should line up with the id for the publication, I fixed

PR #384, opened 2017-02-09
Changing tracker for ddanat

cmungall:
See https://github.com/dictyBase/migration-data/issues/5

### DIFF
--- a/ontology/ddanat.md
+++ b/ontology/ddanat.md
@@ -19,7 +19,7 @@ build:
   source_url: https://raw.githubusercontent.com/dictyBase/migration-data/master/ontologies/dicty_anatomy.obo
   method: obo2owl
   insert_ontology_id: true
-tracker: http://sourceforge.net/tracker/?group_id=76834&atid=974664
+tracker: https://github.com/dictyBase/migration-data/issues
 ---
 
 A structured controlled vocabulary of the anatomy of the slime-mould <i>Dictyostelium discoideum</i>.


pfey03:
@cmungall I just updated ddanat. Added one term. Updated so rarely, I hope it's fine to just live where we have all our ontologies 
https://github.com/dictyBase/migration-data/blob/master/ontologies/dicty_anatomy.obo

balhoff:
@pfey03 do you maintain ddanat elsewhere and copy to that repo? Just wondering because I wanted to point out my pull request there: https://github.com/dictyBase/migration-data/pull/12

pfey03:
@balhoff no, you are correct and I only maintain it in dictyBase/migration/data.  

A couple years ago Chris asked to put it in another repo but that didn't suit us since this ontology so rarely changes and we need it in the current repo for our new emerging database.

balhoff:
Okay, thanks—if you can merge that pull request it would be an improvement for GO. But if it is a problem, let me know and maybe we can discuss.

pfey03:
I merged the pull request. 
Also, about adding xrefs to ddanat: Will you update the ontology so it doesn't cause problems in GO, and maybe also in UPHENO? If you do please let me know.

pfey03:
Ah, I saw already through the merge, the xrefs. I'll update my own ontology with that and on our old database to be everywhere the same, thanks!

Issue #389, opened 2017-02-15
Add Travis check for products


mcourtot:
https://github.com/EBISPOT/OLS/issues/137

ICO doesn't show up in OLS, despite being listed in the OBO registry, as the "Products" list is empty. We could maybe add a check for this?

mcourtot:
See also https://github.com/OBOFoundry/OBOFoundry.github.io/issues/391 for list of ontologies that fail as they are missing products

nlharris:
this is over 3 years old; can it be closed?

matentzn:
@jamesaoverton the schema says that the products field is [optional](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/1118e1f2951cfb363dd1a42e4645ac0d6196a073/util/schema/registry_schema.json#L328);

@cmungall do you still maintain that it should be possible that an ontology has no products? 

You closed Melanie's [attempt](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/390) to at QC to that end in 2017 - not sure why..

In any case, if there is no response from anyone after 4 weeks, I would say we can close this issue. The way it is formulated, it is clearly an OLS problem - OLS should not be dependent on the products field in the OBO metadata. I don't think its its right now.

jamesaoverton:
A brand new ontology might not have any products, I guess.

We need the Travis checks to be consistently green. So in general we've only marked fields as required in the schema if we've already gone through and ensured that all registry entries are passing validation. That can be a lot of work, although probably not bad in this specific case since products is simple with an obvious, safe default. So I think we just haven't tried to verify all the products.

cmungall:
I think my point was that in 2017 we needed to retrospectively populate the
products field before we implemented a hard fail check?

An ontology should have minimally one product

nlharris:
just a note that we may be switching to using github actions rather than travis

PR #390, opened 2017-02-17
Update extract-metadata.py

mcourtot:
as per https://github.com/OBOFoundry/OBOFoundry.github.io/issues/389

Trying to validate that there is a product tag in the configuration files


cmungall:
It looks like the change does what is intended, but I'm not sure we want this. For various reasons we have ontologies that have no product yet.

Why not have this be produced as a separate report?

cmungall:
closing for now but can revisit if necessary

PR #402, opened 2017-03-08
Update nbo.md

leechuck:
Switches source to OWL file (with MIREOT); updated links to new github group

### DIFF
--- a/ontology/nbo.md
+++ b/ontology/nbo.md
@@ -5,7 +5,7 @@ contact:
   email: g.gkoutos@gmail.com
   label: George Gkoutos
 domain: behavior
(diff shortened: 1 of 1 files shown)


cmungall:
Just FYI, you are switching from the 'git' method which does a git clone and then copies the top level nbo obo/owl files to ontologies.berkeleybop.org, to an owl2obo conversion. It doesn't matter that much but you probably want to either (1) take care of the entire build in github yourself and use the git method (or simply have your purls go direct to github) or (2) remove the .obo from your repo since this may get out of sync with the central owl2obo one.

leechuck:
Thank you. We will retire the .obo file shortly.
https://github.com/obo-behavior/behavior-ontology/issues/41

PR #403, opened 2017-03-10
Patch 2 foodon_core.owl and fix to /imports/ path

Public-Health-Bioinformatics:
At moment /imports/ path is wrong, preventing ontology from being loaded in OLS etc.

### DIFF
--- a/ontology/foodon.md
+++ b/ontology/foodon.md
@@ -13,16 +13,17 @@ license:
   label: CC-BY
 products:
   - id: foodon.owl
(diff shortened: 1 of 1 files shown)


cmungall:
These changes look fine but they don't seem to have anything to do with fixing an imports path.

Public-Health-Bioinformatics:
Thanks.  I was editing both the purl file and this one and thinking they were part of the same pull request, forgetting that the purl change goes off to another repo.

Issue #344, opened 2016-12-01
Proposal: obsolete ontologies with dead purls


cmungall:
(also sent to obo-discuss)

If you have an ontology registered with OBO, then there should be a working purl for your ontology. There should be at least one purl for the canonical OWL edition of your ontology, which will look like this:

http://purl.obolibrary.org/obo/envo.owl
Many ontologies also make other editions/products available, for example an obo format edition, or editions for subsets

The list of such artefacts can be seen on your OBO page under 'products', e.g.

http://obofoundry.org/ontology/envo.html
Note that if you edit your ontology in OBO-Edit, and have not yet figured out how to make OWL releases, don't worry. We provide a central build as a free service: http://obofoundry.org/faq/what-is-the-build-field.html (but really you should be looking to make your own releases, we have good tutorials on this).

__IMPORTANT__

A number of PURLs are no longer functional. In some cases this is due to an ontology being abandoned on googlecode (see the archives of this list for repeated messages to migrate from google code). Or in some cases a sourceforge URL that no longer works.

Non-functional PURLs cause a great deal of problems with groups who consume the OBO library ontologies programatically. Previously we've tried to be proactive and chase up and help fix these issues, but some responsibility rests with the ontology producer.

I propose that if you have previously submitted an ontology to OBO, and the PURL for it no longer works (or never worked) then it is considered abandoned, and is obsoleted after an initial warning period.

If you produce an ontology, please take a look at the list here:
https://github.com/OBOFoundry/OBOFoundry.github.io/issues/257#issuecomment-264242016

If your ontology is on the list and you are committed to OBO, don't panic! -- we'll work with you to fix this. Please comment either in this ticket (or the linked ticket that pertains to your ontology).

However, if the ontology PURL doesn't resolve and no one cares about it, I propose it be obsoleted, after which it will cease to cause problems for downstream consumers.

caripark:
For VT, we've done everything in OBO--I'd be interested in getting some help with this issue. Thanks!

cmungall:
Great - I sent you an email earlier, we can follow up there

ramonawalls:
At the OBOF operations meeting on 3/21, we agreed that this can be implemented as policy and promoted on website, on the page that describes PURL policies.

See also issue #257

nlharris:
Does this still need to be completed? Chris noted in the merge of #407 

> I'm merging as this is a good incremental improvement, leaving the parent ticket open

[...]

PR #407, opened 2017-03-21

ramonawalls:
I added a new section on PURL maintenance, resolves issue #344. I also updated the request example to one that is on GitHub. Not sure is there is an issue for that.

@cmungall or @jamesaoverton please check and commit if you approve.

### DIFF
--- a/docs/Policy_for_OBO_namespace_and_associated_PURL_requests.md
+++ b/docs/Policy_for_OBO_namespace_and_associated_PURL_requests.md
@@ -4,7 +4,9 @@ id: Policy_for_OBO_namespace_and_associated_PURL_requests
 title: Policy_for_OBO_namespace_and_associated_PURL_requests
 ---
 
-**This policy has been formally adopted on April 19th 2013. Do not edit this page without consulting with the OBO Technical group.** Comments to [obo-discuss@lists.sourceforge.net](mailto:obo-discuss@lists.sourceforge.net)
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Thanks Ramona. This is a good first draft but I'd like to get more precise.

On the call today we focused on the "product" PURLs (the PURLs for a project's primary OWL and OBO files) which are particularly important. Breakage with these has caused problems downstream with portals. We have a number of tracker items about these problems, and we want to have a clear policy for stepping in to make changes. The products are listed in the registry entries, and configured in the PURL entries (in the other repository).

By "doesn't work" I think we mean that the product PURL does not resolve to a valid OWL or OBO file. By "obsolete" I think we mean removing these PURLs from our lists of all products, so that downstream consumers don't automatically try to fetch them. @cmungall might have something else in mind. I'd prefer to be more specific about the steps we'll take to contact maintainers, and how long we'll wait for a response.

Sorry to be pedantic, but getting clear now might save us headaches later.

cmungall:
We need to document our obsoletion better, but it's this:

 - we add a is_obsolete tag to the yaml-md in *this* repo. This makes the ontology appear in strikeout on the page
 - we don't touch the purls, they continue to redirect as before, whether it's to a 404, invalid syntax, etc

cmungall:
I'm merging as this is a good incremental improvement, leaving the parent ticket open

PR #408, opened 2017-03-24
Update FMA homepage

jamesaoverton:
The FMA home page seems to have moved to http://si.washington.edu/projects/fma


jamesaoverton:
This seems clear-cut to me. Any objections?

PR #421, opened 2017-04-26
Adds license to CDAO metadata

hlapp:
The license (more specifically, the CC0 public domain waiver) under which CDAO is released actually always has been explicitly stated in the OWL file, but apparently the metadata extractor doesn't pick that up.


cmungall:
Correct. We have a SPARQL query for comparing the metadata in the yaml with what is in the published ontology. Currently we just manually check this for differences in an ad-hoc way. We need better and smarter syncing here. But for now PRs such as this are most welcome, thx.

PR #436, opened 2017-06-13
Create mi.md

noedelta:
Updates psi-mi.obo source location


cmungall:
Apologies for the delay!

PR #439, opened 2017-06-30
add cc-by-sa logo to registry

andrewsu:


cmungall:
This is a derived file, the geno entry derives from here: https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/ontology/geno.md

But the actual mapping of a license to a logo is controlled by util/extract-metadata.py, I'll add this here

Issue #285, opened 2016-07-12
Create guidelines for OBO maintainers who want to be included in Wikidata


cmungall:
Most OBOs are CC-BY, Wikidata requires CC-0. Some ontologies have apparently granted Wikidata permission to redistribute part or all of their ontology.

We want to make sure this is streamlined with a common process for everyone. Not clear to me how this should be done, ideas welcome, add below.

mcourtot:
CC-by but attribution chosen is by PURL - as long as wikidata uses the PURL (perhaps replacing the URL formate that currently links directly to AmiGO) I think it should be ok.

elviram:
I agree that this needs to be written down somewhere. It would make everything much clearer and we could avoid getting into lengthy discussion later down the line, or even have to remove data, if something were to happen.

andrewsu:
re: @mcourtot's comment, it's true that Wikidata probably generally satisfies the attribution requirement for CC-BY.  But Wikidata itself is CC0, so if you grant Wikidata permission to use your data, then you also grant downstream users those same CC0 terms.  So someone who downloads your ontology via Wikidata would *not* be required to attribute in any way.

cmungall:
On 13 Jul 2016, at 8:39, Andrew Su wrote:

> re: @mcourtot's comment, it's true that Wikidata probably generally 
> satisfies the attribution requirement for CC-BY.  But Wikidata itself 
> is CC0, so if you grant Wikidata permission to use your data, then you 
> also grant downstream users those same CC0 terms.  So someone who 
> downloads your ontology via Wikidata would *not* be required to 
> attribute in any way.

Additionally, this points to an assumption in the OBO license that 
assumes a PURL for every unit of attributable work. What if I want to 
produce an ontology that is purely axioms on an existing ontology? This 
could be logical axioms (e.g. providing equivalence axioms for DO, or 
logical definitions for CHEBI roles) or annotation axioms (e.g. a 
translation to another language)? If I want to ensure attribution I 
would have to add PURLs for every axiom, which can be a high overhead.

NuriaQueralt:
[...]

PR #444, opened 2017-07-06
modify license choice language

andrewsu:
Given that the choice between CC0 and CC-BY is a nuanced one with many pros and cons on both sides of the issue, I offer three suggestions for this document:

1. linking to https://github.com/OBOFoundry/OBOFoundry.github.io/issues/285 where many issues are explicitly discussed
2. removing the explicit recommendation of CC-BY
3. adding a request for attribution in all cases regardless of license (following [this pattern](http://www.dancohen.org/2013/11/26/cc0-by/))

I of course understand that this policy is ultimately under the purview of the Editorial WG, but I've formulated this as a pull request just to propose something specific.

### DIFF
--- a/principles/fp-001-open.md
+++ b/principles/fp-001-open.md
@@ -32,7 +32,7 @@ OBO Foundry ontologies are resources for the entire biological and biomedical co
 
 3. If an individual term is reused without change to the definition, the original term IRI should be used. If the definition of a term (either text or logical) is changed, the original term IRI should not be reused. Suggestions for changes or improvements to term definitions should be submitted to the appropriate ontology issue tracker.
 
-4. Regardless of which license an ontology uses, we strongly request and recommend that any reuse of an ontology attributes the source in accordance with scientific norms.
+4. Regardless of which license an ontology uses, we strongly request and recommend that any reuse of an ontology attributes the source in accordance with scientific norms and the [OBO Citation and Attribution Policy](http://www.obofoundry.org/docs/Citation.html).
 
 ### Remark
 
--- a/principles/fp-001-open.md
+++ b/principles/fp-001-open.md
@@ -24,6 +24,8 @@ OBO Foundry ontologies are resources for the entire biological and biomedical co
 
 2. We recommend that new ontologies carefully consider the choice of license. For example, [Creative Commons](https://creativecommons.org/) licenses like [CC0](https://creativecommons.org/publicdomain/zero/1.0/) and [CC-BY](https://creativecommons.org/licenses/by/4.0/) are popular open and understandable licenses. Many pros and cons of each license are laid out in [this discussion](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/285).
 
+Regardless of which license is chosen, it should be clearly stated in the ontology file.
+
 ### For ontology re-use
 
 1. The original source of an ontology must be credited according to the terms specified in the comment annotation of the ontology.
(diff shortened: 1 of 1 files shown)


cmungall:
+1

Other members of @OBOFoundry/obo-admin should vote

bpeters42:
This proposal of 'ethical attribution' writes out much better than I could what my gut feeling was we should be doing. I enthusiastically support this.

lschriml:
Enthusiastic 
+1 
Lschriml

mellybelly:
I think this is a good idea. Linking to the ticket for discussion of the issues, but I think the primary link should be to our citation policy http://www.obofoundry.org/docs/Citation.html, we may want to update.

zhengj2007:
+1

If we made the changes, we'd also announce it on obo-discuss list.

kltm:
I would agree with the link to the tracker conversation and request of attribution regardless of license choice.

However, while I agree that there should be no explicit recommendation of CC-BY, I think it would be a very good idea to help people who are new licensing questions with a bit more direction in open licensing, and suggest CC and CC licenses as a starting point (as they do have very good discussion and are quite robust).

Possibly (overly strong) verbiage along the lines of:

We recommend that new ontologies carefully consider the choice of license. For example, [Creative Commons](https://creativecommons.org/) licenses like [CC0]() and [CC-BY]() are popular open and understandable licenses. Many pros and cons of each license are laid out in [this discussion](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/285).

I think directed discussion may be important to keep people on board and concerned about the issue--while the thread is interesting, I think it may be a bit much for people who are new to thinking about licenses and leading to a sensible default may be valuable.

andrewsu:
FWIW, @kltm's suggested revision gets a strong thumbsup from me. I [added]( https://github.com/andrewsu/OBOFoundry.github.io/commit/cbabca526c5750dec2581075b374b0f9f3983895) the change to the pull request...

andrewsu:
and realized I overlooked @mellybelly's suggestion, which I think is also great...  [added](https://github.com/andrewsu/OBOFoundry.github.io/commit/408ee8729e4a2b8d238e4af57459ee7492eb3c0a)...

jamesaoverton:
+1 for the latest version

mcourtot:
+ 1 with suggestion of adding something along the lines of "The licence should be clearly stated in the ontology file."

andrewsu:
I agree with @mcourtot and am happy to add something to the PR. But I'm wondering what the right protocol here is. With every additional change, do people need to re-vote?  (Is there a formal voting process by the Editorial WG?) It seems there is broad support for the changes, so would like to know the best way to wrap this up...

cmungall:
For additions that are informative and clearly improve/clarify I don't think we need to re-vote. For normative changes such as @mcourtot's in general it would be good to at least give people a chance to veto.

However until we can get a sensible policy in our CONTRIBUTING.md (hopefully in a month or so) let's proceed pragmatically. Given the  https://tools.ietf.org/html/rfc2119 definition of SHOULD the proposed addition is non-binding and I don't foresee any objections, so let's do it.

ramonawalls:
+1

------------------------------------------------------
Ramona L. Walls, Ph.D.
Senior Scientific Analyst, CyVerse, University of Arizona
Research Associate, Bio5 Institute, University of Arizona

andrewsu:
done!  added a line that reflects @mcourtot's suggestion...

lschriml:
+1 for the latest version

principles/fp-001-open.md
```diff
@@ -22,7 +22,9 @@ OBO Foundry ontologies are resources for the entire biological and biomedical co
 
     OR released into the public domain under [CC0](https://creativecommons.org/publicdomain/zero/1.0/) (which is not a license). Note that CC0 specifies that the creators of an ontology waive, to the extent that they legally can be, all rights and place the ontology in the public domain. It does not prevent them from requesting that the ontology be properly credited and cited, but prevents any legal recourse if it is not credited.
```
hlapp: _to the extent that they legally can be_: typo, should remove _be_

PR #465, opened 2017-08-22
Removed broken link to obof in principle1

### LINKED ISSUE #460: A broken link on page http://www.obofoundry.org/principles/fp-001-open.html
Dears,
There is a broken link on the page related to the first principle 
 (http://www.obofoundry.org/principles/fp-001-open.html) under obo.files item 3. 
The link to the comment is broken. 
By the way, the same item suggests the readers see examples below which is not clear (at least to me)

The broken link on the page is:
http://www.geneontology.org/GO.format.obo-1_4.shtml#S.2.1

Bes regards
Nader Alishan Karami

cmungall:
Fixes #460


mcourtot:
Needs review or ok to merge?

PR #495, opened 2017-10-09
Change contact and add license (CC-) for ddanat

cmungall:


cmungall:
I reverted this as I accidentally merged before getting approval

PR #497, opened 2017-10-09
Update ddanat, made Petra contact, made license CC-0

cmungall:
Also changed "mould" to the more US-centric "mold" in description.


pfey03:
I approve

PR #517, opened 2017-11-16
Fix broken links in README and other explanatory files

tophat8855:


cmungall:
Thanks for the fixes! I have only had time for a quick review.

Note that the markdown files drive the jekyll site, so the .html is intentional

tophat8855:
Thanks for the feedback! I've reverted the .md changes back to .html, but kept my changes in the READMEs. I had been navigating through the files on Github and not the OBO site, so I didn't see that they needed to be .html.  I hope it is all clean now!

PR #536, opened 2017-12-15
updated source_url

mah11:
don't know whether I was supposed to change the "method: obo2owl" on the next line, so I've left it unchanged


cmungall:
You actually don't need the central build since you release both obo and owl yourself:

https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/fypo.yml#L7-L8

PR #585, opened 2018-03-29
Update ddpheno.md

cmungall:
Update contact to @pfey03


pfey03:
+1

cmungall:
oh and I just noticed there is no license on ddpheno. Do you want to use the same as for ddanat (CC-0)?

PR #606, opened 2018-05-02
Adding .obo products

### LINKED ISSUE #605: Ensure obo format products are registered
There are a small handful of MOD ontologies where the .obo product is not registered. These are important for the MOD community.

cc @doughowe

cmungall:
fixes #605


cmungall:
* fbbt - I added an obo product, unless there was a specific reason this was excluded @dosumis

PR #608, opened 2018-05-02
Update zfa.md

cerivs:
Changed URL for homepage. Added more emails

### DIFF
--- a/ontology/zfa.md
+++ b/ontology/zfa.md
@@ -3,11 +3,11 @@ layout: ontology_detail
 id: zfa
 in_foundry_order: 1
 contact:
-  email: zfinadmin@zfin.org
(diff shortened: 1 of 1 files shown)


cmungall:
Hi @cerivs - this failed the travis check as it has multiple contacts. 

Currently we have a restriction that there must be a single contact, this goes back to the early days of OBO. We have a discussion here: #73. If you like I can bring this up again. Do you feel you need multiple contacts? Or would you be happy with one designated contact and separate authors?

While this is resolved shall I modify your PR to get the home page change in?

cerivs:
Hi Chris
Single email for zfinadmin is ok.  We should revert the contacts so the
functioning URL goes into the release.

Thanks,
Ceri

jamesaoverton:
Fixed by #642.

PR #609, opened 2018-05-03
Add obo product for fypo

cmungall:
#605

caught by @doughowe 

@mah11 can we add any text on how the obo  and owl different in content?

### DIFF
--- a/ontology/fypo.md
+++ b/ontology/fypo.md
@@ -14,6 +14,7 @@ domain: phenotype
 homepage: https://github.com/pombase/fypo
 products:
   - id: fypo.owl
+  - id: fypo.obo
 taxon:
   id: NCBITaxon:4896
   label: S. pombe


mah11:
> @mah11 can we add any text on how the obo and owl different in content?

Sorry, @cmungall , maybe this sounds numptyish, but I don't understand the question.

In the biological sense, the content should be the same. The released fypo.obo and fypo.owl are both generated by Oort (neither @kimrutherford  nor I has had time to try switching to ROBOT) from the same obo-format editing file, so any non-biological differences would arise from that procedure.

cmungall:
Thanks, that's the answer I wanted. Some groups make the .obo more simplified, to accord with implicit expectations of OBO consumers

mah11:
Ah, we spit out a separate fypo-basic.obo for that sort of thing.

PR #612, opened 2018-05-08
Added preferred prefix for WBPhenotype

### LINKED ISSUE #611: Highlight ontologies that use a mixed case prefix
In general we use lowercase for the ontology IRI, uppercase for the class ID/URI.

There are some exceptions that use mixed case: FBbt, WBPhenotype. In these cases there is not a deterministic transform using uppercase/downcase. We annotate a `preferredPrefix` in the metadata file. 

We should show this on pages like

http://obofoundry.org/ontology/fbbt.html

Note: CHEBI does not fall into this category. The official OBO IDs and PURLs use all uppercase, despite the database sometimes being spelled ChEBI

cmungall:
Added highlight on page for ontologies that use mixed case/preferred prefix
Fixes #611
regenerated context


jamesaoverton:
I want to rebuild the generated files so I restored `registry/context.jsonld` in commit cf75a0707f874df1c9f0774348145c10dd476205.

cmungall:
Sorry I should have made a separate PR.

I made a separate ticket to resolve the duplicate issue #634

PR #630, opened 2018-06-13

mcourtot:
Removed requirement for a PURL ID in requests


mcourtot:
Can someone please review and merge? This is only about removing the requirement for a PURL account for PURL requests.

PR #636, opened 2018-06-27
added license

rvita:
Per obo services grant, updating metadata.


cmungall:
Related ticket #497, cc @pfey03

PR #653, opened 2018-07-01
update homepage mondo.md

rvita:


jamesaoverton:
There seems to be a typo in the license title.

PR #658, opened 2018-07-01
update email ontoneo.md

rvita:


jamesaoverton:
The contact name and email address don’t match now. It looks like a personal name but a group email. Is that right?

jamesaoverton:
@rvita I'd prefer a group email and a group name, or a person's email and a person's name, but this is a mixture of both.

rvita:
contact name and label match and are a person now.

PR #660, opened 2018-07-01
update license, homepage oarcs.md

rvita:


jamesaoverton:
I think `contact` should be `homepage` here.

PR #665, opened 2018-07-11
Update dpo.md

dosumis:
Attempting to correct PURL error by correcting id (dpo -> fbcv/dpo).

### DIFF
--- a/ontology/dpo.md
+++ b/ontology/dpo.md
@@ -1,6 +1,6 @@
 ---
 layout: ontology_detail
-id: dpo
+id: fbcv/dpo
 preferredPrefix: FBcv
 contact:
   email: temj2@cam.ac.uk


jamesaoverton:
This is a weird case. There's `dpo.md` file in this repo, but the PURL config is under [`fbcv.yml`](https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/fbcv.yml). I see two options:

1. If DPO has its own ID, then keep the `dpo.md` file here and create a `dpo.yml` file for PURLs
2. If DPO is derivative, then information should be in `fbcv.md` in this repo, and the `fbcv.yml` for PURLs is unchanged

The problem has existed for years now: https://github.com/OBOFoundry/purl.obolibrary.org/issues/15

dosumis:
The problem is historical. DPO was developed in the FBcv namespace (the oldest terms are part of a vocabulary that started in the early 90s) but it is now a self-contained phenotype ontology and is released as a separate artefact:

http://purl.obolibrary.org/obo/fbcv/dpo.owl 

It should certainly be treated as a separate phenotype ontology by the Foundry.  I was hoping it would be enough to get the .md to point to the right purl.  Happy to move to using http://purl.obolibrary.org/obo/dpo.owl if that is straightforward though.

jamesaoverton:
Ok. Then I would prefer to leave this `dpo.md` file unchanged, and have someone create a `dpo.yml` file in the PURL repository. The `fbcv.yml` PURL file should not be changed, so the current `fbcv/dpo` PURL will continue to work.

I'll close this PR (without merging) in favour of https://github.com/OBOFoundry/purl.obolibrary.org/pull/441

PR #690, opened 2018-08-07
update FAO contact

mah11:
merge this PR unless we hear otherwise from @dianeoinglis


cmungall:
Confirmed with Diane via email this is OK

Issue #693, opened 2018-08-13
TradeMarks and OBO licenses


cmungall:
Can an ontology be __both__ CC-BY and have the name trademarked? I believe in general TMs are compatible with CC-BY.

mcourtot:
Not sure anybody here would now. Maybe reach out to CC people?

nlharris:
@kltm might have thoughts about this.

kltm:
In my (not-a-lawyer) understanding, trademarks and copyright are quite different things. I would be surprised if an ontology had a trademark: who would bother registering it, or understanding if it was even applicable? (E.g. what is the good or service that you want your mark to cover?) Trademarks, I believe, involve attorneys and an application process, unlike copyrights which just "happen" naturally.
Copyrights and trademarks can bump against each other in interesting places though. For example, the Iceweasel debacle: https://en.wikipedia.org/wiki/Mozilla_software_rebranded_by_Debian#History_and_origin_of_name

alanruttenberg:
When I was at Science Commons we had suggested the approach of getting
trademark if desired and did not consider it incompatible with CC
licensing. It seemed reasonable to let people protect their good name that
way.
Alan

nlharris:
Are there still questions to be answered here, or can this be closed?

kltm:
Simply, "yes", they are compatible.
E.g. https://creativecommons.org/terms/

nlharris:
Thanks; closing.

PR #694, opened 2018-08-13
Update NCIT license and reference trademarked name.

balhoff:
This entry ended up with conflicting license keys. I have updated to the correct CC-BY 4.0, and removed the old pointer to the terms of use. I added information about the NCIt trademark to the description. However this ends up with a pretty long description value, which will take up an unusual amount of space in the ontologies listing. @rvita @cmungall is there another YAML key where this info can be placed?

cc @mellybelly @fragosog @NCIEVS

### DIFF
--- a/ontology/ncit.md
+++ b/ontology/ncit.md
@@ -6,11 +6,11 @@ contact:
   email: haendel@ohsu.edu
   label: Melissa Haendel
 license:
-  url: https://creativecommons.org/licenses/by/3.0/
-  label: CC-BY 3.0
+  url: https://creativecommons.org/licenses/by/4.0/
+  label: CC-BY 4.0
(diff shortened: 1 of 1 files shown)


cmungall:
see #693 

I don't think TM belongs in the license field, but we could show it other places

cmungall:
Would you be open to moving some of description field to the markdown section? We keep the descriptions short for fitting onto the front page table. We also try not to be redundant, e.g. license info in description.

(really we should not map this to dc:description, but to some other "short description property)

NCIEVS:
Is there something that I need to do?  Do I need to change some text?

balhoff:
I think I got everything you wanted in there (but feel free to check); @cmungall will merge if he is happy with how I shifted things around.

PR #700, opened 2018-08-15
Update mp.md

sbello:
remove termgenie link
correct taxon ID and label to Mammalia
added RGD browser link
corrected liscense

### DIFF
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -1,10 +1,10 @@
 ---
 layout: ontology_detail
-id: mp
-title: Mammalian phenotype
+id: MP
+title: Mammalian Phenotype
 license:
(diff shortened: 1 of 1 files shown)


cmungall:
You you change MP to mp? We use lowercase here, as per the ontology PURL not the class PURL. Uppercase assumed for actual ID prefixes, unless overridden for mixed case prefixes like FBbt

You can do this by going to
https://github.com/OBOFoundry/OBOFoundry.github.io/pull/700/files
clicking edit

or I can do it

Other changes look good, thanks!

sbello:
I did change it to capital MP since that matched how the IDs are formatted and I found the lowercase to look a bit odd. But if you all prefer it to be lowercase feel free to switch it back.

PR #701, opened 2018-08-29
update contact, license for mirnao.md

rvita:
per obo services grant, updated contact and license


jamesaoverton:
The license label doesn't match the URL. I already merged this, but we'll have to fix it.

PR #705, opened 2018-08-29
update upheno.md

rvita:
per obo services grant, updated contact, homepage, description, and license


jamesaoverton:
The license label does not match the license URL -- which is it?

cmungall:
@vasilevsky @dosumis @mellybelly - I suggest we go with CC-0

mellybelly:
+1 on CC-0

jamesaoverton:
Fixed by #727

PR #707, opened 2018-08-29
update tgma.md

rvita:
per obo services grant, updated contact and license


jamesaoverton:
License label does not match the URL.

jamesaoverton:
See #723

Issue #663, opened 2018-07-02
Define JSON-Schema for the OBO YAML metadata format


cmungall:
We should use JSON-Schema to define what is permissible for the yaml metadata files. We would run a schema validator in Travis-CI. This may still be augmented by procedural checks (e.g. check the tracker URL is not a 404) but the core structure can be verified.

If someone wants to speak up for any alternatives:

 - kwalify @kltm @dougl1sqrd
 - ShEx @balhoff (this would leverage the existing conversion to JSON-LD)
 - ??

kltm:
`kwalify`
* pros: 
  * standard part of ubuntu-based distros
  * easy-to-use syntax
  * already used by various GO projects
  * easy to use with travis
* cons:
  * not a standard
  * can be hard to get a hold of without fiddling with ruby gems outside of ubuntu/debian
  * fairly simple, not going much beyond regexps and set inclusion

jamesaoverton:
We use kwalify for the PURL system. I was initially quite happy with it, but this nasty issue has undermined my confidence: https://github.com/OBOFoundry/purl.obolibrary.org/issues/290

kltm:
@jamesaoverton If I'm reading correctly, I think there may be an easy workaround ()https://github.com/OBOFoundry/purl.obolibrary.org/issues/290#issuecomment-401983259

rctauber:
Hi @cmungall - I started a first-draft of the actual JSON-schema based on @rvita's initial work on the ontology metadata. I wrote a quick python script to implement it, but I wasn't sure how you wanted to configure this with the other tests that are already being run on Travis.

Also, I wanted to get some feedback on validation of the `license` field. We could enforce this validation by mapping the URLs of the licenses to their appropriate labels, but that would require maintaining a list of all the licenses. It would be nice, though, to enforce the official label, as some groups use `https://creativecommons.org/licenses/by/4.0/` with the label `CC-BY`, and others with the label `CC-BY 4`. Whereas, on the Creative Commons site, the official label is `CC BY 4.0`. If we do this, there will be many violations of this rule and I'm not sure if that's something we want to enforce... Otherwise, we could just validate that the `url` portion of the license is in URI format and that the `label` exists.

cmungall:
I think using a python lib for the json schema validation is a good idea.

We could change the license field to take a URI rather than object as parameter. Some of the initial modeling decisions were driven by what made things easy in the UI within the jekyll framework.

[...]

PR #710, opened 2018-08-31
Add metadata validation test

rctauber:
See #663 

The new test is included in the Travis CI stuff, so that will fail as there are currently 54 violations.


rctauber:
Current violations (with number of violating ontologies):
* missing 'contact' (11)
* missing 'description' (32)
* invalid 'license' values (6) *
* 'contact.label' is an email address instead of a name (3)
* 'product.id' is not an Ontology product (1) **
* 'homepage' has multiple values (1) ***

\* Auto-populating the license labels from the URLs which would help with this problem
\*\* ECO has a `.txt` product (mappings) - should all products be ontology format?
\*\*\* Could a project have more than one homepage, or should we require exactly one?

cmungall:
Next step to fill in missing values? I can do some, @rvita can you help?

cmungall:
I like making incremental progress - so how about temporarily relaxing some of the constraints so we can get the json-schema framework in place, and then gradually introducing them alongside necessary curation, a field at a time?

jamesaoverton:
@cmungall I agree.

rctauber:
See #740 for updated version.

PR #719, opened 2018-09-12
update license for fao.md

rvita:
per OBO services grant updated license info


mah11:
I don't have strong feelings, so no reason to oppose using CC0.

ValWood:
Yes, fine by me too.

PR #735, opened 2018-10-06
Update URIs

marijane:
This ontology is now located in the data2health GitHub organization.


marijane:
I noticed it doesn't resolve... the ontology isn't building because of it.  I figured it was because the URI was stale.  I will see if I can figure out how fix it.  Thanks!

marijane:
OK, other PR created!

PR #751, opened 2018-10-26
Vote Required: Update fp-006-textual-definitions.md

### LINKED ISSUE #750: Improve text of fp-006-textual-definitions
> Textual definitions MUST be unique within an ontology (i.e. no two terms should share a definition). Aristotelian definitions (“A is a B that Cs” or simply “a B that Cs” where B is the parent and C is the differentia) enable meeting this requirement

I don't know what "enable meeting this requirement means"

> Most terms that have children should have definitions

Where did this come from? Why should terms lacking children be exempted? This is arbitrary. Keep it simple, same rule applies to all terms.

> Instances, such as organizations or geographical locations, can benefit from definitions although it is understood that definitions for instances are not required. It is recognized that OBO format (e.g., versions 1.2 and 1.4) does not allow this as an option

I think this was added for GAZ, but let's keep it simple and focused on ontologies

> Textual definitions should be identified using the annotation property: ‘definition’ http://purl.obolibrary.org/obo/IAO_0000115. The source of the definition should be provided using the annotation property ‘definition source’ http://purl.obolibrary.org/obo/IAO_0000119

Most of the "Foundry" ontologies violate this, because they use the oio standard (ie axiom annotations on the definition axiom)

cmungall:
Fixes #750

### DIFF
--- a/principles/fp-006-textual-definitions.md
+++ b/principles/fp-006-textual-definitions.md
@@ -35,6 +35,8 @@ Logical definitions should agree with textual definitions. In fact, logical defi
 
 Textual definitions should be identified using the annotation property: ‘definition’ http://purl.obolibrary.org/obo/IAO_0000115. The source of the definition should be provided using the annotation property ‘definition source’ http://purl.obolibrary.org/obo/IAO_0000119, or as an axiom annotation on the definition assertion.
 
+Instances, such as organizations or geographical locations, can benefit from definitions although it is understood that definitions for instances are not required. It is recognized that OBO format (e.g., versions 1.2 and 1.4) does not allow this as an option
+
 Examples
 --------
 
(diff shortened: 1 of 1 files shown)


jamesaoverton:
Why drop the paragraph on instances?

cmungall:
How many ontologies have instances?

We could add it back in, I just wanted to simplify

nataled:
The principles are to offer guidance in addition to fiat proclamations. Instances were mentioned because it is/was a point of clarification. This will be discussed at the Editorial Working Group meeting today.

nataled:
+1 Darren N, Chris S, Chris M, Lynn S

nataled:
Can this be closed? I see the changes are public. It's unclear whether or not my review is still needed. If so, there does not appear to be a way to do it.

jamesaoverton:
Yes, this PR is already marked "Merged" (purple) to indicate that the proposed changes were merged into the master branch.

PR #759, opened 2018-11-12
Updated obofoundry.github.io entry for HANCESTRO 

daniwelter:
The file now includes a link to the HANCESTRO tracker and alternative versions of ontology (pre-reasoned and including BFO upper)


daniwelter:
I was told I should only list the main ontology under products in the yaml and the other versions under entries. I have updated the yaml now to point at the inferred version. Same with the md here.

daniwelter:
Would it be possible to merge this PR soon please? The related PR for the yaml has gone through, so the ontology PURL now redirects to hancestro_inferred while the ontology page on the Obofoundry page still states the old link.

Issue #19, opened 2015-08-08
Create portals or pages for specific domains


cmungall:
Many of our pages have a domain field. This is somewhat arbitrary but could be normalized.

It would be useful to have a page that lists all ontologies within a domain. Even better if this can be customized. For example, different community members could take responsibility for different domains: health, anatomy, environment, ...

cmungall:
I have created a first pass at this.

Check out the branch in the PR and run jekyll and look at the example page:

http://127.0.0.1:4000/community/microbial.html

![image](https://user-images.githubusercontent.com/50745/48369690-2667e680-e66c-11e8-91e1-95792b58eaef.png)

cmungall:
It would be cool to have browsers like OLS pick up the metadata here, and to allow browsing of groups of ontologies? @simonjupp 

TBD: how does the notion of a community overlap with the notion of a set of ontologies that are guaranteed to be mutually coherent, all inherit from obocore? @jamesaoverton

jonquet:
Indeed, this is a cool feature. 
Which is why BioPortal and AgroPortal have this for a bunch of years with the group/category and slices mechanisms. 

An extract of https://dx.doi.org/10.1016/j.compag.2017.10.012 which give an updated description of the feature:

> Within AgroPortal, each time an ontology is uploaded into the portal, it is assigned a group and/or category. Groups associate ontologies from the same project or organization, for better identification of the provenance. We have created a group for each use case, except the fifth one that is not a source of ontologies, and another one for the OBO Foundry. For each group we have deployed a specific slice (a restriction of the user interface to a specific group of ontologies) as explained later. Categories indicate the topic(s) of the ontology, providing another way to classify ontologies in the portal independently from their groups or provenance. As of now we have defined 20 general categories such as Farms and Farming Systems, Plant Phenotypes and Traits, Plant Anatomy and Development, Agricultural Research, and Technology and Engineering. These categories were established in cooperation with FAO Agricultural Information Management Standards (AIMS), which has maintained the VEST Registry since 2011.
> 
[...]

PR #761, opened 2018-11-12
WIP: Community pages

cmungall:
See #19

### DIFF
--- a/_layouts/community.html
+++ b/_layouts/community.html
@@ -100,10 +100,12 @@ layout: default
           </dd>
           {% endif %}
 
-          <dt>Trackers</dt>
+          <dt>Members</dt>
+          {% for member in page.members %}
           <dd>
-            <a href="{{page.tracker}}">{{page.tracker}}</a>
+            <a href="https://github.com/{{member}}">@{{member}}</a>
           </dd>
+          {% endfor %}
 
           {% if page.taxon %}
           <dt>Taxon</dt>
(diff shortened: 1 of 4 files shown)


matentzn:
Ah this is great! Very good idea. Once the schema is vetted by your reviewers I can create the community page for phenotype ontologies. It almost seems like a very shallow OBO foundry taxonomy against which the 151 OBO ontologies are tagged would be quite a simple way to generate these groupings you propose for the anatomy page automatically (and would be great for surveys too). Just a small one: There seems to be a mismatch between the ids listed under ontologies, and the ontologies mentioned later on. For example, fma, wbbt and fbbt are missing in the initial list. plana should be in there as well.

cmungall:
good point, not sure why they were omitted

yes, we could do this by scraping the domain tag in the ontology metadata, but the idea here is that each community can be a bit more than the sum of ontologies, and they can 'annotate' the role of each ontology in the community

mcourtot:
> It almost seems like a very shallow OBO foundry taxonomy against which the 151 OBO ontologies are tagged would be quite a simple way to generate these groupings you propose for the anatomy page automatically (and would be great for surveys too)


Yes! This would be very useful for us as we also wanted to leverage the 'domain' of each ontology to say things like "all values from this attribute are under the anatomical domain, so this attribute may be denoting an anatomical property and we should focus curation on this one'

I think there was some work done in @simonjupp team about this - Simon?

cmungall:
Merging now as there is no links to it so it won't disrupt anything; once in master easier for others to propose changes

PR #766, opened 2018-11-16
Editorialwg 

ramonawalls:
New page for principles review workflow.

### DIFF
--- a/docs/EditorialWG.md
+++ b/docs/EditorialWG.md
@@ -39,6 +39,8 @@ Current activities are focused on creating guidelines for reviews. Once those gu
 # Current Members #
 (in alphabetical order)<br>
 <br>
+Melissa Haendel<br>
+Suzanna Lewis<br>
 Darren Natale<br>
 Bjoern Peters<br>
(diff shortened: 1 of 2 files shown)


cmungall:
Looks good, thanks for doing this

Minor obsessive comment: I would replace the `<br>`s with a markdown list. But no big deal, we should eventually drive more of these things like roles for members from our metadata

nataled:
I don't see the need for two sections that differ (I think) in only one sentence. I suggest revising the following:

1) Change intro sentence to "This page describes the process for reviewing, updating, and creating [OBO Foundry Principles](http://obofoundry.github.io/principles/fp-000-summary.html). 

2) Remove the subheadings

3) Remove the second set of steps describing the 'creating'  part (but keep the intro sentence)

4) Change first step to "Discuss the principle on working group call 
 
5) Keep the intro sentence to the now-removed section ("The need for a new...") as a closing sentence for now. I suspect there will be a better place for it eventually. Also, remove the first occurrence of 'existing' in that sentence.

PR #767, opened 2018-11-19
Create planp.md

srobb1:


matentzn:
@cmungall apparently i dont have write access, so we need your blessing. Thanks.

cmungall:
hmm, you should have write access @matentzn, ping me again next time you try and  it doesn't work

PR #770, opened 2018-11-20
Harmonise Jekyll-synonymous URLs

katrinleinweber:
I used [html-proofer](https://github.com/gjtorikian/html-proofer/) to verify that none of those changes broke a link ;-)

There were ~~192~~ a few other ~~404s~~ broken URLs though. I may have time now and then to update a few old links: new PRs or a single one with as many as possible, or shall I push them into this one?

PR #774, opened 2018-11-22
Repair some URLs

katrinleinweber:
Even a bunch of PURLs are broken :-( Please [merge or keep open](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/770) as you see fit. I'll try to commit/PR more repairs when I need a break from studying this topic.


katrinleinweber:
You're welcome, and thanks for merging :-)

PR #781, opened 2018-12-08
Add activity_status

rctauber:
See #733

Each ontology is given an `activity_status` - `active` or `inactive` (if obsolete). The JSON schema includes a check for this required property.

**To do before merging**: add `orphaned` status to orphaned ontologies.


cmungall:
@rctauber looks like I may have caused a conflict by making a new release of derived products earlier this week. These have to be done on the command line. I recommend always doing any changes to derived products as a separate PR to avoid this

rctauber:
@cmungall I added the changes discussed in #787

PR #785, opened 2018-12-14
update ehdaa2.md license

rvita:
per OBO service grant & Jonathan Bard added license info, but there is no working homepage, changed to none. ontology is inactive, but not orphaned.


cmungall:
We shouldn't use fake values like "none" for homepage, in future we'll have a check to see if the value is a valid URL. Better to leave the field out. But if there is a github repo, the repo can serve as the homepage, so I added this

Issue #470, opened 2017-09-04
Check status: FIX


cmungall:
@jannahastings is this ontology still active?

jannahastings:
Hi @cmungall, as far as I know there was no active development on FIX (or REX for that matter) since the last date of update -- I just checked and that was May 2005 (!). We kept them alive as archives within the ChEBI downloads area for the last few years, because there were no ontologies that directly adopted them or took over all the content, and there were users at that time. I'm not sure what the current situation is with respect to usage.

rvita:
This ontology will be maintained at EBI and ChEBI is working on finding it a permanent home, but at this time it is deemed inactive.

rvita:
Janna Hastings @jannahastings states that fix is orphaned

nlharris:
FIX not yet marked obsolete on obofoundry.org

jamesaoverton:
FIX is not obsolete, it is orphaned, which means it has no active contact. This PR made updated its status #786 

We should do a better job of distinguishing orphan status on the main table, but that's being dealt with here #1126.

PR #787, opened 2018-12-14
update rex.md

rvita:
per OBO services grant, Janna Hastings & Gareth Owen, removed contact info for rex. It will be maintained at EBI in the future, but is currently inactive. It has no homepage or license info.


cmungall:
For inactive ontologies, add `activity_status: inactive`

Note that rex represents a solid serious ontology, I think we should at least preserve the version at a github repo to ensure it doesn't blink out of existence altogether, @G-Owen @jannahastings do you have interest in doing this?

cmungall:
Ah I see that @jannahastings provided an answer here: #470 

We should at least make sure we are pointing to the maintained archive at EBI

cmungall:
@rctauber shouldn't this validation checks as the email slot is empty?

rctauber:
@cmungall the test looks at the `ontologies.jsonld` file so it won't catch something like this until the registry files are updated. We could build a new `ontologies.jsonld` in Travis first before running the test?

rctauber:
I could add the following to the Makefile in #781 and that would help with the conflicts, as well:

```
### Validate Configuration Files

validate: $(ONTS) | tmp/ontologies.jsonld
	./util/extract-metadata.py validate $^ && \
	./util/validate-metadata.py && \
	rm -rf tmp

# Uses temporary directory to store builds
# Without overwriting the actual build files

tmp:
	mkdir -p $@

tmp/ontologies.yml: $(ONTS) | tmp
	./util/extract-metadata.py concat -o $@.tmp $^  && mv $@.tmp $@

tmp/ontologies.jsonld: tmp/ontologies.yml
	./util/yaml2json.py $< > $@.tmp && mv $@.tmp $@
```

jamesaoverton:
In #733 we agreed that "inactive" still requires a contact. No contact would mean "orphaned". Which applies to REX?

rvita:
It is not orphaned. Chebi is actively trying to find someone to take it over, but currently it is in purgatory and there is no contact or responsible person.

cmungall:
Sounds like it's orphaned. If someone comes available it can become unoprhaned but we should be trying to give an accurate picture of the ground truth

cmungall:
Where are we with this @rctauber @rvita ?

rvita:
It belongs to chebi & they cannot work on it at the moment. we can make it an orphan?

matentzn:
What needs to be done here?

jamesaoverton:
Fixed #788

PR #795, opened 2018-12-14
update ero.md contact and license

rvita:
per OBO services grant, updating contact person. license info coming soon.


jamesaoverton:
My understanding is that @rvita confirmed these changes over email, so I'm going to approve and merge.

PR #803, opened 2018-12-15
Removing all stale docs pertaining to old PURL server and sourceforge

### LINKED ISSUE #782: Sourceforge and Google Code links on Foundry pages
I have found a number of links to old pages on some Foundry pages. Specifically, 

http://www.obofoundry.org/docs/OBOPURLDomain.html
http://www.obofoundry.org/docs/Policy_for_OBO_namespace_and_associated_PURL_requests.html

Even the pointer to the obo-discuss mailing list on the latter page is to sourceforge.

cmungall:
Fixes #782
Also refreshing request policy doc


jamesaoverton:
I think this is ready to go, but it would be good for @mcourtot or @ramonawalls to do a quick review.

PR #805, opened 2018-12-15
Changes to id-policy doc

### LINKED ISSUE #732: Link to Figure 1 died
http://www.obofoundry.org/id-policy.html

cmungall:
Trimming down id-policy document
    
 - removing references to legacy OBO PURLs from a decade ago. No longer relevant
 - included a link to the history of the doc on github
 - removed outdated OBO format links
 - simplified mappings of OBO IDs to PURLs in terms of CURIEs
 - improved formatting
   
Fixes #732


jamesaoverton:
This is definitely an improvement. The only major problem I see is the out-of-date email addresses that I pointed out.

I would start by explaining IDSPACEs before OBO IDs, but I don't object to the current order, and I'm too lazy to do the work to change it. :)

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+be short form such as `part_of`, but these are mapped to CURIEs via the xref field.
+
+#### Format of Foundry-compliant URIs
+
+`FOUNDRY_OBO_URI ::= "http://purl.obolibrary.org/obo/" IDSPACE "_" LOCALID`
+
+#### Mapping of OWL ids to OBO format ids
+
+The recommended way is to use the [OBO JSON LD context](https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/registry/context.jsonld)
+
```
mcourtot: Would it be useful to add an example here?

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+This provides mappings between the OBO prefix and a base URL. See the [JSON-LD spec](https://json-ld.org/spec/latest/json-ld/) for a full specification of the expansion/contraction mechanism.
+
+CURIE to URI:
+
+`s@((**[A-Za-z_]***):(d+))@http://purl.obolibrary.org/obo/$1_$2@`
+
```
mcourtot: Is tat accurate for CURIE like d-acts?

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+## Policy for OBO Foundry ontologies
+
+All OBO ontologies should use PURLs that have a base URL within their allocated space (in CURIE form these would be prefixed with their ID space)
+
+
+## Response to Web requests for OBO URIs
+
+It is expected that the Foundry-compliant URIs behave, on the web,
+usefully. It will be the role of the OBO Foundry to supply generic
```
mcourtot: Suggest rephrasing to present tense: "The OBO Foundry supplies generic..." with maybe link to Ontobee?

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+software for responding to requests at URIs that identify OBO terms.
+
+
+We borrow the criteria from the Shared Name Initiative
+(http://sharedname.org/)as a base line. The OBO Foundry may issue
+further recommendations if experience shows them to be considered
+generally useful.
+
```
mcourtot: Not sure the above paragraph is still useful. SharedName is AFAIK dead, and it is kind of implicit that we'll update as practice evolves. If we want to keep the SharedName ack maybe move as credit at the bottom?

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+    ability to provide it may have to outlive the original ontology
+    developer's group or creator.
+
+5.  The provider of the URI documentation must be responsive to
+    community needs, such as the need to have mistakes fixed in a timely
+    manner.
+
+6.  URI documentation must be open so that it can be replicated and
+    reused.
+
```
mcourtot: I think the above are a bit verbose and not specific. Eg 1. refers to a database, whereas in most cases there are no database for ontologies.

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+## Policy for OBO Library ontologies
+
+OBO Library ontologies are not constrained by this policy, however, we
+recommend that they follow it nonetheless, for three reasons. First, it
+provides a uniform experience and sets expectations for ontology
+clients. Second, by doing so library ontologies will be able to take
+advantage of shared infrastructure. Third, ontologies that eventually
+join the foundry would have to disrupt their ids if they had to change
+to follow this policy.
+
```
mcourtot: We have been requesting prefix and purl registration as part of the registering in the OBO library, de facto 'forcing' adoption. If we want to keep the above as is maybe add 'To promote this without creating additional burden on ontology maintainers registering a prefix and associated PURL with the OBO library automatically creates an entry in the Ontobee service.'

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+**Allocating IDSPACEs**
+
+IDSPACEs within the OBO library are unique for a given project and are
+chosen not to conflict with prefix for xrefs. Although IDSPACEs are
```
mcourtot: refs is very GO centric - not clear this is still true.

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+case-sensitive, there will never be more than one IDSPACEs that are the
+same when compared in a case-insensitive manner. Therefore,although
+"GO" and "go", "Go" and "gO" are different IDSPACEs, the IDSPACE "go",
+"Go" and "gO" will not be used as "GO" has already been allocated.
+
+
+
+A registry of allocated IDSPACEs will be maintained. Requests for an
+IDSPACE should be made by sending mail to
+obo-discuss@lists.sourceforge.net, cc obo-admin@fruitfly.org. A request
```
mcourtot: wasn't there a new group email address? Though looking at the latest emails it seems everybody still uses the sf one.

id-policy.md
```diff
@@ -0,0 +1,231 @@
...
+
+For example, for the version of OBI released 2009-11-06, the OWL
+document is accessible
+http://purl.obolibrary.org/obo/obi/2009-11-06/obi.owl
+
+
+## Home page
+
+
+If the ontology has a home page on the Web, it is accessible at
```
mcourtot: we recommend it is accessible at...? I don't think this happens in practice. Similarly for tracker browser etc. Or is that auto generated from the registry entry?

id-policy.md
```diff
@@ -0,0 +1,133 @@
...
+Additionally other ontology products can be included underneath this. For examples, see "products" under [http://obofoundry.org/ontology/go](http://obofoundry.org/ontology/go)
+
+## Versions of ontologies
+
+Versions are named by a date in the following format: YYYY-MM-DD. For a given version of an ontology, the ontology should be accessible at the following URL, where *idspace*; is replaced by the IDSPACE in lower case:
+
+- OWL: http://purl.obolibrary.org/obo/*idspace*/YYYY-MM-DD/*idspace*.owl
+- OBO: http://purl.obolibrary.org/obo/*idspace*/YYYY-MM-DD/*idspace*.obo
+
+For example, for the version of OBI released 2009-11-06, the OWL document is accessible at <http://purl.obolibrary.org/obo/obi/2009-11-06/obi.owl>.
```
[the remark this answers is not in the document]
cmungall: ah yes, I had noticed that, I figured I'd first preserve the original URLs etc and then do a second pass, but we should have a working URL here

id-policy.md
cmungall: They are not auto-generated, and in practice this is only done for 35 ontologies, so I second the change of wording to a recommendation

cmungall: In fact, why recommend it at all? It's one more URL that can fall out of sync (for humans we have the tracker link in the obo metadata). I don't really see the use case for a PURL for a tracker. But can maybe be tackled on 2nd pass

id-policy.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Allocating IDSPACEs
+
+IDSPACEs within the OBO library are unique for a given project and are chosen not to conflict with prefix for xrefs. Although IDSPACEs are case-sensitive, there will never be more than one IDSPACEs that are the same when compared in a case-insensitive manner. Therefore,although "GO" and "go", "Go" and "gO" are different IDSPACEs, the IDSPACE "go", "Go" and "gO" will not be used as "GO" has already been allocated.
+
+A registry of allocated IDSPACEs will be maintained. Requests for an IDSPACE should be made by sending mail to <obo-discuss@lists.sourceforge.net>, cc <obo-admin@fruitfly.org>. A request should include information about the ontology, such as scope and maintainer and a confirmation that the ontology is open access.
```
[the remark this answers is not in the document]
cmungall: yes

id-policy.md
cmungall: will do

id-policy.md
cmungall: probably not, not for PRO either

id-policy.md
cmungall: +1

id-policy.md
cmungall: agreed

id-policy.md
cmungall: google groups

id-policy.md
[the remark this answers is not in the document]
zhengj2007: Is Allocating IDSPACE part of registering an ontology? Is it better to put the following link here,
http://www.obofoundry.org/docs/Policy_for_OBO_namespace_and_associated_PURL_requests.html?

cmungall:
Merging for now, will return to deal with other comments later

PR #810, opened 2018-12-30
update go.md contact, license, homepage

rvita:
per OBO services grant and email with Suzi Aleksander, updated homepage, contact, and license.


cmungall:
Sorry, shouldn't check github on my phone, I see it's actually removing David as a contact, but David should be the contact

Issue #792, opened 2018-12-14
sep needs contact info


rvita:
per OBO services grant, checking that all ontologies can be contacted and was unable to contact sep. Sample   processing and separation techniques contact label is  'SEP developers via the PSI and  MSI mailing lists' with email address of psidev-gps-dev@lists.sourceforge.net. This address was used on several occasions and no answer was ever received. sep also needs license info, will request once I have a contact person.

cmungall:
All mailing lists here:
http://psidev.info/mailing-lists

General docs for full PSI family here:
http://www.psidev.info/groups/controlled-vocabularies

From this:
https://github.com/HUPO-PSI/gelml/commits/master/CV/sep.obo

All commits were by Yasset Perez-Riverol @ypriverol, so I nominate them as the contact, if they are willing

rctauber:
SEP now has contact info, but it's still missing a license.

beckyjackson:
SEP is marked as obsolete and inactive. Do we want to close this issue?

nlharris:
Seems like this issue could be closed.

Issue #791, opened 2018-12-14
ncbitaxon needs contact & license


rvita:
per obo services grant checking that all ontologies have a human's email address, noted that obo-taxonomy@lists.sourceforge.net is the provided contact for ncbitaxon. This email address did not respond to me when I emailed it. Also noted license is missing, will request it once there is a human to contact.

cmungall:
From
http://obofoundry.org/ontology/ncbitaxon

```
Releases of the obo/owl happen when the Continuous Integration Job is manually triggered. Currently this must be done by an OBO administrator. There is currently no fixed cycle, and this is generally done on demand. The team that informally handles this are:

James Overton, IEBD/OBO
Frederic Bastian, BgeeDb/Uberon
Chris Mungall, LBNL/GO/Monarch/Uberon/OBO
Peter Midford, Phenoscape
```

In practice it's me that presses the button on Jenkins so it should be me listed. This would be more useful than the generic obo contact. However, my name is on too many ontologies. Does anyone else from the @OBOFoundry/community or OBO taxonomy group want to be listed?

@pmidford? @balhoff

fbastian:
You can add me, yes.

balhoff:
@rvita sorry, I used to moderate obo-taxonomy@lists.sourceforge.net, but it mostly just gets spam these days. I wonder if there is any interest in reviving it.

cmungall:
upstream ticket: https://github.com/obophenotype/ncbitaxon/issues/14

beckyjackson:
License taken care of in #837. We still need a single contact though. It looks like @fbastian agreed to be added, should we move forward with this?

cmungall:
Yes, please list @fbastian

cmungall:
@rvita can you update the metadata? I think some people have been emailing the list and not getting a response (sorry @sierra-moxon)

rvita:
update was done & merged

yongqunh:
Great. Thanks!

PR #829, opened 2019-02-21
Metadata Validation Updates

rctauber:
Included in this update:
* Separate each metadata field into its own schema to allow for better error messages
* Add more verbose error and warning messages
* Remove `validate: false` fields
* Exclude orphaned ontologies from contact and license checks

Issues:
* ERO needs license (#795)
* NCBITaxon needs license (#791)
* MIRO needs license (no issue)
* SEP needs license (#792)

We can always add the `validate: false` back in for those four above, but I think it would be better just to resolve those outstanding issues (if possible).


rctauber:
ERO and NCBITaxon have labels now. We are still waiting on MIRO and SEP. The `validate: false` field has temporarily been returned to these entries, but will be removed as soon as possible.

PR #837, opened 2019-03-05
Add CC0 license to NCBITaxon

rctauber:
See https://github.com/obophenotype/ncbitaxon/issues/14


jamesaoverton:
There was agreement (and no disagreement) on the upstream issue, so I'm in favour of this. Do you approve @cmungall ?

Issue #779, opened 2018-12-06
Add grid for metadata report results


rctauber:
Each ontology should have a result from the metadata validation: pass, warning, or fail. Currently, the system only has pass or fail, with a few ontologies not included in the checks. The status of the metadata should be visible on the front page of the OBO registry website.

**Proposal:**
An ontology will receive a `pass` when it passes the full `metadata-schema.json` check.
An ontology will receive a `warning` when it fails the full check, but passes the `metadata-schema-lite.json` check.
An ontology will receive a `fail` when it fails both checks.

`metadata-schema.json` includes restrictions on licenses. In the future, we can make the `github` contact information required as well.

bpeters42:
We will implement a new status 'failing validation', working just like obsolete, meaning an incomplete registry entry is allowed for these. Like the obsoletes, these will be displayed at the bottom of the list. For non-failing, non-obsolete, non-orphan ontologies, all pull requests will have to the metadata report with 'pass' or 'warning'. For the warnings, we will develop a system that classifies them into critical (non-standard license) vs. 'info' (like no publication associated), and decide on how to display them in the future.

rctauber:
Work on branch: https://github.com/rctauber/OBOFoundry.github.io/tree/metadata-grid

The validation process outputs two products. The first is the violation report, which only contains only the error messages (these are also logged in the console so you can easily see an issue in the Travis build). The second is a metadata grid with all fields, for example:

![image](https://user-images.githubusercontent.com/16750150/53896548-109a7a80-3fe9-11e9-8466-7c90aecc38e4.png)

These results are sorted in the following order:
1. Foundry status (true, false)
2. Ontology activity status (active, orphaned, inactive)
3. Validation status (pass, info, warn, fail)
4. Alphabetical

cmungall:
@dougli1sqrd @kltm this is analogous to the checks we do on GO annotation sources, isn't it (of course, the OBO check is ontologies rather than annotations)? We have checks on one axis and source on the other. Is there any code that can be reused?

@rctauber do we plan to meld the robot report and reasoning checks into the same summary table? If so I expect we'll want to do this in the OBO jenkins job.

jamesaoverton:
Note that in this grid, most of the columns correspond directly to a JSON schema file: https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/util/schema

[...]

Issue #830, opened 2019-02-26
Ordering of ontologies on OBO page is no longer sorted alphabetically


cmungall:
We should restore this (within each category, e.g. foundry)

jamesaoverton:
Fixed by #842

PR #842, opened 2019-03-12
Create metadata grid & sort homepage by results

rctauber:
Includes a new validation product, for #779. This is found at `reports/metadata-grid.csv` after running validation locally (`make validate`, or just `make` if you want to do the full build). There is also an HTML output (`reports/metadata-grid.html`) that we can potentially use later.

When building `registry/ontologies.yml`, this grid is now used to sort the ontologies in the following order:
1. foundry
2. active
3. orphan
4. inactive
5. obsolete

Within each of those categories, the ontologies are sorted in order of their validation results:
1. pass
2. info
3. warn
4. fail (no items on the master branch should be failing, though)

Finally, they are sorted alphabetically within those sections (see #830).

Orphaned and inactive ontologies appear right about the obsolete ontologies with a grey background. Inactive ontologies are marked with 'inactive':

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,6 +96,8 @@ registry/publications.md: util/extract-publications.py registry/ontologies.yml
 
 # generate both a report of the violations and a grid of all results
 # the grid is later used to sort the ontologies on the home page
+validate: reports/metadata-grid.csv reports/metadata-grid.html
+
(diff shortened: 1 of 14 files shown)


jamesaoverton:
We'll continue to improve this stuff, but this is already good progress, so I'm going to merge it.

Fixes #779 and #830.

PR #887, opened 2019-06-04
Update new-ontology-request.md

kltm:
Merge form information in to GH template, as discussed on operations call this morning.

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-ontology-request.md
+++ b/.github/ISSUE_TEMPLATE/new-ontology-request.md
@@ -65,24 +65,32 @@ Another other significant information that you'd like to let us know.
 
 
 
+```
 ---
-id: ONTOLOGY ID SPACE E.g. obi
+layout: ontology_detail
+id: ONTOLOGY IDSPACE (LOWERCASE) E.g. obi
 title: FULL NAME E.g. Ontology of Biomedical Investigations
 description: BRIEF ONE SENTENCE DESCRIPTION OF YOUR ONTOLOGY
 homepage: URL
+contact:
+  label: CONTACT NAME
+  email: CONTACT EMAIL
 license:
+  label: CC-BY 4.0
   url: http://creativecommons.org/licenses/by/4.0/
-tracker: GITHUB TRACKER URL
+tracker: TRACKER URL
 source_url: URL THAT RESOLVES TO OWL FOR YOUR ONTOLOGY
-formats: Note that OWL-RDF/XML is the only required format.
-domain: E.g. the GO covers the processes, functions and cellular locations of gene products
+products:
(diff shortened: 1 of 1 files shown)


kltm:
@cmungall 
Note the differences between the two versions:
https://github.com/OBOFoundry/OBOFoundry.github.io/pull/887/commits/2dcc97c207f338427ecce34a1e979986135dc4fc
https://github.com/OBOFoundry/OBOFoundry.github.io/pull/887/commits/777bff8c48516aeb7bccbdf5eae4cac825d4223b
The first is more like what you originally had: much more "computable" possibly, but hard to use. The second is more like the form and easier for users to deal with. I'm not sure which the group would like more, but one of them should be correct.

jamesaoverton:
Thanks @kltm.

The Google Form has lots of good information that we want users to think about. We want to track these requests in GitHub with all our other issues. As I said yesterday, while it's possible to hook the Google Form directly to GitHub, it won't be easy and the result will be fragile. So putting the information from the Google Form into the GitHub issue template is a solution. That was the consensus on the call yesterday (as I heard it), and that's what you've done with 777bff8c48516aeb7bccbdf5eae4cac825d4223b. For a GitHub issue, it will be quite verbose, but there's a lot of important information here that goes beyond the YAML+Markdown registry file that we need in the end.

My opinion is that we use the longer 777bff8c48516aeb7bccbdf5eae4cac825d4223b for now, see how it goes, and then revise.

kltm:
@jamesaoverton The second version is the one prepped in the PR. I'm not sure how these pass through to get OBOF consensus, but it is ready to go and can always be reverted.

jamesaoverton:
Yes, I was just commenting, but I'll make an Approve review. I'd still like to hear @cmungall's opinion.

kltm:
Just so nobody is waiting on me, I was not intending on doing the final merge, waiting instead for somebody more familiar with the OBOF process to take it.

jamesaoverton:
In working on #888 I was surprised that the YAML from this template is missing several required fields, and so it fails validation. Although it's a bit noisier, I think it's less confusing if all the required fields are present. So I added these required fields to the bottom of the template.

cmungall:
I'm confused by this PR. It seems to contain two alternate commits? Anyway I will go ahead and merge..

cmungall:
OK, I merged. We seem to have a merger of both styles (verbose and yaml-like)

Personally I think there is too much boilerplate now

cmungall:
EWG and @nataled will iterate on this

nataled:
I modified the template to strip out pretty much all the text except for question titles (so to speak). All explanatory text was extracted to a separate document: http://obofoundry.org/docs/NewOntologyRegistrationInstructions.html

PR #893, opened 2019-06-18
Refactoring and other code improvements

lmcmicu:


jamesaoverton:
For the OBO Technical Services grant we said that we would review and refactor the code for our shared infrastructure to facilitate long-term maintenance. I asked @lmcmicu to do this refactoring a couple of months ago, and then I dropped the ball. @rctauber and I will review this now. If @cmungall or anyone else has time to review in the next little while, please let us know.

util/extract-metadata.py
```diff
@@ -1,242 +1,227 @@
...
-    args = parser.parse_args()
+  args = parser.parse_args()
 
-    func = args.function
-    func(args)
+  func = args.function
+  func(args)
 
 
 def validate_markdown(args):
```
jamesaoverton: This job is now done better by the `validate-metadata.py` script and schemas. It isn't used in the `Makefile`. I think we should remove this function and subcommand parser.

lmcmicu: Hi James (and all), thanks for having a look. FYI you probably already know this, but you can hide whitespace changes temporarily while reviewing the commit by clicking on the little settings (gear) icon on the Files Changed tab.

jamesaoverton: Thanks @lmcmicu. Actually I didn't know about that option. You learn something new every day.

lmcmicu:
Thanks, everyone, for taking a look.

Issue #618, opened 2018-05-29
Request for ORNASEQ prefix and PURL


safisher:
We are creating an application ontology for RNA next-generation sequencing. The creation of the ontology included adding new terms to OBI. These terms have been added but OBI (https://github.com/obi-ontology/obi/pull/934) and I believe a new release is imminent. If that turns out to not be the case then we will update ORNASEQ with temporary term IDs for the new OBI terms.

Requested prefix: ORNASEQ
Ontology Title: Ontology of RNA Sequencing
Download link: https://github.com/safisher/ornaseq/blob/master/ornaseq.owl
Home Page: https://github.com/safisher/ornaseq, http://kim.bio.upenn.edu/software/ornaseq.shtml
Contact e-mail: safisher@upenn.edu
Purl.org user ID: safisher@upenn.edu
Description: An application ontology designed to annotate next-generation sequencing experiments performed on RNA.

cstoeckert:
Do you still need this? Apologies for the delay.

safisher:
Yes, we would still like to include our ontology in OBOFoundary.


On May 21, 2019, at 5:50 PM, cstoeckert <notifications@github.com<mailto:notifications@github.com>> wrote:


Do you still need this? Apologies for the delay.

cstoeckert:
I found an issue with the posted owl file. Please see https://github.com/safisher/ornaseq/issues/3

cstoeckert:
Discussed on OBOF operations call June 18, 2019 and approved. Now moves to Technical Working Group to implement. I can help as needed.

cstoeckert:
Will close once the registry and PURL is done.

zhengj2007:
@safisher Since your request has been approved, we will add your ontology to the OBO Foundry. For doing so, could you please provide following required information:
1. Ontology license
Note: The recommended OBO Foundry ontology license is CC-BY (version 3 or later).
2. Issue tracker 

It would be better if you can also provide information about:
1. Related OBO Foundry ontologies
2. Intended use/related projects
3. Data source

The detailed instruction can be found on the page: http://obofoundry.org/docs/NewOntologyRegistrationInstructions.html

Thanks!

safisher:
Thank you for your help.

**1. Ontology license:** CC-BY version 4.0
**2. Issue tracker:** https://github.com/safisher/ornaseq/issues
[...]

PR #895, opened 2019-06-20
Create ornaseq.md

zhengj2007:
Issue #618


zhengj2007:
Hi, @cmungall, Could you please help on reviewing and adding the ontology to the OBO Foundry homepage? Thanks!

Issue #891, opened 2019-06-18
Proposed schema change: active ontologies should have a tracker


cmungall:
I'm trying to think of a case where there is a reason for an active ontology to not have a tracker

I propose we force inclusion of a tracker for active ontologies

This would presumably be added here: 

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/util/validate-metadata.py#L124-L138

nlharris:
I think this is now required or at least recommended (see principle 20, http://www.obofoundry.org/principles/fp-020-responsiveness.html)?

matentzn:
Yes is required now, can be closed.

PR #897, opened 2019-06-24
Using correct tag for tracker in MICRO

cmungall:
Any idea why this wasn't flagged as a schema error @rctauber ?

See also #891


jamesaoverton:
@rctauber is on vacation this week.

The schemas do not require a tracker, and the validator does not complain about unrecognized keys. I think we should start doing both those things.

PR #969, opened 2019-08-05
Replaced (outdated) list of members with link to membership page

nlharris:


nlharris:
I'm not empowered to add reviewers, but presumably @ramonawalls or @cmungall could approve this and the other two minor PRs...

PR #992, opened 2019-08-07
add tracker rs.md

rvita:
per Jennifer Smith


jamesaoverton:
The `tracker` field points to the repository. I think it should point to the actual issue tracker: https://github.com/rat-genome-database/RS-Rat-Strain-Ontology/issues

PR #1026, opened 2019-08-14
Adding period at end to make valid ttl

cmungall:


cmungall:
I thought I checked with riot, and it does parse, but it also emits an error message which I had missed. Really we should do an extra validation step of all products...

regarding your previous comment, yes the syntactic structuring is a bit odd, maybe I should have used rdflib here...

PR #1052, opened 2019-09-02
Creat iceo.md

Lemon-Liu:
ICEO was approved as an Open Biomedical Ontology (OBO) library ontology.


jamesaoverton:
Also merged PURL file https://github.com/OBOFoundry/purl.obolibrary.org/pull/570

PR #1064, opened 2019-10-08
New usage examples and imports

pbuttigieg:
@cmungall apropos examples provided by @kaiiam

### DIFF
--- a/ontology/envo.md
+++ b/ontology/envo.md
@@ -38,6 +38,9 @@ dependencies:
  - id: ro
  - id: chebi
  - id: ncbitaxon
+ - id: foodon
+ - id: po
+ - id: go
 usages:
  - type: data-annotation
    description: "describing species habitats"
@@ -82,6 +85,11 @@ usages:
     example:
       - url: https://www.ncbi.nlm.nih.gov/nuccore/NC_016642
         description: "Annotation of habitat of Pseudovibrio sp. FO-BEG1 to marine environment"
+  - user: https://www.planetmicrobe.org/project/
+    description: Annotation and semantic search over microbial data sets
+    example:
+      - url: https://www.planetmicrobe.org/project/#/samples/200
+        description: "Exposure of ENVO annotations associated with an omic dataset derived from a sample of marine water near Lisboa, taken as part of the Ocean Sampling Day Project (https://www.microb3.eu/osd.html)"  
 activity_status: active
 ---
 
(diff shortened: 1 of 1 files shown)


cmungall:
I don't seen any examples of use of ENVO on the linked page?

kaiiam:
> I don't seen any examples of use of ENVO on the linked page?

@cmungall in Planet Microbe we are using OBO terms to standardize the metadata of select marine 'omics projects. If you look at the ``Ontology Label`` and ``Unit`` columns you'll see that were employing a variety of OBO semantics to richly describe the data.

cmungall:
@kaiiam - it looks like PM is removing the zero padding in the IDs this is a really bad idea can you follow up?

kaiiam:
@cmungall what I think you are referring to is the value column, which was sourced form the data-stets themselves in this case OSD. We've added to that data the ``Ontology Label`` which is the full purl as annotation in this case biome http://purl.obolibrary.org/obo/ENVO_00000428. I had planned on further cleaning the datasets themselves to change them from (for example) ``estuarine biome (ENVO:1000020)`` to ``estuarine biome (ENVO_01000020)`` or ``estuarine biome http://purl.obolibrary.org/obo/ENVO_01000020`` or perhaps just the purl itself.  Let me know what you think it should be changed to. 

| Ontology Label | Dataset Label | Value |
| -- | -- | -- |
| [biome](http://purl.obolibrary.org/obo/ENVO_00000428) | Biome | estuarine biome (ENVO:1000020) |

cmungall:
Right, this is the usage field for the ENVO record, so we want to record
non-trivial use of ENVO as values. It's great that PM is using OBO as field
descriptors too, we don't really have a place for this - though you could
make a small application ontology for this?

Cleaning up the fields would be great. Happy for them to remain CURIEs, but
you can link out using PURLs

kaiiam:
The important usage of OBO terms (primarily envo) in planet microbe is for
semantic search based on dataset metadata. From our site:
https://www.planetmicrobe.org/project/#/search users can search based on
unified OBO semantics (see the Add Filter tab) for a variety of common
oceanography metadata such as concentration of chlorophyll, particulate
carbon, turbidity of water etc.

We do have an application ontology which you helped me to create which
imports form ENVO, OBI, IAO, UO etc. Bringing together a rich OBO semantic
layer for term annotation and standardization.

PR #1067, opened 2019-10-19
Add optional separate aberowl_id for an ontology.

### LINKED ISSUE #1053: incorrect external link
On the Relations Ontology page (http://www.obofoundry.org/ontology/ro.html), the link for AberOWL goes to the Radiomics Ontology.

tophat8855:
NOTES: in the case of the Relations Ontology, the OBO id is 'RO' but the AberOWL
id is 'RELO'. Using the OBO id of 'RO' to link to the AberOWL page, sends the
user to the Radiomics Ontology page of AberOWL. This change allows a contributor
to request a particular AberOWL id/redirect link.

Resolves #1053

### DIFF
--- a/faq/how-do-i-edit-metadata.md
+++ b/faq/how-do-i-edit-metadata.md
@@ -38,7 +38,6 @@ taxon:
   id: NCBITaxon:33208
   label: Metazoa
 domain: cells
-aberowl_id: ab
 tracker: https://code.google.com/p/cell-ontology/issues/list
 termgenie: http://cl.termgenie.org
 mailing_list: https://lists.sourceforge.net/lists/listinfo/obo-cell-type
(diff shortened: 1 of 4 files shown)


faq/how-do-i-edit-metadata.md
```diff
@@ -38,6 +38,7 @@ taxon:
   id: NCBITaxon:33208
   label: Metazoa
 domain: cells
+aberowl_id: ab
```
jamesaoverton: I commented on the general approach of this PR in issue #1053, but my only specific problem with this PR is that I don't want `aberowl_id` in the basic example: I don't want people to use it unless they absolutely have to.

tophat8855:
Absolutely. It would create code debt if RO is the only ontology with this problem and is probably not worth it for just that one.

faq/how-do-i-edit-metadata.md
cmungall: agreed

tophat8855:
I removed aberowl from the general example. Is there anything else I can do for this PR? Thanks.

jamesaoverton:
Sorry for the long delay. I added `aberowl_id` to our JSON schemas, which I think addresses the only remaining issue.

PR #1073, opened 2019-11-14
Update gno.md

edwardsnj:
Provide link to the GNOme Glycan Composition Browser and Subsumption Navigator as a browser.

Note, the example provides a link with an accession in it, but its not clear whether this is necessary/required as the OBOFoundry user-interface does not appear to provide any way to set this accession for the user.


jamesaoverton:
This seems fine to me, so I'll merge it.

If you need more control over `browser` stuff, please open an issue describing your use case.

PR #1074, opened 2019-11-18
Update bto.md

matentzn:
Need one item of help:

On http://obofoundry.org/, the license is not rendered; but there is a corresponding image

The html says: `CC BY-ND 4.0` - Probably because it is ND?


matentzn:
Comment above redundant. Its not ND, its just CC-BY 4.0, so after merging, should be fine.

Issue #1085, opened 2019-11-26
Extend metadata to include review information


cmungall:
We currently have reviews in noncomputable form here:

http://obofoundry.org/docs/CompletedReviews

We should instead put this in the metadata and derive the displays

E.g. foo.md:

```
...
reviews:
 - date: 2016
    url: <link to pdf... hopefully in future to a github ticket>
    status: pass
    comments: ...
```

 * date: REQUIRED can be EITHER 4-digit year OR ISO-8601 YYYY-MM-DD string
 * url: OPTIONAL <because we don't have old ones>
 * status: REQUIRED pass OR fail OR awaiting...?

Once this is in place we can dynamically

 * include the URL from the front table
 * include this in per-ontology pages
 * make the table in http://obofoundry.org/docs/CompletedReviews

Note that for a while the existing foundry tag in the md will be redundant but that is OK. We should have a consistency check.

cmungall:
to be clear, the idea is to put the review metadata into the main metadata file.


ie in `ontology/bfo.md` we would have:

```
id: bfo
title: basic formal ontology
...
reviews:
 - date: 2016
    url: <link to pdf... hopefully in future to a github ticket>
    status: pass
    comments: ...
```

All: Do we want reviews to be a list? I think so, as we may have re-reviews. Should we also have some marker to say one of the reviews is the 'current' one? I think this is better than relying on ordering.

nlharris:
what's the status of this?

wdduncan:
@nlharris I fear this ticket has gone stale.

nlharris:
Should we close it?

wdduncan:
Let's ask @cmungall

matentzn:
I dont think this should be closed. This is one of the things we need to take care of when we get new funding; maybe you can have a github tag that expresses this sentiment. No one will volunteer to do such a thing, but its a good item for a wishlist for when we are doing a serious overhaul of the site.

matentzn:
But there is one practical aspect to this ticket we can do right now: add the reviews element to the json schema, and add the 8 reviews we currently have in the way chris suggests to the md files.

nataled:
Closing since no longer relevant to current procedures.

PR #1088, opened 2019-12-10
[WIP] Add functionaly for completeted reviews to be data driven.

wdduncan:
This PR address issue #1085 

New reviews are entered as an md file in the reviews directory.
The logic for displaying the review table is in _includes/completed_reviews_table.html

I could not find where to add schema information (e.g., datatype, required, optional, etc.) regarding the keys in the reviews/*.md files. Perhaps @jamesaoverton or @beckyjackson know?


cmungall:
There is a json schema file. I'm surprised this passes checks...

I prefer the more normalized screens 3i suggested in the ticket,  where
review is an object

jamesaoverton:
I don't see the point of ten `reviews/*.md` files and code for combining them. I suggest either:

1. one `reviews.yml` file, or
2. add the review data to the relevant registry file

JSON schemas for the registry entry YAML files are here: 

https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/util/schema

If review data is separate from registry data, then I'd prefer the the review schema also be separate from the registry schemas, somewhere else in `util/`.

cmungall:
ah yes, I agree with James. I was looking at the diffs on my phone before, i didn't see these were separate files (this explains why there was no schema violation!)

My intent in specifying #1085 was James' option 2 "add the review data to the registry file of each reviewed ontology". This also explains my comment about having each review be an object.

I will go back and further specify in the ticket

cmungall:
@wdduncan can you address @jamesaoverton's comments

also, it's usually a good idea not to check in derived files - we have a conflict now.. may be easier to make a new PR?

also can you change the schema to what I specified in the original ticket? Right now we have:

```
- id: chebi-review
  layout: completed_reviews
  ontology_label: Chemical Entities of Biological Interest
  ontology_url: http://www.obofoundry.org/ontology/chebi
  review_document_label: n/a
  title: ChEBI Review
  year_reviewed: 2010
```

I think we should record the actual date, not just the year. @nataled do we have the dates for all our original reviews?

nataled:
@cmungall I am not aware of such. I presume you are referring to those reviewed before the EWG existed? We never tried figuring it out. Even now I think we only display the year. Our thinking was that it was extremely unlikely we'd review an ontology more than once a year. What is the concern here? Does it have to do with which release was reviewed? I'm not even sure we captured that, though it does seem like a good idea, even in lieu of dates.

cmungall:
@wdduncan - do you want to start a new PR for this?

wdduncan:
@cmungall @jamesaoverton 
Just getting back around to working on this.   
What exactly do you want with this?  I **thought** what we wanted was for  the Completed Reivews to be maintained by adding new `.md` file.  For example, when the UBERON review is completed, we would add an `reviews/uberon-review.md` file.
Currently, this is maintained by modified the html in the bottom portion of `docs/CompletedReviews.md`.   
Not trying to be difficult, but if you have a different workflow in mind could you please describe it?  I don't know how to put all the reviews in one file and have it read on gh-pages. If this is what you wish can please provide some instructions?

cmungall:
Add the review to the existing md file for each ontology

wdduncan:
Ok. The review information can be included in the `.md` file in the `ontology` directory.   
This will enable review information to be added directly in the `.md` file. Is that fine or do you want a different workflow?  
I will create a new PR when ready. I need some time to figure out how to read the review info from the ontology `.md` file and output on the Completed Reviews page.

cmungall:
> This will enable review information to be added directly in the .md file. Is that fine or do you want a different workflow

That is the idea. All in one place!

> I will create a new PR when ready. I need some time to figure out how to read the review info from the ontology .md file and output on the Completed Reviews page.

The beauty of the system is that you don't need to do this. Everything is compiled into _config.yml, the results of which are available to all liquid templates on the site

cmungall:
See https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/README-sitedev.md for details on how the site works.. LMK if anything isn't clear, and/or make PRs to the docs to make it easier for others who also want to do development!

Issue #1019, opened 2019-08-09
Principle #1 automated validation


beckyjackson:
FP 1 - Open

Automated checks:
1. The ontology must have a license
2. The license must be open

Mechanisms:

We already check for a license in the metadata validation. If there is a missing license, that check fails. We also issue a warning if the license is not a valid open license (CC0 or CC-BY).

This check would slightly change from the metadata validation in that it is required that the license is open. So if the ontology has a license other than one of those, this check will fail.

beckyjackson:
The ontology file should also have a license in the file. This license should match the license in the registry file. The ontology should use the property http://purl.org/dc/terms/license.

If the license is not CC0/CC-BY, it receives a registry warning in the registry metadata validation. This check should be the same (WARN on non CC0/CC-BY).

balhoff:
A lot of the ontologies with red Xs in the dashboard ("missing ontology license") do have a license. But they are using the `http://purl.org/dc/elements/1.1/rights` property. Could you check for this one and issue a warning on the property instead of a missing license error?

jamesaoverton:
@balhoff I believe that using the wrong predicate is just a WARN, which seems right to me: https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/util/principles/fp_001.py#L25

balhoff:
I think the issue is that `dc:rights` is not looked for as a "bad" property here: https://github.com/OBOFoundry/OBOFoundry.github.io/blob/47822c41a8ea7fda406063b291f21c4b9d09250c/util/principles/fp_001.py#L369-L372

jamesaoverton:
Ok, thanks. I think we should add `dc:rights` as a predicate to check and WARN on. Would you mind making a PR?

matentzn:
@beckyjackson if you could check this (docs with examples on how to format the licenses correctly): 

https://github.com/INCATools/ontology-development-kit/blob/master/docs/License.md

I think its correct; I am using it for updating all the ontology I work with. If this is useful, we can move it here or to your own documentation; I dont mind where it lives!

allysonlister:
In a slight variation on this ticket, SWO fails because I have my license in a sentence, e.g. "swo is provided under a creative commons attribution 4.0 international (cc by 4.0) license (https://creativecommons.org/licenses/by/4.0/)." Are you saying OBO is requesting that we just have the URL to the license? Or would such a sentence still be acceptable? Thanks!

[...]

PR #1120, opened 2020-02-25
Additional license properties to warn on for principle 1

balhoff:
I added two additional properties that ontologies may be using for license info. This way we can warn about standardizing rather than stating the ontology doesn't have license info (see discussion in #1019).

### DIFF
--- a/principles/checks/fp_001.md
+++ b/principles/checks/fp_001.md
@@ -128,14 +128,14 @@ class OpenValidator():
        annotations = ontology.getAnnotations()
        license = dash_utils.get_ontology_annotation_value(annotations,
                                                           license_prop)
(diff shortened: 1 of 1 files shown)


balhoff:
@beckyjackson I didn't actually run this code (not sure how to, sorry!). Would you be able to check it?

jamesaoverton:
Thanks for the PR @balhoff, but could you please move it here?

https://github.com/OBOFoundry/OBO-Dashboard

I know it's confusing -- we'll get better organized.

balhoff:
No problem @jamesaoverton, moved to https://github.com/OBOFoundry/OBO-Dashboard/pull/8.

Issue #451, opened 2017-07-28
Annotate ontologies with their usages


cmungall:
One of the principles is having a plurality of independent users. We should annotate this in the yml/md so we can auto-evaluate.

Proposed structure:

```
usages:
   - user: <URL of resource or agent>
     description: <text description of usage (not of resources or agent) >
     type: <e.g. analysis, annotation>
     examples:
        - url: <URL e.g. resolves to a database query with class from ontology>
          description: <description of example>
```

we should exclude owl:imports usages here as this is easily queryable separately.

we should exclude trivial usages. For example, bioportal NER uses all ontologies, we would not mention this.

jonquet:
Hi, if you look for a smeantic Web vocabulary to express this property, in MOD1.2
https://github.com/sifrproject/MOD-Ontology
and
http://www.lirmm.fr/~jonquet/publications/documents/Article_MTSR-2017_MOD1.2.pdf 

We (with @biswanathdutta) recomand to use omv:knownUsage property. An alternative could be vann:usageNote
This is work in progress that shall be discussed now on the RDA VSSIG: 
https://vocabulary-services.slack.com/messages/C7GJM7U57/

cmungall:
Great suggestion @jonquet - however, knownUsage is a Datatype property. If you look at our yaml/json-ld you'll see we're using an n-ary pattern. We connect an ontology to a usage object, then that usage object has links to objects and literals

E.g.

```
id: cl
usages:
 - user: http://fantom5-collaboration.gsc.riken.jp/
   type: annotation
   description: FANTOM5 is using Uberon and CL to annotate samples allowing for transcriptome analyses with cell-type and tissue-level specificity.
   examples:
    - url: http://fantom.gsc.riken.jp/5/sstar/CL:0000540
      description: FANTOM5 samples annotated to neuron
```

We can't map usages to knownUsage in the context file, we'd need an equivalent ObjectProperty

thoughts?

cmungall:
Fixed with #453 - @jonquet you can open another ticket for the mapping to properties in the json-ld context file

PR #1136, opened 2020-02-28
chebi metadata

cmungall:


cmungall:
See #451

jamesaoverton:
This seems fine to me, but I don't know much about ChEBI.

PR #1154, opened 2020-03-11
Update eupath.md

zhengj2007:
Add domain and users information, see trackers: https://github.com/VEuPathDB-ontology/VEuPathDB-ontology/issues/43 and https://github.com/VEuPathDB-ontology/VEuPathDB-ontology/issues/44

### DIFF
--- a/ontology/eupath.md
+++ b/ontology/eupath.md
@@ -17,9 +17,15 @@ products:
 title: VEuPathDB ontology
 tracker: https://github.com/VEuPathDB-ontology/VEuPathDB-ontology/issues
 usages:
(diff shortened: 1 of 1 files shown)


jamesaoverton:
I think there should be only one `user` URL, with multiple `examples`, see http://obofoundry.org/principles/checks/fp_009

zhengj2007:
> I think there should be only one `user` URL, with multiple `examples`, see http://obofoundry.org/principles/checks/fp_009

Updated according to the instructions on http://obofoundry.org/principles/checks/fp_009

PR #1157, opened 2020-03-13
Update gno.md

edwardsnj:


jamesaoverton:
Hi @edwardsnj. The last part of what you've added is PURL configuration (`entries`) and should go in this file instead: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/gno.yml

Let us know if you need a hand.

edwardsnj:
Thanks for the hint. I've removed those problematic edit. Should I make a new pull request or is the revised commit OK?

jamesaoverton:
This is good. Thanks!

Issue #1007, opened 2019-08-09
Principle #11 automated validation


beckyjackson:
FP 11 - Locus of Authority

Automated checks:
1. Does the contact have a name as `label`?
1. Does the contact have an email?
2. Does the contact have a GitHub?

Mechanism:

Checks 1 and 2 are already implemented in the metadata validation that runs across all ontology md files. If the `label` or `email` are missing from the `contact` tag, the check will fail. Additionally, the `email` must be an actual email address. Currently, `github` is not a required property.

As we get more `github` values in the `contact` field, we should consider moving `github` to a required field. I propose to move missing `github` properties to a warning (it can later be changed to be an error).

PR #1161, opened 2020-03-17
Removing mention of 2ary 3ary contacts

cmungall:
This text makes no sense as we only allow a single contact, see #1007 and http://obofoundry.org/principles/checks/fp_011


nataled:
We don't restrict the number of contacts, just the type of contacts. Each must be an individual. I suggest removing the sentence as you say, but adding this sentence to the end of the paragraph:

Note that alternative contacts can be designated in case the primary contact is unavailable. However, as for the primary contact, each alternative contact must be an individual.

jamesaoverton:
@nataled I'm pretty sure that all the registry entries have exactly one contact (with the exception of orphaned projects, which have no contact). The JSON schema and other code/tests expect exactly one contact. So there's a disconnect.

nataled:
The automated checks are supposed to follow the principle, not the other way around. This principle has been discussed by the EWG and vetted by the Ops committee. If we want to change it, it should be done through the proper discussion channels.

jamesaoverton:
I hear what you're saying @nataled, but unlike the recent dashboard work, the tooling around a single contact goes back to our migration to GitHub, around July 2015, and all 200+ registry entries conform to it. @cmungall can correct me if I'm mistaken about that.

nataled:
I don't doubt you. Just pointing out that we don't make ad hoc decisions of the type that will actually change the principle. Even though there are good reasons for keeping it as is, I don't personally object to the change proposed--but I do object to changing it without discussion.

cmungall:
James is correct

cmungall:
In fact it predates that. Always been single contact.

mellybelly:
Registry entries have one contact. I had emailed all of the contacts a number of years ago because we had a lot of listserv entries/multiple entries that were dead, at that time we confirmed/updated all the entries in the registry as one single individual. The rationale was that given that it was a person, we'd be able to follow up more readily than a dead listserv. If we want to add other fields then that is a different issue. 

+1 on the PR.

yongqunh:
Having alternative contacts is important in many cases when the primary contact is unavailable, esp. for the collaborative development of ontology among different groups.

nlharris:
Sounds like it would be good to have two fields, a "Primary contact" field (with only one person) and an "Alternate contact(s)" field.

yongqunh:
yes

cmungall:
folks, there is a separate ticket for this that has been open since 2015:
https://github.com/OBOFoundry/OBOFoundry.github.io/issues/73

Let's continue the discussion there. For now, this PR should be merged, to
reflect reality.

balhoff:
In discussion at OBO Operations Committee meeting 2020-03-24, we suggest appending to the proposed text: "Note that alternative contacts can be designated in case the primary contact is unavailable. However, as for the primary contact, each alternative contact must be an individual." 

We also agreed that the metadata schema should be extended with a new field for 'alternative contacts' (or similar).

yongqunh:
That’s good. Thanks!

pbuttigieg:
Operations Committee call today: as this requires quite a few changes to tooling to, the question is more one on who will act on this principle change.

nlharris:
Can this be merged?

cmungall:
darren approved it, so yep!

nlharris:
"Draft pull requests cannot be merged"

PR #1166, opened 2020-03-24
Get rid of Legacy menu

### LINKED ISSUE #25: Decide what to do about legacy pages
The old http://obofoundry.org site had a lot of pages that haven't been touched since 2007ish

These are currently in the navbar under "LEGACY"

I suggest removing forever

### LINKED ISSUE #883: update and rearrange nav bar content
It would be terrific to have a Resources tab on the nav bar. There is a draft file here: https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/resources.md
This could grow and be split out into Tutorials, Educational manuscripts, Blogs, Ontology Browsers, and Ontology-related Tools.

In addition, I think the nav bar could do with some rearrangement, suggested ordering:

Ontologies
Principles
Participate
Citation
Resources
FAQs
About
 (here move the OBO working groups to Participate tab?)

Time to retire Legacy?

also some of the important things, such as setting up your purls, are buried in the operations page: http://www.obofoundry.org/docs/OperationsCommittee.html
Maybe there should be a Requests or Configuration tab
Also Setting up Protege could move to the Resources page. 

I would be happy to try to make some rearrangements for review if people are game.

nlharris:
as per @cmungall's direction in #883 
fixes #883 
fixes #25

### DIFF
--- a/_includes/navbar.html
+++ b/_includes/navbar.html
@@ -118,21 +118,6 @@
           </ul>
         </li>
 
-        <li class="dropdown">
-          <a href="#" class="dropdown-toggle" data-toggle="dropdown" role="button" aria-haspopup="true" aria-expanded="false">Legacy <span class="caret"></span></a>
-          <ul class="dropdown-menu">
-            <li><a href="/legacy/literature.html">Literature</a></li>
-            <li><a href="/legacy/participate.html">Participate</a></li>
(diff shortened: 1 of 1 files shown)


nataled:
Found only one page with appreciable content/links. Noted here for posterity:

https://code.google.com/archive/p/obo-foundry-operations-committee/wikis

PR #1181, opened 2020-04-13

ramonawalls:
Under requirements, it says the requestor must be on obo-discuss, but the link is a mailto link. I changed it to a link to the google group, so it is easier to find. 

I also updated the example request to one for ECTO, since the old one (for OHMI) was pointing to Source Forge.


jamesaoverton:
This is a good improvement. Thanks @ramonawalls!

Issue #1170, opened 2020-04-03
The Amphioxus Development and Anatomy Ontology (AMPHX)


hescriva:
## Ontology title
The Amphioxus Development and Anatomy Ontology

## Requested ID space
AMPHX

## Ontology location
https://github.com/EBISPOT/amphx_ontology
https://raw.githubusercontent.com/EBISPOT/amphx_ontology/master/amphx.owl

## Contact person
Name: Hector Escriva
Email address: hescriva@obs-banyuls.fr
GitHub username: @hescriva

## Issue tracker
https://github.com/EBISPOT/amphx_ontology/issues

## Ontology license

- [ ] CC0 (public domain)
- [X] CC-BY (version 3 or later)
- [ ] Other: please specify

## Available ontology formats
OWL, OBO

## What domain is the ontology intended to cover?
The ontology should be used in the domain of Developmental Biology and Evolution of the different species of the cephalochordate genus Branchiostoma ([NCBITaxon:7737](https://www.ebi.ac.uk/ols/ontologies/ncbitaxon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FNCBITaxon_7737)).

## Related OBO Foundry ontologies
UBERON

## Intended use/related projects
This ontology is intended to be used for description of gene expression in amphioxus (e.g. Insitus, RNA-seq). The ontology was created in the context of the European project CORBEL (https://www.corbel-project.eu/home.html), and used in the database MARIMBA (http://marimba.obs-vlfr.fr/home)

## Data source
The ontology was created through literature review describing the different developmental stages, embryonic structures and adult structures and tissues from three different amphioxus species (B. lanceolatum, B. floridae and B. belcheri)

cmungall:
The 2 week request for comments period has passed and there have been no objections raised, so the ontology can be registered. Congratulations!

@matentzn are you taking care of making the pull request?

Some comments/requests. Acceptance into OBO is not contingent on these:

 - map to uberon where you have species-specific versions of uberon classes
 - consider making the ontology is-a complete (I'm guessing you are making this with oboedit which will one day stop working...)
    - e.g. ventral fin should be is-a fin
 - are you sure your relationships follow the all-some pattern? E.g. you have an axiom "dorsal fin SubClassOf part-of some whole embryo". But this isn't true as these finds are also found in adults
 - avoid capitalization except for proper names
 - you have cycles, e.g. Gut part-of Gut
[...]

PR #1190, opened 2020-04-21
Create amphx.md

matentzn:
See also: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1170


jamesaoverton:
Can you please add a short description (Markdown) after the YAML header. I think it will look weird if that's missing.

PR #1193, opened 2020-04-22
marking RNAO as inactive

### LINKED ISSUE #867: Status of RNAO project
Hi!

Currently, RNAO is listed as active on the OBO Foundry, but it has not been updated in 9 years, and contains many errors in both the OBO (unescaped quotes, odd relation semantics) and the OWL product (cardinality errors mostly). It can't be opened by `robot` because of these issues, so I tend to think it is virtually unusable. Should it be listed as *inactive* ?

cc @cmungall 
cc @AntonPetrov

nlharris:
as agreed
fixes #867
(actually i'm not sure--is there anything else that needs to be done to mark RNAO inactive, or will this do it?)

### DIFF
--- a/registry/ontologies.yml
+++ b/registry/ontologies.yml
@@ -4197,7 +4197,7 @@ ontologies:
     title: The pathway ontology - updates and applications.
   title: Pathway ontology
   tracker: https://github.com/rat-genome-database/PW-Pathway-Ontology/issues
-- activity_status: active
+- activity_status: inactive
   browsers:
   - label: RNAO
     title: RNA Ontology jOWL Browser


nlharris:
so now that chris has approved, do i go ahead and merge?

is this the only thing that needs doing to mark the ontology inactive?

jamesaoverton:
Sorry, this wasn't quite right, and I clicked "merge" too soon.

@nlharris Changes to the registry must be made to the project files in the `ontologies/` directory. The `ontology/` files are used to generate the files in the `registry/` directory, and so manual edits to `registry/` files will be overwritten.

I fixed this in #1207 by editing `ontologies/rnao.md`.

PR #1210, opened 2020-05-12
Proposed changes to versioning principle

cmungall:
Added additional clarifications on how "version identifier" related to "version IRI"

Added recommendations for versionInfo, related to https://github.com/ncbo/bioportal-project/issues/168

comments sought from bioportal team, @jvendetti @graybeal


balhoff:
@cmungall for OBO `data-version`, for ontologies that have ended up with the `/releases/` prefixing the release date, I'm not sure there is a way to not have the `data-version` be something like `releases/2020-05-11` rather than `2020-05-11`. It is automatically generated from the `owl:versionIRI` in the OWL API translation using a pattern based on the OBO identifier space.

But, if we put the clean date value in `owl:versionInfo`, BioPortal can grab it from there from either OWL or OBO formats.

pbuttigieg:
In the OBO Operations call today, this PR was positively received.
The principle update will have to go through the Editorial WG for final approval.

It was recommended that an e-mail be sent to obo-discuss, explaining the new requirements and referencing some documentation or other implementation guidance.

nlharris:
> It was recommended that an e-mail be sent to obo-discuss, explaining the new requirements and referencing some documentation or other implementation guidance.

Did this happen?

wdduncan (changes requested):
We need to use consistent spelling for 'artifact'. Please change 'artefact'.

wdduncan (changes requested):
Remove content:  
Version IRI is a part of the OWL2 specification. While this field is optional for OWL ontologies, OBO imposes stricter standards and requires this for releases.

We only need to state each OBO ontology needs an version IRI.

wdduncan (approved):

PR #1219, opened 2020-05-22
Removed out of date info on license

cmungall:
this was inconsistent with the license declared in the metadata

I chose to remove it altogether since it's already shown on the navbar but feel free to close/edit and replace with a PR where it is in sync

Also IMHO the text "To use OBI, remember the licensing terms" should be more welcoming in the open OBO spirit but YMMV


nlharris:
@cmungall! is this the best possible use of your time? 😀
(the PR looks fine to me)

matentzn:
hahahha good one @nlharris

PR #1230, opened 2020-07-07
Update fp-004-versioning.md

wdduncan:
Remove Note section at top.  
Remove empty Counter-example section.


matentzn:
These note sections you are removing, was this discussed at a call, or do you just think they are unnecessary? If the former, we can merge this, if the latter, I guess we would need to raise this at a call? I agree with you though, the sentence is unprofessional.

lschriml (approved):

PR #1260, opened 2020-09-01
Update mp.md

sbello:
adding link to mp-base.owl file in product list

### DIFF
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -20,6 +20,7 @@ products:
   - id: mp.owl
   - id: mp.obo
   - id: mp.json
+  - id: mp-base.owl
 browsers:
   - label: MGI
     title: MGI MP Browser
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -20,7 +20,13 @@ products:
   - id: mp.owl
   - id: mp.obo
   - id: mp.json
-  - id: mp-base.owl
+    title: MP in obographs JSON format
+    description: "For a description of the format see https://github.com/geneontology/obographs."
+    page: https://github.com/obophenotype/mammalian-phenotype-ontology/releases/tag/current
+  - id: mp/mp-base.owl
+    title: MP Base Module
+    description: "The main ontology plus axioms connecting to select external ontologies, excluding axioms from the the external ontologies themselves."
+    page: https://github.com/obophenotype/mammalian-phenotype-ontology/releases/tag/current
 browsers:
   - label: MGI
     title: MGI MP Browser
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -26,7 +26,7 @@ products:
     description: "A direct translation of the MP (OWL edition) into OBO format."
     page: https://github.com/obophenotype/mammalian-phenotype-ontology/releases/tag/current
   - id: mp.json
-    title: MP in obographs JSON format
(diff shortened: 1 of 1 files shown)


matentzn:
@cmungall @jamesaoverton @balhoff 
This raises the question for me whether base files should be registered directly? If we do that we should do it everywhere.

sbello:
@matentzn I added this in response to @cmungall comments during the COB workshop. If we don't want to list these here that is fine. On the other hand should we list all the products of the MP release? I worry slightly that this will confuse users as to which is the preferred release file for using the ontology? Currently on the MGI ftp site we have only the 3 versions of the mp file (obo, owl and json) as these are what we would prefer most users of the ontology to consume. The other files really are for a smaller subset of users.

matentzn:
You were totally right and amazingly proactive to do this :) I just want to get a sense whether `products` is meant just for the primary products (mp.owl, mp.obo, mp.json) or whether we should include all products we consider primary; Nothing for you to worry about - unless of course you have an opinion!

jamesaoverton:
This would be a change. I don't know what the corresponding PURL for `mp-base.owl` (specified here) would be. So far we have been scrupulous about reserving the top-level name space for http://purl.obolibrary.org/obo/{idspace}.{format}

matentzn:
All the base files are currently tagged and available as http://purl.obolibrary.org/obo/{idspace}/{idspace}.{format}; I would not want to change that to http://purl.obolibrary.org/obo/{idspace}.{format}! Does this mean mp-base is conceptually not a `primary` product?

sbello:
So should this have followed the GO pattern? Although oddly hovering over the link on the GO page shows a link ...go/go-base.owl but the link resolves http://current.geneontology.org/ontology/go-base.owl (although clicking on this link causes the tab to crash, maybe it works if you just save the link). If the text needs to be changed to mp/mp-base.owl to make the distinction that this is not a 'primary' product that is fine with me.

matentzn:
Yeah lets wait to here from @cmungall about this - I dont think this is important to MP users; mp-base should really only be relevant for people that know what they are doing! And they know where they can find it!

sbello:
Fine with me. Do you want me to close the pull request?

matentzn:
Lets leave it open for a few days and see whether others have an opinion!

balhoff:
I think all products should be listed, but described. There is a lot of info on the different products in the GO listing: http://obofoundry.org/ontology/go.html [[source]](https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/ontology/go.md)

I think for mp-base it should `mp/mp-base.owl`.

matentzn:
@balhoff @sbello I have now [adjusted the markdown](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1260/files) to Jims suggestions. What do you think? Other than `taste` does anyone have any arguments against this? @cmungall @jamesaoverton

sbello:
Thanks @matentzn ! I think it looks good.

cmungall:
Need any more input from me?

matentzn:
If you are fine with this general approach no :)

jamesaoverton (approved):
I think `mp/mp-base.owl` is good, and so are the descriptions.

matentzn (approved):

Issue #1116, opened 2020-02-13
Proposed efficient criteria for evaluating registration requests


cmungall:
PROPOSAL:

Criteria for registration in OBO. The goal is to formalize what is
currently an ad-hoc process, to filter out ontologies that are clearly
out of scope for OBO. It is intentionally weaker than full
conformance, because (a) we want to allow registration of imperfect
ontologies (b) partial conformance is easier to quickly and
objectively check.

The checks are designed to be fast and objective to evaluate.
In future we may be able to run the dashboard on a new ontology to help (https://github.com/OBOFoundry/OBO-Dashboard/issues/2) but as written these checks are trivial for anyone with basic ontology skills to evaluate quickly.

Note this does not change the 2 week rule. Anyone from the OBO
community is free to object to the ontology on the basis of any of
these checks. However, if no objections are raised in the 2 week
period, then the ontology becomes registered. Once registered an
ontology is never unregistered.

We use ISO language here for SHOULD/MUST. Anyone is free to raise an
objection based on a SHOULD but this is informative and
non-blocking. If a MUST is clearly violated then this is basis for
exclusion/delaying registration.

 - p1 open:
     - MUST be conformant with this principle
 - p2 format:
     - MUST be conformant with this principle
 - p3 URIs:
     - MUST be conformant, with the caveat the prefix is not yet registered
 - p4 versioning:
     - The ontology MUST have a versionIRI
 - p5 scope:
    - IF there is significant conceptual overlap or matching labels
      THEN there MUST be computable linkages (one of: skos, obo:xref, owl)
      these SHOULD be reasonably comprehensive
      AND there SHOULD be a written plan (e.g ticket) for coordinating with overlapping ontologies
 - p6 definitions:
    - SHOULD conform to principle
    - MUST include some definitions
 - p7 relations
    - SHOULD conform
 - p9 users:
    - SHOULD state existing users and assumed users
 - p10 collaboration:
    - The submitter SHOULD be committed to collaboration
 - p11 locus: 
     - MUST comply
 - p12 naming:
    - All classes MUST have labels
 - p16 maintenance
    - The ontology MUST have an open tracker where anyone from the community can submit
    - The maintainers SHOULD be responsive to issues in this tracker

jamesaoverton:
Thanks @cmungall. I think this is a good level of conformance to require from new projects, and it's clearly expressed.

matentzn:
> MUST include some definitions

[...]

PR #1266, opened 2020-09-08
Creating Minimum Ontology Registration Request Checklist

matentzn:
This checklist should contain more or less the consensus of #1116. Comments welcome @cmungall @jamesaoverton @wdduncan @nlharris @balhoff


nlharris (approved):
Looks good to me!

cmungall (approved):

cmungall:
discuss on next call?

matentzn:
On agenda already!

docs/RegistrationChecklist.md
```diff
@@ -0,0 +1,58 @@
+# Minimum Ontology Registration Request Checklist
+
+The _Minimum Ontology Registration Request Checklist_ (MORRC) is intended to provide a first pass to fairly evaluate ontology registration requests to the OBO foundry. 
+The goal is to formalize what is currently an ad-hoc process, to filter out ontologies that are clearly out of scope for OBO. It is intentionally weaker than full
+conformance, because (a) we want to allow registration of imperfect ontologies (b) partial conformance is easier to quickly and objectively check.
+
+Note this does not change the 2 week rule. Anyone from the OBO community is free to object to the ontology on the basis of any of these checks. 
+However, if no objections are raised in the 2 week period, then the ontology becomes registered. Once registered an ontology is never unregistered.
```
matentzn, line 8: In the Operations call, we discussed that two week the rule is too tight. I suggest changing to "two weeks from next operations call" or "four weeks rule".

docs/RegistrationChecklist.md
```diff
@@ -0,0 +1,58 @@
...
+- Principle 8: [Documentation](http://obofoundry.org/principles/fp-008-documented.html)
+  - The ontology SHOULD be conformant with this principle
+- Principle 9: [Documented Plurality of Users](http://obofoundry.org/principles/fp-009-users.html)
+  - The submitter SHOULD state existing users and assumed users
+- Principle 10: [Commitment To Collaboration](http://obofoundry.org/principles/fp-010-collaboration.html)
+  - The submitter SHOULD be committed to collaboration
+- Principle 11: [Locus of Authority](http://obofoundry.org/principles/fp-011-locus-of-authority.html)
+  - The ontology MUST be conformant with this principle
+- Principle 12: [Naming Conventions](http://obofoundry.org/principles/fp-012-naming-conventions.html)
+  - All classes MUST have labels
```
matentzn, line 55: What about obsolete ones?

nlharris: Maybe we need to add text something like, "To be added to the OBO Foundry, ontologies must meet these requirements. Ontologies that are already listed that do not meet these requirements will be moved to 'inactive' or 'obsolete' status."

docs/RegistrationChecklist.md
```diff
@@ -0,0 +1,58 @@
...
+
+The checklist is expected to evolve over time; anyone can contribute to the discussion [here](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1116).
+
+## Checklist
+
+- Principle 1: [Open](http://obofoundry.org/principles/fp-001-open.html)
+  - The ontology MUST be conformant with this principle
+- Principle 2: [Common Format](http://obofoundry.org/principles/fp-002-format.html)
+  - The ontology MUST be conformant with this principle
+  - The ontology MUST be parseable in Protégé (or OWLAPI) and be logically consistent with a standard OWL reasoner such as ELK or HermiT
```
matentzn, line 26: Technically the right phrase would be `consistent and coherent`, as we not only want consistency, but also no unsatisfiable named classes (coherency).

docs/RegistrationChecklist.md
matentzn: I meant more, what about obsolete classes; do they have to have labels? I would probably not go into too much depth around when to obsolete an ontology right in this checklist, but we can raise a separate issue!

nlharris: Ah, sorry, I misinterpreted.

docs/RegistrationChecklist.md
```diff
@@ -0,0 +1,58 @@
...
+- Principle 2: [Common Format](http://obofoundry.org/principles/fp-002-format.html)
+  - The ontology MUST be conformant with this principle
+  - The ontology MUST be parseable in Protégé, ROBOT or OWLAPI and be logically consistent with a standard OWL reasoner such as ELK or HermiT
+- Principle 3: [URI/Identifier Space](http://obofoundry.org/principles/fp-003-uris.html) 
+  - The ontology MUST be conformant with this principle, with the caveat the prefix is not yet registered
+  - The requested namespace MUST be available (not already used by another ontology)
+- Principle 4: [Versioning](http://obofoundry.org/principles/fp-004-versioning.html) 
+  - The ontology MUST have a versionIRI
+- Principle 5: [Scope](http://obofoundry.org/principles/fp-005-delineated-content.html) 
+  - IF there is significant conceptual overlap or matching labels THEN there MUST be computable linkages (one of: skos, obo:xref, owl) which are reasonably comprehensive
```
cmungall, line 33: "IF there is significant conceptual overlap" --> "IF there is significant conceptual overlap to ontologies in OBO"

docs/RegistrationChecklist.md
cmungall: This should read: all non-obsolete classes must have labels

jamesaoverton:
My view is that this is a big step forward, and that we should move forward with it. We can always make refinements later.

nlharris:
Oh, oops, I meant to ask before merging this. Hope it was ok to merge.

PR #1269, opened 2020-09-11
Testing to set pyrsistent==0.16.0

matentzn:
Problem with:
https://travis-ci.org/github/OBOFoundry/OBOFoundry.github.io/builds/726221090

### DIFF
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -41,4 +41,4 @@ mailing_list: https://groups.google.com/forum/#!forum/phenotype-ontologies-edito
 activity_status: active
 ---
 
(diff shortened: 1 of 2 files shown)


matentzn:
@jamesaoverton this pr basically has the fixed `pyrsistent==0.16.0` as discussed on slack (indeed, they just upgraded one of these days to version introducing breaking changes!) https://pypi.org/project/pyrsistent/0.16.0/#history

Now they say version has been yanked; but its probably better to fix the version just in case.

jamesaoverton (approved):

jamesaoverton (approved):

PR #1279, opened 2020-09-22
Update Membership.md

ddooley:


nlharris:
Hmm, it seems I can request myself as a reviewer but then I can't actually do the review?

Anyway, this change is fine. See also #1279 suggesting a more general review of the membership list.

matentzn (approved):

PR #1300, opened 2020-09-25
converted md markup to html

nlharris:
Apparently you can't include markdown in a .html! Who knew?


nicolevasilevsky (approved):
looks good!

jamesaoverton:
It is possible to include Markdown in HTML if we change some settings, but it's probably better more clear to avoid that. If you're making these edits using the GitHub website, then there should always be a Preview button for checking Markdown content before you commit.

PR #1307, opened 2020-09-25
unsatisfiable and non-obsolete classes, overlap with other OBO ontologies (ontology review)

matentzn:
The changes were noted in #1266 but not implemented yet, so here it is.


nlharris (approved):
I only reviewed the text that was changed--looks good to me!

wdduncan (approved):

PR #1308, opened 2020-09-25
Delete ancient about.md page

matentzn:
see #1287 

@nlharris You are totally right, this is seems like an ancient, totally irrelevant page which should be deleted. Well spotted!

@cmungall @jamesaoverton


jamesaoverton:
Yeah, that's a weird page and should be removed.

jamesaoverton (approved):

PR #1334, opened 2020-10-06
OLD: Update fp-001-open.md

matentzn:
This is an old, unmerged branch. We will merge or delete this.

@nataled


nataled:
The indicated change is already reflected on the public site. This can be deleted (I don't know how).

matentzn:
Deleting as per request. Thanks @nataled

PR #1344, opened 2020-10-26
Update context map generation in processor.py

matentzn:
Without this, entires like DPO will never make it into the context map, if the preferred prefix is shared with another ontology:

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/dpo.md

This kind of change could be huge - I don't know now how to evaluate/test it.

### DIFF
--- a/util/processor.py
+++ b/util/processor.py
@@ -92,8 +92,9 @@ def extract_context(ontologies, args):
   prefix_map = {}
   for obj in ontologies:
     if has_obo_prefix(obj):
+      oid = obj['id'].upper()
       prefix = obj.get('preferredPrefix') or obj['id'].upper()
-      prefix_map[prefix] = "http://purl.obolibrary.org/obo/" + prefix + "_"
+      prefix_map[oid] = "http://purl.obolibrary.org/obo/" + prefix + "_"
 
   print(dumps({'@context': prefix_map}, sort_keys=True, indent=4, separators=(',', ': ')))
 


jamesaoverton:
This PR looks dangerous, so I marked it as a draft to avoid an accidental merge.

I'm going to ask @cmungall to review instead of me. I don't understand the intent of 'preferredPrefix' or how the 'FB*' names are supposed to work.

matentzn:
Agreed. Without it, we will have to find an alternative strategy to extract a base for the dashboard - which we should have to anyways. Think of some ontologies that have multiple base iris.

cmungall (commented):
see alsohttps://github.com/bioregistry/bioregistry/issues/46

nlharris:
What do we want to do with this?

matentzn:
This should never be merged, but it should be dealt with - I was thinking of leaving this draft pull request just as a reminder when the day comes and someone donates 1 million to the OBO foundry to fix biomedical semantics..

nlharris:
Haha, and what would we do with the $1M, clone you? 😀

matentzn:
Hire a full-time developer just for OBO tools & infrastructure, for 4 years, and a half-time community manager that organises events, does outreach and provides trainings.. Something like that. We could even pay some hosted services like the ultimate OBO knowledge graph (and pay for its standardisation programme).

nlharris:
We should write a grant proposal. Seriously. Funding continued support and outreach for OBO would be a bargain for the funding agency, since there's already so much infrastructure and knowledge embedded in the project.

matentzn:
I agree.. I would love to contribute to something like that!

matentzn:
This is now obsolete with the way we do preferred prefixes.

PR #1349, opened 2020-11-02
Regenerate files

jamesaoverton:
Update generated files with recent changes


_config.yml
```diff
@@ -6845,45 +6941,45 @@ ontologies:
     label: Zea
   title: Maize gross anatomy
 principles:
-- id: fp-000-summary
-  layout: principle
-  title: Overview
-- id: fp-001-open
+- id: fp-004-versioning
```
matentzn, line 6944: Somehow the order of the principles got randomised, is this a problem?

jamesaoverton: That's annoying. I think I fixed it.

matentzn (approved):

PR #1350, opened 2020-11-03
Adding link to slack channel

cmungall:


matentzn:
Weird. I would have thought this would violate the foundry schema! Seems I was wrong!

matentzn (approved):

cmungall:
The schema is very lenient and open. I think we need to constrain it more and make it closed. I thought there was a ticket about this but it seems not.

jamesaoverton:
@cmungall: @jannahastings has been working on a merged JSON schema for the registry that has a restricted list of keys. It will include everything we use right now but complain about new keys (until they're added to the schema). It's almost ready to go. You can see this PR #1351 she made yesterday, but there's a bit more work to do to make sure all the current registry entries pass.

jannahastings:
@jamesaoverton @cmungall Yes (sorry) I deleted my first PR and made another with all the changes needed to pass the Travis build. But will we need to add another field for this "link to slack channel"?

matentzn:
@jannahastings so if we want to make changes to the schema, what is required - just editing `util/schema/registry_schema.json` or do we also have to meddle with util/validate-metadata.py?

jannahastings:
@matentzn Just editing the json schema. The change to validate-metadata.py is just to support having a single merged schema as opposed to separate schema files, nothing about the content.

matentzn:
Thank you! :)

PR #1353, opened 2020-11-06
Adding merged registry schema file

jannahastings:
Adding a merged registry metadata JSON schema file, with added check that there are  no additional fields, and added metadata for giving helpful messages within the Metadata Editor. Also changes to validate-metadata.py needed to support validation with a merged schema where the error levels are specified per field, not per overall schema.


jamesaoverton:
@cmungall and @OBOFoundry/tech: Please take a look at this unified and revised JSON Schema file for the OBO registry. The plan is to replace the dozen small JSON Schema files in https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/util/schema

Travis checks show that all current registry entries are passing. (We will add the new `slack` key to this schema.)

Benefits of the new schema:

- all in one place, instead of separate schema files
- whitelist of allowed properties -- this is something we've wanted for a while, and will help prevent typos from getting into the registry
- better documentation of the most important keys
- still uses the current approach to ignore most errors for obsolete projects

@jannahastings and I developed this new schema to better support the new OBO Metadata Editor (coming soon!). For that purpose, we added "suggest" and "suggestion" fields. In the editor, fields that are not marked "suggest: false" will pop up as auto-completes with the "suggestion" string as the default.

On this branch we will delete the dozen old schema files and show that things still work.

I think this change is a clear win, but if you have questions or concerns please don't hesitate to raise them! Otherwise I plan to merge this after double checking with OBO Operations Committee on the November 17 call.

matentzn:
I think this is really great! Thank you for the effort! Just having this file alone is really helpful understanding our metadata better, and I think its a good move!

cmungall (approved):
This is great!

jamesaoverton:
Still to do:

- [ ] delete old schema files that we're replacing
- [ ] remove the loop over schema files from `util/validate-metadata.py`
- [ ] add `slack` to new schema
- [ ] update this branch, make sure all tests pass 
- [ ] merge

PR #1363, opened 2020-11-18
Fix and regenerate list of ontologies

jamesaoverton:
Fixed a bug with the recent schema update, fixed some warnings in registry files, and regenerated the ontology lists.


matentzn (approved):
Excellent! All changes to ontology markdown files make sense.

Issue #1008, opened 2019-08-09
Principle #9 automated validation


beckyjackson:
FP 9 - Documented Plurality of Users

Automated checks:
1. Is there a valid issue tracker?
2. Are there stated usages?

Mechanism:

We can pull the `tracker` value from the ontology YAML. We should ensure that this tracker resolves (does not return HTTP status > 400). It would be nice to check if there is activity on the tracker, but I'm not sure if that is possible at this time. I'm open to suggestions. If the ontology does not have a tracker, this check fails.

We can also look at the `usages` tag from the ontology YAML. If there are no documented usages, the ontology will get a warning. The `usages` should contain a `user` property with a valid URL. Perhaps if the URL does not resolve, we just return an info message.

We may need to standardize the `usages` tag. Currently, there are multiple ways that people have inserted usages. For example, ENVO contains two different examples of usages:
```
usages:
 - type: data-annotation
   description: "describing species habitats"
   examples:
     url: http://eol.org/pages/211700/data
   resources:
     url: http://eol.org
     label: EOL
```

```
usages:
  - user: http://oceans.taraexpeditions.org/en/
    description: Samples collected during Tara Oceans expedition are annotated with ENVO
    example:
      - url: https://www.ebi.ac.uk/metagenomics/projects/ERP001736/samples/ERS487899
        description: "Sample collected during the Tara Oceans expedition (2009-2013) at station TARA_004 (latitudeN=36.5533, longitudeE=-6.5669)"
```

I propose the following format for usages:
```
usages:
  - user: required URL
    type: optional text
    description: required text
    example:
      - url: required URL
        description: required text
```

nataled:
From the EWG discussion on this:

Partial automation possible, especially with respect to use of its terms in other ontologies and citations.

Chris M commented: The curation of usages must be manual and closely vetted by OBO Foundry.

We have usages partially curated here:

https://github.com/OBOFoundry/OBOFoundry.github.io/issues/451

Once in place the checks themselves can be automated.

Also easy to check things like GH activity. While it's conceivable that some ontologies with multiple users don't use GH it is at least a meaningful signal

jamesaoverton:
The principle says "Use of the target ontology’s term IRIs in other ontologies. This can be evidenced by linking to the other ontology that uses an ontology term IRI from this ontology" We could search for term use in other ontologies.

cmungall:
I agree with @beckyjackson's proposed standardization of the usages tag.

[...]

PR #1378, opened 2020-12-08
Implement schema for usages

### LINKED ISSUE #1371: Implement schema for usages
Issue #1179 asks to add usage information to the ontology detail pages. Before we can do that, we need to standardize that information. #1008 points out that there are different formats in the registry right now. The current schema just allows the `usages` key without any constraint:

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/util/schema/registry_schema.json#L417

So the steps are:

- [ ] implement a JSON Schema for usages based on @beckyjackson's suggestion in #1008 
- [ ] update all registry entries to conform to the new schema
- [ ] update [ontology_detail.html](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/_layouts/ontology_detail.html) layout to add the usages #1179

apmody:
Resolves #1371


util/schema/registry_schema.json
```diff
@@ -415,7 +415,37 @@
...
+			"items": {
+        			"type": "object",
+        			"properties": {
+					"user": {
+          			      		"type": "string",
+          					"format": "uri"
+       				 	},
+					"type": { "type": "string" },
+					"description": {"type":"string"},
+        				"example": {
```
jamesaoverton, line 431: Since this is an array, it should be named "examples".

util/schema/registry_schema.json
```diff
@@ -415,7 +415,37 @@
...
+						"type": "array",
+						"items" : {
+          						"type": "object",
+      							"additionalProperties" : false,
```
jamesaoverton, line 435: Indentation is inconsistent in this file.

util/schema/registry_schema.json
apmody: From the issue in #1008, it was saying to use "example".

jamesaoverton: Since there are good reasons to allow more than one example, let's go with "examples".

apmody:
Two options to show Usages. 

One is seen in commit dfe0f6e where it is in the sidebar.

mro.html with usages tag on sidebar

The other is seen in commit fb125f1 where it is under products.

mro.html with usages tag below products

_layouts/ontology_detail.html
```diff
@@ -106,16 +106,48 @@ <h2>Products</h2>
             </tbody>
           </table>
         </div>
-        
+        {% if page.usages %}
+      <div>
+          <h2> Usages </h2> <ul>
+        {% for p in page.usages %}
+          <li>
+           <b> User: </b> <a href="{{p.user}}"> {{p.user}} </a> <br/>
```
jamesaoverton, line 114: Please don't use `<br/>`, since it has no semantics. Try using `<did>` and `<dt>`, as in the sidebar.

ontology/envo.md
```diff
@@ -42,40 +42,32 @@ dependencies:
...
-      url: http://eol.org
-      label: EOL
-  - type: data-annotation
+      - url: http://eol.org/pages/211700/data
+        description: "Pseudobarbus burchelli (Tradou Redfin) is a species of bony fishes in the family Cyprinidae. They are associated with freshwater habitat. Individuals can grow to 13.5 cm. They have sexual reproduction."
+  - user: http://globalbioticinteractions.org
+    type: data-annotation
     description: "describing stomach contents"
-    datasets:
-      url: https://s3.amazonaws.com/globi/snapshot/target/data/taxa/interactions.csv.gz
```
jamesaoverton, line 55: This is getting dropped, but I think that's OK since it's not a very good example.

jamesaoverton (changes requested):
I would like one small change to the HTML, then I think it's good to go.

matentzn (approved):
Great work! Love it!

jamesaoverton (approved):

Issue #1382, opened 2020-12-10
Fail registry validation on duplicate keys


jamesaoverton:
@apmody just noticed that `envo.md` has two `usages` keys. `pyyaml` doesn't complain about this, it just overwrites. I think we should fail validation on duplicate keys.

nlharris:
It seems like this was fixed by #1384; if that's wrong, please reopen.

PR #1384, opened 2020-12-18
Implement YAML Linting

apmody:
Resolves issue #1382


ontology/ddpheno.md
```diff
@@ -8,7 +8,6 @@ contact:
 license:
   url: http://creativecommons.org/publicdomain/zero/1.0/
   label: CC-0
-tracker: https://github.com/obophenotype/dicty-phenotype-ontology/issues
```
matentzn, line 11: shouldnt be removed

ontology/cteno.md
```diff
@@ -20,15 +20,12 @@ homepage: https://github.com/obophenotype/ctenophore-ontology
...
 dependencies:
- - id: uberon
- - id: ro
+  - id: uberon
+  - id: ro
 taxon:
   id: NCBITaxon:10197
   label: Ctenophore
 tracker: https://github.com/obophenotype/ctenophore-ontology/issues
-jobs:
```
matentzn, line 29: why is this removed?

ontology/emapa.md
```diff
@@ -19,19 +19,17 @@ taxon:
   id: NCBITaxon:10088
   label: Mus
 title: Mouse Developmental Anatomy Ontology
-tracker: https://github.com/obophenotype/mouse-anatomy-ontology/issues
```
matentzn, line 22: Why removed?

matentzn (changes requested):
First of all, this is GREAT we do that, very nice :) This is an incomplete review:

I saw a bunch of tracker entries removed (see comments) and job entries; why?

jamesaoverton:
@matentzn One of the things we're checking for is duplicate keys. I think that in all the cases you point out, there was another `tracker` or `job` with the same information.

I'm also reviewing to make sure we don't drop anything.

ontology/fbbt.md
```diff
@@ -51,6 +51,8 @@ usages:
     example:
       - url: http://www.virtualflybrain.org/site/stacks/index.htm?id=FBbt_00003651
         description: "Ring neuron R2 in VFB"
+  - user: http://www.virtualflybrain.org/
+    description: VFB uses FBbt to annotate brain images
```
jamesaoverton, line 55: I think this should be multiple examples instead of multiple users.

jamesaoverton (changes requested):
There are two things I want to change, but I'll make those changes myself.

jamesaoverton (approved):
I went through this quite carefully, and I don't think anything has been lost.

The major benefit is that we've resolved many problems with duplicate keys. We've also fixed indentation, which can be a problem for interpreting YAML. A large number of changes are just removing leading and trailing whitespace, but I think that sort of clean up is also valuable to downstream consumers.

matentzn (approved):

matentzn:
This is great! Thanks for the clarification of the "lost keys" :) approved now!

PR #1389, opened 2021-01-02
Add preferred prefix to 5 ontologies

matentzn:
Not sure this is the right thing to do, but I noticed that:
1) All ExO purls are like http://purl.obolibrary.org/obo/ExO_0000062
2) The context seems also to link to the wrong thing:
https://github.com/OBOFoundry/OBOFoundry.github.io/blob/355415024a5956a45a0efd33d29a2ac30fc54300/registry/obo_context.jsonld#L63

@jamesaoverton is there a downside if we add preferredPrefix to the md? Should this PR also add a manual change to the context?


matentzn:
Same for:
- [ ] http://purl.obolibrary.org/obo/HsapDv_
- [ ]

jamesaoverton:
We discussed this today and came up with a plan. Please request my review when you're ready.

matentzn:
Aside from one ([pdumdv](https://github.com/OBOFoundry/purl.obolibrary.org/pull/718)), the idspace configs were all correct!

jamesaoverton (approved):

Issue #1393, opened 2021-01-06
Missing trackers 2020


matentzn:
- [x]	[ERO tracker](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/909): 'tracker' is a recommended property
- [x]	KISAO tracker: 'tracker' is a recommended property
- [x]	[MS tracker](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/976): 'tracker' is a recommended property
- [x]	SWO tracker: 'tracker' is a recommended property
- [x]	VT tracker: 'tracker' is a recommended property
- [x]	ZFS tracker: 'tracker' is a recommended property

matentzn:
@ybradford for ZFS FYI

matentzn:
Action items:
- [ ] Contact KISAO to ask for tracker
- [ ] Set up ZFS repo (Nico and @ybradford)
- [ ] Decide what to do if an ontology does not want to provide a tracker (I would like trackers to be mandatory)

matentzn:
We have decided in the OBO foundry meeting of Jan 12 2020 that every ontology _must_ provide a tracker and that this check will be part of the metadata. 

@gambardella Would you be able to provide us with a public issue tracker for KISAO (some place where users can ask for new terms etc)?

gambardella:
Sure. https://sourceforge.net/p/kisao/feature-requests/

matentzn:
Thank you!

matentzn:
This has been completed!

PR #1400, opened 2021-01-12
Add Tracker to ZFS

matentzn:
#1393


jamesaoverton (approved):
I guess this is right but the different GitHub organizations used for different fields is weird.

matentzn:
ZFS is managed in UBERON: https://github.com/obophenotype/developmental-stage-ontologies/tree/master/src/zfs

but the actual curation is done by the ZFA team! @cmungall suggested this as the most likely tracker.

PR #1412, opened 2021-01-15
Moved some inactive OBO committee members to Alumni (after discussion during 2020-11-17 OBO Operations meeting)

### LINKED ISSUE #1278: Update Operations Committee members?
I think some of the members listed on  http://obofoundry.org/docs/Membership.html may have moved on. It may be time to revise the list. I made a spreadsheet last August--can update. https://docs.google.com/spreadsheets/d/1VurNA9SahgNUxeU9ZxeK4JCcGICNs3oP21FUmy01968/edit#gid=0

(No rush--just wanted to note this while I was thinking of it.)

Also, what is the formal process for adding new people to the Operations Committee? I would like to nominate @nicolevasilevsky.

nlharris:
fixes #1278


cmungall (changes requested):
make a separate PR for adding Becky - this is just correcting an accidental omission

proposals to move people to alumni should be done as a separate PR. We currently have no SOP/rules for this and bears more discussion

nlharris:
Ok, makes sense to separate the add and the moves.

Re moving these people to alumni, these moves were proposed, discussed and (I thought) approved during the 2020-11-17 OBO Operations meeting (see meeting notes).

cmungall:
oh so it is there in the meeting notes. and I was even there at the meeting, it says. though it seems less than conclusive about who should be moved though. I'd expect emails to be sent first to avoid surprises. (this is a criticism of our process, or lack of, not your PR!)

nlharris:
This PR should now only be moving some members to alumni (adding Becky was done in a separate PR) but not sure I did that right.

matentzn (approved):

matentzn:
I think this is fine and can be merged. We moved mb back to where he was; now its Eric and Philipe, which we explicitly agreed to in the meeting..

matentzn:
@proccaserra @dougli1sqrd We are trying to the OBO foundry committee member lists up to date - you two came up as possible Alumni; Do you consider yourself active members or Alumni at this point?

matentzn (approved):

cmungall (approved):

PR #1420, opened 2021-01-21
Update schema

matentzn:


jamesaoverton:
I looked through the code pretty carefully, and it looks fine.

But I'm worried that the changes to `registry/obo_context.jsonld` will break stuff, especially once ROBOT is updated to use them. On the other hand, it might also fix things that are currently broken...

But I don't know how to properly test it. Do you have a plan @matentzn ?

matentzn:
My main concern is that robot gets the new and updated curie map of the context before it's next release (needed for dashboard) but I don't know exactly what could break.. in your experience, what are the risks?

jamesaoverton:
See also https://github.com/prefixcommons/biocontext/pull/33

jamesaoverton (approved):
Mixed case is correct for these ontologies (ExO, HsapDv, MmusDv, OlatDv, PdumDv) based on their own OWL files and their PURL configs, which @matentzn and I both double-checked. I also double-checked OLS. @matentzn grepped through all the OBO OWL files looking for the uppercase versions of these five IDSPACEs and did not find any.

So I'm fairly confident that this change will fix things and not break anything.

PR #1424, opened 2021-01-29
Require at least one 'products' entry

### LINKED ISSUE #389: Add Travis check for products

jamesaoverton:
Resolves #389

### DIFF
--- a/util/schema/registry_schema.json
+++ b/util/schema/registry_schema.json
@@ -329,6 +329,7 @@
 			"level": "info",
 			"description": "'products' is a recommended field. The value is a list of resources, each with an ID. The resources should be in one of the following formats: OWL, OBO, JSON, OFN, OWX, TTL, or g-zipped OWL.",
 			"type": "array",
+			"minItems": 1,
 			"items": {
 				"type": "object",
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Does the CI test actually checck all metadata entries against this schema constraint?

jamesaoverton:
Yes, although it's a bit complicated.

The simple part is that `products` is marked as required at the bottom of the schema file. I've added a `"minItems": 1` constraint to `products` in this PR. Then `make test` runs the schema against YAML headers of all the registry entries.

The tricky part is that if the ontology is marked as obsolete, orphaned, or inactive, then JSON Schema failures are downgraded so that `make test` still passes. That code is kind of ugly:

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/util/validate-metadata.py#L151

I tested this PR locally by setting `"minItems": 10` and getting a lot of failures. I'm pretty sure that it works.

Issue #20, opened 2015-08-08
PR tracker is still sourceforge


cmungall:
Is this correct?

lschriml:
Is this ticket referring to Public Relations for the OBO Foundry ? 
http://obofoundry.org/registry/publications.html

cmungall:
This pertains to PRO

nataled:
@cmungall I'm trying to do the migration (finally). It worked for only four issues, then quit (via "die") because an issue was assigned to someone not in the collab file. I then fixed that problem and ran again. The script successfully ran, but nonetheless seems to have stopped at about the same point. Maybe the already-existing issues on github blocked it somehow? Anyway, I now have several duplicated issues. I'm afraid to run it again.

nataled:
Switch has been made. Issues ported to GitHub and SourceForge tracker has been disabled.

PR #1455, opened 2021-03-09
Update fp-020-responsiveness.md

nataled:
Wording has been fully approved, so the 'NOTE' has been removed. Also changed the layout type to 'principle'

### DIFF
--- a/principles/fp-020-responsiveness.md
+++ b/principles/fp-020-responsiveness.md
@@ -1,12 +1,8 @@
 ---
-layout: responsiveness
+layout: principle
 id: fp-020-responsiveness
 title: Responsiveness (principle 20)
 ---
-
-NOTE
--------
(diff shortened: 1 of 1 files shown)


nlharris (approved):

nlharris:
so is this new principle #20 now going to appear on the website (http://obofoundry.org/principles/fp-000-summary.html)? and i should also add it to the MORRC checklist (https://docs.google.com/spreadsheets/d/1aH8ivqS1pE5IchJtkWYOfebmtkd-s2Zsdq4b5PK8hgQ/edit#gid=0)

nataled:
@nlharris I'm unfamiliar with the MORRC checklist, but eventually it should be treated as all other principles pages are treated. Before that happens though, the automated validation needs updating. I need to submit a ticket for the next step.

Issue #21, opened 2015-08-08
FMA


cmungall:
Currently hardcoded to be the ancient obo format version I made. Should be made the complete ontology now it is in OWL. obo-basic can be a subproduct

cmungall:
cc @trapperkeeper

Now FMA natively produces OWL (not valid OWL2 but will be soon) do we just point directly to it? Do we create a distinct OBO-compatible version? (e.g. OBO PURLs, standard annotation properties and object properties)?

trapperkeeper:
Of course my vote, speaking for the FMA team, would be to use the FMA directly. It is confusing having multiple differing versions in play.

cmungall:
What should the status of the OBO purls be? Both class purls and ontology purls?

 1. Keep these active, and have them direct to an 'OBO translation' of FMA
 2. Obsolete them
 3. FMA obsoletes its existing purls and joins the OBO system

We are effectively doing 1 right now. With the new FMA being natively OWL we can update the translation to be a minimal one of rewriting IRIs, thus the OBO shadow will be semantically equivalent

2 may well break lots of things

3 is up to @trapperkeeper but I suspect the answer is no

See also https://github.com/OBOFoundry/purl.obolibrary.org/issues/63#issuecomment-161012932

trapperkeeper:
Sorry for the late response. I brought this issue up with the FMA team but, as suspected, we are not prepared to obsolete our existing purls. First, some of our purls are already in use, as data annotations, referenced from other sources like UMLS, etc. Second, there are multiple community ontology efforts out there, and sometimes professional tension between them. The FMA, in an attempt to remain neutral, has adopted a purl that is not aligned with any specific effort (just trying to remain Switzerland here).  

Of course, there are factors that may affect FMA purls. The future of purl.org being one that we are watching. 

Regarding choices 1 or 2, from our perspective if would, of course, be nice if there were only one set of purls. But, for legacy reasons and to maintain exiting references I can certainly understand that some mapping or translation method may be necessary.

cmungall:
Thanks

Just to point out one thing: purl.obolibrary.org is now completely 
independent of purl.org. You should not use purl.org for anything as the 
future is, as you hinted, in doubt. purl.org could vanish tomorrow and 
the obolibrary ontologies would not be affected

trapperkeeper:
[...]

PR #1458, opened 2021-03-10
DO NOT MERGE: one solution to FMA problem #21

cmungall:
This is one solution to #21

It removes the old FMA version, and makes it clear to human users that FMA is available outside OBO

There are many positives - OBO infrastucture does not have to make exceptions about OBO class PURLs

Downside - if FMA is to be browsable in OLS, OntoBee (as surely it should be) then OBO is not fulfilling its role as being a central registry of note, and additional ad-hoc registration required

Also FMA will not be evaluated in dashboard

do not merge without discussion!

### DIFF
--- a/ontology/fma.md
+++ b/ontology/fma.md
@@ -23,13 +23,14 @@ products:
 taxon:
   id: NCBITaxon:9606
   label: Homo sapiens
-title: Foundational Model of Anatomy Ontology (subset)
-build:
-  source_url: http://svn.code.sf.net/p/obo/svn/fma-conversion/trunk/fma2_obo.obo
-  insert_ontology_id: true
-  method: obo2owl
+title: Foundational Model of Anatomy Ontology (old deprecated version)
 tracker: https://sourceforge.net/p/obo/foundational-model-of-anatomy-fma-requests/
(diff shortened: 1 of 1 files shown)


lschriml:
Melissa Clarkson at University of Kentucky is now leading the FMA, with updates in the works.
mclarkson@uky.edu

cthoyt:
Can we close this one?

matentzn:
Not yet, unless there is a compelling reason to close? I think @dosumis, @lschriml  and @cmungall should sort out the FMA relationship for OBO and make a proper informed call..

cthoyt:
Well we already marked FMA as inactive in #1752. I guess discussion about whether it's obsolete is still ongoing

lschriml:
The developers, in emails with Chris, asked that the FMA be kept alive.

matentzn:
I will close this now - we will revisit this when we do.

cmungall:
it's unfortunate that this issue was closed. It would have solved the issue. Now everyone is confused by the ancient version of FMA we continue to promote

nlharris:
Should we add this to the agenda for the next OBO Ops meeting?

PR #1464, opened 2021-03-16
Update fp-016-maintenance.md

nataled:
Reordering to conform to standard.


lschriml (changes requested):
Need to update to:
In cases when an area remains controversial, and no consensus is reached, then it is up to the ontology maintainer to either leave out the controversial term, or pick a viewpoint for practical considerations, and note the presence of controversy in an editor note.

lschriml (approved):
Reviewed and approved.

PR #1475, opened 2021-03-27
Moved RBO repo in GH

DanBerrios:
Moved repo to be owned by new org, Radiobiology Informatics Consortium. Update license to CC-BY3

### DIFF
--- a/ontology/rbo.md
+++ b/ontology/rbo.md
@@ -11,8 +11,8 @@ build:
   path: "."
 contact:
   email: daniel.c.berrios@nasa.gov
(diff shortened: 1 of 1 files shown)


jamesaoverton:
The PURL configuration will also have to be updated: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/rbo.yml

jamesaoverton (approved):

DanBerrios:
@jamesaoverton  I believe GH redirects so the existing purls should still work?

PR #1480, opened 2021-04-08
Fix reversed contact email and contact label for several ontology pages

cthoyt:
This PR addresses the issue where the `contact.label` and `contact.email` field were swapped for five ontologies: `aao`, `adw`, `ehda`, `ehdaa`, and `ev`.

I'm aware these were marked as deprecated/inactive, but it's still an issue with anyone trying to automatically consume content from OBO Foundry. I'm not sure the best way to update the other artifacts based on this. @matentzn are you the boss here?

### DIFF
--- a/ontology/aao.md
+++ b/ontology/aao.md
@@ -3,8 +3,8 @@ layout: ontology_detail
 id: aao
 title: Amphibian gross anatomy
 contact:
-  email: David Blackburn
-  label: david.c.blackburn@gmail.com
+  email: david.c.blackburn@gmail.com
+  label: David Blackburn
 taxon:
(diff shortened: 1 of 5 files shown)


matentzn (approved):

matentzn:
Great, thank you @cthoyt 

The other artefacts will be updated manually - I can do it quickly if you deem it urgent.

cthoyt:
@matentzn no rush, though maybe this is something we could automate with GHA to happen on all pushes to master?

matentzn:
No problem, we tend to want to do a bit of manual review :P but yeah, we will probably automate this at some point with github desktop!

PR #1488, opened 2021-04-20
Suggested amendment of the New Ontology Request template

matentzn:


.github/ISSUE_TEMPLATE/new-ontology-request.md
```diff
@@ -49,3 +49,28 @@ GitHub username:
...
+
+## Metadata
+
+Please fill in the following metadata record which will be used by the OBO Foundry website.
+
+```
+id: yourfourletterid
+title: The Title Of Your Ontology
+contact:
+  email: email@email.fr
```
nlharris, line 61: the .fr might be confusing to some. how about
youremail@domain.com

.github/ISSUE_TEMPLATE/new-ontology-request.md
```diff
@@ -49,3 +49,28 @@ GitHub username:
...
+homepage: https://github.com/EBISPOT/amphx_ontology
+products:
+  - id: yourfourletterid.owl
+  - id: yourfourletterid.obo
+dependencies:
+  - id: ro
+  - id: otheroboontology
+tracker: https://github.com/YOURORG/your_repo/issues
+license:
+  url: http://creativecommons.org/licenses/by/3.0/
```
nlharris, line 74: is there a way to (succinctly) make it clearer here that this is just an example, and not the only license we accept?

nlharris (approved):
looks good, two minor comments about the examples

.github/ISSUE_TEMPLATE/new-ontology-request.md
matentzn: I changed it to tdl :D to be very general.

.github/ISSUE_TEMPLATE/new-ontology-request.md
matentzn: Not succinctly, but I added some comments up top. Can you check again?

.github/ISSUE_TEMPLATE/new-ontology-request.md
```diff
@@ -52,17 +52,17 @@ GitHub username:
 
 ## Metadata
 
-Please fill in the following metadata record which will be used by the OBO Foundry website.
+Please fill in the following metadata record which will be used by the OBO Foundry website. Note that all the values are all examples in nature, for example `yourfourletterid` could be something like `aism`, `cohm`, `hpo` (it does not have to be four letters). `your_domain_like_for_example_anatomy` could be simply `anatomy`, and the license should be whatever your actual license is. An example can be found [here](https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/ontology/amphx.md), but you really only need to fill in the metadata mentioned here.
```
nlharris, line 55: Note that all the values are all examples in nature ->
Note that the values shown are just examples

> (it does not have to be four letters)
then why not make it something like "yourshortID"? i think we want to discourage the use of 2 or 3-letter IDs, though, right? so probably best not to show HPO as an example? (it is grandfathered)

nlharris (approved):
thanks, this is helpful.
(i know it seems obvious that you should replace the example text with your own specifics, but if there's a way to be confused, some user will be confused)

i'm not familiar with "tdl" but it's fine. at least it doesn't sound like you have to have an email account in france. :-)

.github/ISSUE_TEMPLATE/new-ontology-request.md
matentzn: Thanks! Fixed these I think!

PR #1490, opened 2021-04-20
Updated OBO metadata

matentzn:


matentzn:
Update: I ran the site locally with `jekyll serve` and could find no obvious problems!

jamesaoverton (approved):
I also check it locally and it looks good to me.

Issue #1474, opened 2021-03-26
Ontology for the Anatomy of the Insect SkeletoMuscular system (AISM)


JCGiron:
Use this form to register a new ontology with the OBO Foundry. Please read the instructions provided here:
http://obofoundry.org/docs/NewOntologyRegistrationInstructions.html

## Ontology title
Ontology for the Anatomy of the Insect SkeletoMuscular system

## Requested ID space
AISM

## Ontology location
https://github.com/insect-morphology/aism

## Contact person
Name: Jennifer Girón
Email address: entiminae@gmail.com
GitHub username: JCGiron

## Issue tracker
https://github.com/insect-morphology/aism/issues

## Ontology license

- [X] CC0 (public domain)
- [ ] CC-BY (version 3 or later)
- [ ] Other: *please specify*

## Available ontology formats
owl, obo, jason

## What domain is the ontology intended to cover?
Insect anatomy

## Related OBO Foundry ontologies
CL, PATO, CARO, BSPO, RO, UBERON

## Intended use/related projects
Multispecies insect anatomy ontologies are used as bases for robust knowledge bases for specific anatomical terms (Yoder et al. 2010, HAO Portal), improve the accessibility of morphology descriptions ([Balhoff et al. 2014](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0094056)), and are useful for improving the modeling of discrete morphological characters in phylogenetic context (Tarasov [2019a](https://academic.oup.com/sysbio/article/68/5/698/5298740), [2019b](https://academic.oup.com/isd/article/3/6/1/5584145)).

There are two major reasons why insects are poorly represented amongst multispecies anatomy ontologies: lack of a base insect anatomy ontology and the lack of infrastructure to build such resource.

The ontology for the Anatomy of the Insect SkeletoMuscular system contains terms used to describe the cuticle - as a single anatomical structure - and the skeletal muscle system in insect biodiversity research. The cuticle is the product of the single layer epithelium and mirrors changes in the morphology of this single sheet (like a blanket). This ontology serves as a root ontology for other multispecies insect anatomy ontologies.
This ontology repository was created using the [Ontology Development Kit](https://github.com/INCATools/ontology-development-kit).

zhengj2007:
Discussed on 2021-04-20 call. 

@matentzn will look at it and reply soon.

matentzn:
[...]

PR #1498, opened 2021-05-05
Create aism

JCGiron:
Added metadata for the AISM following issue https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1474


matentzn (approved):

ontology/aism.md
```diff
@@ -0,0 +1,33 @@
+---
+layout: ontology_detail
+id: aism
+title: Ontology for the Anatomy of the Insect SkeletoMuscular system (AISM)
```
matentzn, line 4: @JCGiron I am wondering - is that way to spell SkeletoMuscular really the right one? Seems bad syntax?

JCGiron: The uppercase letters are highlighting where the AISM acronym comes from, but it can be changed to "Ontology for the anatomy of the insect skeletomuscular system" if that is more appropriate syntax.

matentzn: In this case, I think its fine.

PR #1502, opened 2021-05-13
Rename 'Review' and make it less prominent

jamesaoverton:
I just answered a question from a potential OBI user who was worried that OBI has not been maintained since 2013. This baffled me, but there it is on the OBI details page, on the right, before telling the user how to contact OBI: "Review Date 2013" and "Review Document PDF" with link.

<http://obofoundry.org/ontology/obi.html>

This PR moves it to the bottom of that list and renames it "OBO Review".


nlharris (commented):
This seems like a good idea! Does it need to be discussed on an OBO Operations call?

matentzn (approved):
We dont need to discuss it at the committee, but lets leave it open for a few days. I think this is non-controversial.

PR #1537, opened 2021-06-08
Unobscure issue trackers

cthoyt:
Some issue trackers have been obscured by PURLs. Since there isn't any direct way to identify if an ontology is hosted on GitHub, this is a bit of a problem for automatically interacting with the trackers. This PR unobscures some of these PURLs and replaces them with full GitHub links.

This affects OBI, FBbt, and FBdv. It also should affect [OHD](http://www.obofoundry.org/ontology/ohd.html), but I was unable to follow [its tracker's PURL](https://purl.obolibrary.org/obo/ohd/issues) either because the PURL service is down (I experience this quite often) or it points to a dead link.

The only inactive ontology that had this issue was the Influenza Ontology, but it links to [this SourceForge](https://sourceforge.net/p/influenzo/support-requests/) so I can safely assume it's abandoned.

### DIFF
--- a/ontology/fbbt.md
+++ b/ontology/fbbt.md
@@ -25,7 +25,7 @@ build:
   checkout: git clone https://github.com/FlyBase/drosophila-anatomy-developmental-ontology.git
   system: git
   path: "."
-tracker: http://purl.obolibrary.org/obo/fbbt/tracker
+tracker: https://github.com/FlyBase/drosophila-anatomy-developmental-ontology/issues
 browsers:
   - label: FB
     title: FlyBase Browser
--- a/ontology/fbdv.md
+++ b/ontology/fbdv.md
@@ -25,7 +25,7 @@ build:
   checkout: git clone https://github.com/FlyBase/drosophila-developmental-ontology.git
   system: git
   path: "."
-tracker: http://purl.obolibrary.org/obo/fbdv/tracker
+tracker: https://github.com/FlyBase/drosophila-developmental-ontology/issues
 browsers:
   - label: FB
     title: FlyBase Browser
--- a/ontology/obi.md
+++ b/ontology/obi.md
@@ -12,7 +12,7 @@ description: An integrated ontology for the description of life-science and clin
 domain: experiments
 homepage: http://obi-ontology.org
 mailing_list: http://groups.google.com/group/obi-users
-tracker: http://purl.obolibrary.org/obo/obi/tracker
+tracker: https://github.com/obi-ontology/obi/issues
 contact:
   label: Bjoern Peters
   email: bpeters@lji.org


matentzn:
Cool, never thought of this use but makes sense. @jamesaoverton also Fine?

jamesaoverton:
I've been around long enough to have migrated projects from SourceForge to Google Code and then to GitHub. We should (and are) keeping our options open, so we can move away from GitHub to whatever comes next.

PURLs for issue trackers seem like a good idea to me, and I wouldn't change them in the registry without asking each project what they think.

All of these PURLs worked for me. I have not noticed any PURL outages at all, and would like to hear about them it someone does -- that's serious.

matentzn:
This is not about the PURLs themselves - this is a "trick" to determine where an ontology is located just from the metadata, and using that information for automated analysis of the github repos in question. We can use this to automatically analyse trackers, responsiveness, number of issues open/closed, even to open issues OBO wide if needed (ok sparingly, don't worry). 

I think its a good idea in this case. We do need better analytics of repos associated with ontologies. Migrating away from Github is something we will have to do all at once - one or two ontologies using their purl trackers wont cut it. Plus, we can still _use_ the purl - we just want to have the fact its a github tracker somewhere in the metadata to do the above things? 

If you are against it changes, then @cthoyt you can still achieve the same goals with a redirect analysis - which is not that much of an issue either. Whatever has the more advantages.

jamesaoverton:
Changing these registry entries from PURLs to GitHub will change how they are displayed on the respective ontology pages, e.g. <http://obofoundry.org/ontology/obi.html>. It could change how people bookmark those links. But it's up to the respective projects to choose these URLs.

For this case, I would suggest resolving the PURLs, which is a bit annoying but should be robust.

If you just want to check for GitHub repo, we have a [repository](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/util/schema/registry_schema.json#L368) field in the schema. It's not required (like 'tracker' is) so I guess it won't be filled out as often. I guess I would prefer to update the registry to use 'registry', which could be done semi-automatically. I suppose projects might use a GitHub repo and some other tracker.

cmungall:
FWIW, pure anecdata: users always bookmark what is resolved to.

But I agree - let's just populate the repository field. I think it should
be easy to use in one PR.

cmungall:
@cthoyt how do you feel about a new PR

 - extend json schema to add a new field version_control_url
 - pre-populate this for most projects by hacking issues field
 - manually pad for the ones in this PR

it's a lot to ask for someone not associated with the project but it may be fastest way to do what you need, I think it's the right approach overall

cthoyt:
I think automatically populating the repository field with whatever we can automatically parse out of the issue tracker field, then manually filling in the rest, should be enough. Wouldn't a `version_control_url` be redundant?

I'd be happy to take care of this

matentzn:
I dont mind the redunancy either!

alanruttenberg:
I'm with James on this.
Alan

matentzn:
Alright, lets keep the tracker field as is and populate the repository field instead, sounds like we are all agreed this is ok.

After this is done, we probably have to decide how to maintain this going forward though - Since its not a required field, and wont be, its likely going to be unmaintained for new ontologies..

matentzn:
As we decided against changing the purls, we will leave it to @cthoyt to make a new PR for the repository field. I will close in the meantime.

PR #1544, opened 2021-06-16
register obi-base

cmungall:
useful for dealing with issues like https://github.com/obi-ontology/obi/issues/1375


jamesaoverton:
Sorry, I'm dragging my feet on this. I'm not entirely happy with the OBI base file, which is why we haven't advertised it widely.

nlharris:
Should we discuss this during an OBO Operations call?

matentzn:
No no, not necessary. We have it on our radar and it will solve itself eventually.

jamesaoverton (approved):
The file still needs work, but might as well advertise it.

PR #1545, opened 2021-06-28
Update description, contact info, and tracker for KiSAO

jonrkarr:


cthoyt:
@matentzn not related to this PR in specific, but maybe we could set up a GitHub action that gets triggered to rebuild all downstream files anytime one of the markdown files in this directory are changed in a push to master

matentzn:
We are very conservative - we want to really look at the content of each such change. However, once merged into master, yes, we should consider autobuilding

matentzn (approved):

PR #1552, opened 2021-07-13
Mark gaz as inactive until issues resolved

matentzn:
Discussed at OBO operations call today.

There are some fundamental problems with GAZ:
- geographic location is an instance 
- GAZ has not been updated
- ENVO should be the classes from GAZ, GAZ should be for instances
- GAZ is inconsistent.


cmungall (approved):

cmungall:
I think it's undecided whether the source of classes should be GEO or ENVO cc @mbrochhausen @pbuttigieg -- these ontologies are not orthogonal -- but the overall point remains that GAZ shouldn't be the source of classes here

cmungall:
Note there are many tickets for discussing plans to rescue gaz here: https://github.com/EnvironmentOntology/gaz/issues -- I welcome discussion over here

mellybelly:
I don't understand why we need an ontology for geonames when other standards exist. I view it more like the issues with making gene names available for axioms.

addiehl:
Revision in progress per Lynn.  Making a file without instances.  Will be just a country list.

matentzn:
In conversation with @lschriml on slack, we will mark GAZ as inactive until she finishes her revisions, at which point we will change it back to active. If anyone sees this comment uncommented after 1 week, you can merge.

nlharris:
This PR has one approving review; does it need more reviews before it can be merged?

lschriml:
Reopening this pull request, to 're-activate' the gaz. See ticket:
https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2108

Lynn

PR #1557, opened 2021-07-27
Update gno usage with examples

Hendricks27:
To whom it may concern,

I am part of the development team for GNOme ontology.
Here are a few examples of other groups who use our ontology.
Hope it follows the guidelines.
Thanks.

Best,
Wenjin


matentzn (approved):

matentzn:
I checked all the links and they seem ok! Thank you.

Issue #1446, opened 2021-02-23
Migrate remaining obo format ontologies from dependency on central OBO build


cmungall:
Original ticket #290  

Documentation: http://obofoundry.org/faq/what-is-the-build-field.html

This central job is no longer running. This means that the OWL for some ontologies is lagging behind what the provider intends with their provided obo format version.

We could restart the build, but we think it may be better to get groups set up with providing the OWL themselves. Now with github actions and ODK it should be easier to do this. There is still the concern that many of these groups (e.g model organism communities) may have no technical resources, but we would potentially do this just with PRs in their ontologies

 - [ ] update http://obofoundry.org/faq/what-is-the-build-field.html to say this no longer runs
 - [ ] find list of ontologies that rely on this
 - [ ] work with each such that they are providing owl themselves
 - [ ] clarify principles such that it is clear each group is responsible for providing their own OWL

matentzn:
Quick analysis if this helps:

Out of 179 active ontologies
- 119 raw.github to master/main
- 13 raw.github branch/tag (not master) - see below
- 22 Old Berkley Server - see below
- 3 http://purl.oclc.org - see below
- 4 Github releases (latest)
- 8 ideosyncratic github ones (all for @matentzn to fix)
- 10 other

## Purls to raw.github branch/tag:

https://raw.githubusercontent.com/BFO-ontology/BFO/v2019-08-26/bfo_classes_only.owl  -- bfo
https://raw.githubusercontent.com/biobanking/ontology/20210202/obib.owl  -- obib
https://raw.githubusercontent.com/infectious-disease-ontology/infectious-disease-ontology/2017-11-03/ido.owl  -- ido
https://raw.githubusercontent.com/information-artifact-ontology/IAO/v2020-12-09/iao.owl  -- iao
https://raw.githubusercontent.com/information-artifact-ontology/ontology-metadata/v2020-06-08/omo.owl  -- omo
https://raw.githubusercontent.com/ISA-tools/stato/dev/releases/latest_release/stato.owl  -- stato
https://raw.githubusercontent.com/monarch-initiative/GENO-ontology/develop/geno.owl  -- geno
https://raw.githubusercontent.com/obi-ontology/obi/v2020-12-16/obi.owl  -- obi
https://raw.githubusercontent.com/OHMI-ontology/OHMI/v2019-09-17/ohmi.owl  -- ohmi
[...]

PR #1562, opened 2021-08-03
Update fp-002-format.md

nataled:
Revised to conform to new standard format.


principles/fp-002-format.md
```diff
@@ -14,15 +14,17 @@ The ontology is made available in a common formal language in an accepted concre
 
 A common format allows the maximum number of people to access and reuse an ontology.
 
-## Implementation
+## Recommendations and Requirements
+
+All ontologies MUST have at least one OWL product whose name corresponds to the registered id. Thus the ontology whose IRI is http://purl.obolibrary.org/obo/ro.owl (known to the  OBO Foundry as 'RO'), must have at least the product ro.owl. Developers are free to use whatever combination of technologies and formats is appropriate for development. However, the official OWL PURL for the ontology must resolve to a syntactically valid OWL file using the [RDF-XML](https://www.w3.org/TR/rdf-syntax-grammar/) syntax.
 
-### Recommendations and Requirements
+Ontologies can OPTIONALLY produce an OBO-Format file. This is conventionally the same IRI as the owl, but with .owl changed to .obo. Note that an obo product is not listed by default. If you produce an OBO format product, you should register it under the 'products' field in the appropriate metadata file found in this [folder](https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/ontology).
```
cthoyt, line 21: what about OBO Graph JSON?

principles/fp-002-format.md
```diff
@@ -14,15 +14,17 @@ The ontology is made available in a common formal language in an accepted concre
...
+It does not matter to us if you maintain the source for your ontology in obo or owl or some hybrid. You have the option of either publishing the alternate format yourself (using a tool like ROBOT) or you can have the OBO central build pipeline do this for you. For more information, see the FAQ entry What is the Build field?.
 
-Note: some groups publish an .obo version, and the OBO Foundry pipeline takes care of making the valid .owl file. See the FAQ for details. You may also submit the ontology for review as OBO, see 'criteria for review' below.
+## Implementation
 
-Note also that previously we recommended that ontologies may be available in Manchester syntax or OWL-XML, but we have revised this in order to make the official OWL release consumable by a wider variety of tools.
+ROBOT offers functionality to convert a variety of formats, including OBO, to RDF/XML. Protégé allows you to save ontologies in RDF/XML, as well. The [Ontology 101 Tutorial](https://ontology101tutorial.readthedocs.io/en/latest/StartingProtege.html) has directions on starting and saving in Protégé.
```
cthoyt, line 27: Recommendations on validating OBO file format if curation is done primarily there?

principles/fp-002-format.md
nataled: (Just an FYI: The text above was pulled from existing text found within the criteria for review link for this principle.)

I don't know enough about that format. I would say that any format convertible by ROBOT or Protege is fair game. Is that the case for OBO Graph JSON? If so, I'll generalize the statement and add that as another example.

principles/fp-002-format.md
nataled: There is an ongoing discussion about whether or not the Foundry will take up the task of converting file formats. If the answer is 'yes', then we'd indeed need to say something about validating OBO format (and OBO Graph JSON and Manchester and others). Perhaps it would be enough to recommend that the developers at least attempt to use their file as input for ROBOT or Protege, both of which would complain if the format is incorrect.

principles/fp-002-format.md
cthoyt: Yes, ROBOT also will generate OBO Graph JSON. I think you're right that it's good to mention that any other exports will probably help people

principles/fp-002-format.md
cthoyt: For parsing OBO, there is also @althonos's [pronto](https://github.com/althonos/pronto) package. I'd recommend it with the caveat that it's quite strict and curators don't tend to like being told they've made mistakes by a computer program

althonos: For the purpose of validation, i also wrote [`fastobo-validator`](https://github.com/fastobo/fastobo-validator/) that is available in the ODK, it's a bit better at reporting errors (and can detect some additional issues `pronto` can't find, like frames with duplicate definitions or broken ISBN references).

matentzn: All the curators I know love to be told they made a mistake by a computer program.. In fact, many of them love to code nowadays!

matentzn: @nataled  I don't think we should take on any such responsibilities (transformations into OBO etc) - we need to decentralise and enable (fishing rod) through tools and training rather than providing services that maintain the technical illiteracy. I would just add a link to fast-obo validator here as a way for people to validate their OBO formats (90% of all OBO files use ROBOTs ominous `--check false` which means they are broken). As EWG you need to decide wether we would want to _require_ non-broken OBO as part of dashboard checking - I think that's a huge churn, but possible useful. No idea whats best.

cthoyt: @matentzn you're lucky to know such good curators! I am thinking back to the dark days of my PhD where I became a heretic when introducing the PyBEL biological expression language validator.

I agree with @matentzn - centralizing working with the data itself wouldn't be sustainable. Reporting on what could be converted might be a useful metric, but I wouldn't say it should be required at the moment

nataled: @matentzn @cthoyt the decision to provide or not provide conversion services is not one the EWG makes. ALL principles decisions are made by Ops; EWG just puts them into words. The wording reflected above is merely a rewrite of what's already given in the principle and in the FAQ (IIRC). As I mentioned in my previous comment, there is an unfinished discussion about what services the Foundry are willing to provide in this regard. It's been proposed that we do NOT provide the conversion service, and so far as I recall, that seems to be the way it will be, but, again, until it is finalized the EWG cannot take the sentence out.

matentzn (approved):
I am happy with the content, @nataled if you want to add anything minor I am still fine, but I am happy for this to go.

jamesaoverton:
Sorry to be slow to look at this.

Regarding "the OBO central build pipeline will do this for you", my understanding is that this service is no longer provided, and so this part of the Principle need to be changed. More details in #1446.

principles/fp-002-format.md
jamesaoverton: My suggestion would be to replace the first sentence of this paragraph with: "Ontologies can OPTIONALLY produce products in other formats. For historical reasons, OBO-Format files are often produced." Then the rest of the paragraph (as written) is an example of a .obo product.

principles/fp-002-format.md
jamesaoverton: I don't see the need to mention OBO-format validators on the Principle page. In general, there's so much to know about ontology development best practises, and that sort of documentation should live somewhere else.

PR #1565, opened 2021-08-10
remove link to uberon tracker for tao

cmungall:
I'm not sure including this link makes much sense, cc @wdahdul @balhoff?


cmungall:
this was also confusing https://cthoyt.com/obo-community-health/ --- but then maybe we should just have a dedicated field for the github repo

balhoff (approved):

PR #1568, opened 2021-08-12
Regenerate files

jamesaoverton:


cthoyt:
Thanks @jamesaoverton, this was long overdue! Are there instructions anywhere on how to rebuild all of this? I would like to make a GitHub action that could automate this either on request, when one of the files that might require a rebuild is needed, or on a chronological basis

matentzn:
@cthoyt https://github.com/OBOFoundry/OBOFoundry.github.io#generation-of-downstream-artefacts

We are quite conservative, but I guess if there is a pull request in the middle a gh action approach would work.. It was a bit of a pain though to set up the exact version of jekyll and other needed to do it, but you can find infos around the above if you are keen to automate!

matentzn (approved):

PR #1569, opened 2021-08-14
Proposed removal of confusing text

kltm:
`In general, copyright legislation says that facts that are not copyrightable are excluded from copyright protection. Therefore, some ontology content may not be copyrightable.`

I find this statement a little confusing. I'm not sure who this remark is intended for and what action they would take with this information.

 If there was copyright legislation that said that some things couldn't be copyrighted, those things by definition wouldn't be copyrightable anyways, so I'm not sure what additional clarity this statement provides. (As well, definitions of copyrights are different in different jurisdictions (e.g. France)).
I might hazard a guess that the "legislation" is a reference to Feist v. Rural, or the "can you copyright a phonebook" case? If so, it's actually a little more subtle than that (e.g. '"at least modicum” of creativity’ (https://guides.lib.umich.edu/copyrightbasics/copyrightability).

Tagging @cmungall


matentzn (approved):

nlharris:
Can we merge this, or are more reviews needed?

nlharris (approved):

Issue #1563, opened 2021-08-09
Many OBO ontology repositories don't specify a LICENSE file


cthoyt:
The OWL and OBO formats have ontology-level metadata for linking to various accepted Creative Commons (or equivalent) permissive licenses.

GitHub also has a first-class notion of a repository-level `LICENSE` file that allows programmatic access to the licensing information, as well as adding the information to the repository overview page. For many purposes, such as reviewing the overall health of the OBO community, having direct programmatic access to this data is favorable to parsing all ontologies. The OBO Foundry also maintains the license information for some ontologies, but it's hard to tell if these are up to date and it is not always complete.

It would be nice for repositoies for OBO Foundry ontologies to maintain the LICENSE file in parallel to the structured metadata in the OWL format. Several already do. Similarly to #1538, I am thinking about automating adding issues to all repositories that do not currently have a LICENSE file. Here are two resources that I will include:

1. A tutorial on how to add a LICENSE file through the github interface - https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/adding-a-license-to-a-repository
2. Information about how to choose a permissive license, if one hasn't already been assigned (though to be in OBO Foundry,   I think it's already required to have a permissive license based on [fp-001](http://www.obofoundry.org/principles/fp-001-open.html)) - https://choosealicense.com/non-software/

matentzn:
I think this is a great thing to push for. The OBO foundry could really do with a tutorial on licenses - For example, what does the LICENSE file really refer to? The OWL ontology file? The Makefile ODK pipeline? The documentation?

nlharris:
This is a good suggestion, @cthoyt. Where on the obofoundry website should the instructions about license files go?

Also ccing @kltm (Seth) who is an expert on licenses.

kltm:
I think the only thing I'd add here for the moment is that I'd be wary of CC-BY-SA-4.0; the Recommendations and Requirements at http://www.obofoundry.org/principles/fp-001-open.html I think do a good job explaining what is desired.

From experience, how licensing information is best communicated can depend on the audience. Options include clearly and unambiguously on website, in a repo, or in the ontology itself. It would be good to get recommendations (and howtos) into the official documentation. It would be pretty easy in a lot of cases to also have the automatic checks to make sure that these existed and were in sync.

matentzn:
> I think the only thing I'd add here for the moment is that I'd be wary of CC-BY-SA-4.0;

[...]

PR #1586, opened 2021-09-08
Clarifying license properties

matentzn:
See #1563.


principles/fp-001-open.md
```diff
@@ -18,7 +18,7 @@ OBO Foundry ontologies are resources for the entire biological and biomedical co
 
 ### For ontology providers
 
-OBO Foundry Ontologies MUST EITHER be released under a [Creative Commons](https://creativecommons.org/) [CC-BY license version 3.0](https://creativecommons.org/licenses/by/3.0/) or later, OR released into the public domain under [CC0](https://creativecommons.org/publicdomain/zero/1.0/). It should be clearly stated in the ontology file.
+OBO Foundry Ontologies MUST EITHER be released under a [Creative Common Attribution 3.0 International (CC BY 3.0) license or later](https://creativecommons.org/licenses/by/3.0/) or later, OR released into the public domain under [a Creative Commons CC0 public domain dedication](https://creativecommons.org/publicdomain/zero/1.0/). The license MUST be [clearly stated](https://github.com/INCATools/ontology-development-kit/blob/master/docs/License.md) using the http://purl.org/dc/terms/license property in the ontology file.
```
cthoyt, line 21: Still possibility for confusion - what comes after http://purl.org/dc/terms/license? Maybe someone will put free text instead of the proper software URL. Could you give an example, plus a list to the controlled vocabulary that could be used?

matentzn: Thank you @cthoyt, done.. Better now?

cthoyt (approved):

nataled (changes requested):
The new text incorrectly indicates the name of the license, which is Creative Commons Attribution 3.0 Unported. The International version did not start until 4.0. The following is the text after correcting that and other issues as well:

OBO Foundry Ontologies MUST EITHER be released under a [Creative Commons Attribution 3.0 Unported (CC BY 3.0)](https://creativecommons.org/licenses/by/3.0/) license or later, OR released into the public domain under [Creative Commons CC0 1.0 Public Domain Dedication (CC0 1.0)](https://creativecommons.org/publicdomain/zero/1.0/). The license MUST be clearly stated using the http://purl.org/dc/terms/license property followed by the URL representing the license (e.g. https://creativecommons.org/licenses/by/3.0/) in the ontology file ([OWL example](https://github.com/INCATools/ontology-development-kit/blob/master/docs/License.md)).

Please revise accordingly.

cthoyt:
if that's the case, then the text should also include the full title of CC BY 4.0 so users know there's a difference in addition to the other 2 examples

nataled:
CC BY 4.0 isn't mentioned, nor does it need to be since it isn't a requirement. That being said, I think CC BY 4.0 is a better choice of license. The Ops Committee would need to approve this change (it was decided some years ago to keep it at 3.0, but that was when 4.0 was pretty new). For such a discussion, I think it would be useful to count how many still use 3.0. If none, then the change is easy. If there are any that still use it, we could request that they update, and if all do so, the change again is easy.

matentzn:
Is 4.0 more permissive such that a migration from 3.0 to 4.0 is actually possible?

matentzn:
@nataled thanks a lot for your review; I have changed the text as per your suggestion, and also added explicitly the CC-BY 4 license an example in brackets (which is safe I would think). We can discuss recommending it officially, but currently, more than half (77) of the active ontologies use CC-BY 3.0.

nataled:
Yes, but that's not the point. Remember, no changes to the *meaning of* principles occur without Ops discussion. The decision from previous discussion was to use CC BY 3.0 even though CC BY 4.0 was available at the time, so that's what we say in the principle. If we don't have any ontologies using 3.0 than I don't imagine there'd be any issue making the change, but that needs verification.

matentzn:
Can you double check what comment you commented on, I think we got our comments crossed :D Sorry.

nataled (approved):

nataled:
My comment was on the permissiveness question, which I hadn't yet uploaded before you commented again.

matentzn:
Thank you!

kltm (commented):
The changes appear to be correct and an improvement--thank you!

I might note that for consistency (and technical correctness) that line 23's "CC-BY" should probably be "CC BY".

kltm:
NM--merged earlier!

PR #1593, opened 2021-09-17
adding usages

sbello:
Added MGI, RGD and IMPC to usages section

### DIFF
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -58,7 +58,7 @@ usages:
     description: "IMPC annotates abnormal phenotypes of mice carrying null alleles found following the application of a standardised set of physiological tests"
     examples:
       - url: https://www.mousephenotype.org/data/phenotypes/MP:0013522
-        description: "All IMPC alleles that have been annotated to the MP term "decreased memory-marker CD4-positive NK T cell number"."
+        description: "All IMPC alleles that have been annotated to the MP term 'decreased memory-marker CD4-positive NK T cell number'."
 jobs:
   - id: http://build.berkeleybop.org/job/build-mp-edit
     type: DryRunBuild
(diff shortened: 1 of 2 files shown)


matentzn (approved):

ontology/mp.md
```diff
@@ -43,6 +43,22 @@ browsers:
...
+        description: "Term browser page for embryonic lethality showing information about the term including definition, placement in the MP hierarchy, and link to mouse models annotated to this term or any of its decendants"
+  - user: https://rgd.mcw.edu/rgdweb/ontology/view.html?acc_id=MP:0000001
+    description: RGD annotates phenotypes associated with rat genes and alleles using the MP
+    examples:
+      - url: https://rgd.mcw.edu/rgdweb/ontology/view.html?acc_id=MP:0008762
+        description: "Term browser page for embryonic lethality showing information about the term including definition, placement in the MP hierarchy, and link to annotations to this term or any of its decendants"
+  - user: https://www.mousephenotype.org/
+    description: "IMPC annotates abnormal phenotypes of mice carrying null alleles found following the application of a standardised set of physiological tests"
+    examples:
+      - url: https://www.mousephenotype.org/data/genes/MGI:97490#phenotypesTab
```
jamesaoverton, line 60: I don't see a use of MP on this page. @matentzn suggests something like https://www.mousephenotype.org/data/phenotypes/MP:0013522 instead.

sbello: The significant phenotype table on the page uses MP terms for all the phenotype terms. They just don't show the MP IDs.
IMPC_pheno

sbello: I can switch to the other page if you think that is easier to understand or better as an example.

matentzn: Just because we are still clarifying the guidelines, I think it would be great if you would use URLs with MP ids in it.. :) Thank yOU!

PR #1612, opened 2021-09-29
Update doid.md

lschriml:
added AGR

### DIFF
--- a/ontology/doid.md
+++ b/ontology/doid.md
@@ -43,11 +43,11 @@ license:
   url: https://creativecommons.org/publicdomain/zero/1.0/
   label: CC0 1.0 Universal
 usages:
-  - user: http://www.informatics.jax.org/disease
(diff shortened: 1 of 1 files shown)


ontology/doid.md
```diff
@@ -43,11 +43,11 @@ license:
   url: https://creativecommons.org/publicdomain/zero/1.0/
   label: CC0 1.0 Universal
 usages:
-  - user: http://www.informatics.jax.org/disease
-    description: MGI disease model annotations use DO
+  - user: https://www.alliancegenome.org
+    description: Alliance of Genome Resources
     examples:
-      - url: http://www.informatics.jax.org/disease/DOID:4123
```
matentzn, line 49: Is there a reason why you delete the usage from JAX? Wouldnt it be better to have both AGR and JAX show up under usages? :)

PR #1620, opened 2021-10-07
Add missing preferred prefix annotation for VariO

cthoyt:
Following the style of FBbt and related, this PR adds the lexical variant of VariO as the "preferred prefix" to the ontology metadata.


cthoyt:
I think I wrote a harness to mass update the stuff in this repo before... I'll try and dig it up

PR #1621, opened 2021-10-07
Add inferred preferred prefixes to active ontologies

cthoyt:
As a follow-up to PR #1620 where I added an exotic preferred prefix for VariO, this PR automatically infers that the uppercase of the ontology `id` field in the metadata should be the preferred prefix. 

This PR explicitly excludes ontologies that already have a preferred prefix as well as ones that are marked as obsolete, inactive, and orphaned to cut down on the burden of checking them all (for now). I'm curious if there's a semi-automated way to check that these are all in fact right, since it sounded like @matentzn's check for #1620 was by hand.

## Consider Also

The following code is the block that generates the ontology description page, such as http://www.obofoundry.org/ontology/ro.html for the Relation Ontology.

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/a0811cc4362fc4cf1088ee3a47941b432303d683/_layouts/ontology_detail.html#L48-L55

The wording right now makes it sound like the preferred prefix should only be annotated when it's mixed case. If this PR is to be accepted, we should also amend this wording, or perhaps do a bit more logic in this file to note when something is a "simple" preferred prefix vs. a mixed-case preferred prefix.

### DIFF
--- a/ontology/aeo.md
+++ b/ontology/aeo.md
@@ -21,6 +21,7 @@ build:
   method: vcs
 activity_status: active
 repository: https://github.com/obophenotype/human-developmental-anatomy-ontology
+preferredPrefix: AEO
 ---
 
 The AEO is an ontology of anatomical structures that expands CARO, the Common Anatomy Reference Ontology, to about 160 classes using the is_a relationship; it thus provides a detailed type classification for tissues. The ~100 new classes were chosen for their use in categorizing the major vertebrate and invertebrate anatomy ontologies at a granularity adequate for tissues of a single cell type. This site is to be used for posting details of the ontologies and updates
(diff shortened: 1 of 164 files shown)


cthoyt:
@matentzn okay sounds good. I wonder what the possible downstream implications will be? Right now I'm scrambling to update the Bioregistry to work with this change, for example ;)

matentzn:
No, its good this is only done for active ones, otherwise we run the risk of never knowing we have added something wrong. 

Yes, we can make this required, I don't see why not (other than that I don't have time to think of the implications :P). Once it is required in the schema, I will add this to the NOR issue template and the ODK, which should make it easy to fill in. I don't really want to have to explain to people that want a different preferredPrefix than there ID scheme (say I want `VARIO` is my preferred prefix, but I use `VariO` in my IDs). Its a bit hard to communicate that it is called `preferredPrefix` but means: whatever you used to describe your OBO purls. Going forward, we will explicitly prevent odd capitalisation, and require all-uppercase I think

matentzn:
No haven't done that, but I can't do anything at the moment as my laptop is in the hospital.. this will have to wait for at least 2 weeks.

cthoyt:
Thanks @jamesaoverton :)

Issue #1633, opened 2021-11-02
Cite new OBO Foundry paper 🎉


nlharris:
- [x] Add to http://obofoundry.org/registry/publications.html
- [x] Replace menu item "Cite the OBO Foundry"

PR #1634, opened 2021-11-02
added new OBO Foundry paper

nlharris:
while i was at it, added a GO paper and also added a disclaimer to the "Ontology project publications" header.
(possibly we should get rid of that list of publications, since it is not maintained, but not trying to do that in this PR)


nlharris:
addresses part of #1633

matentzn (approved):

PR #1643, opened 2021-11-05
Update contact information for ARO to conform with FP11

cthoyt:
Currently, ARO violates FP11 (http://www.obofoundry.org/principles/fp-011-locus-of-authority.html) (based on my interpretation of it) because it uses a group email instead of denoting the single responsible person.

I've found the information for @raphenya via one of [his publications](https://www.mdpi.com/1999-4915/12/8/895) that lists his email address and added it as the contact information here. However, it appears Amogelang is a graduate student and might not be the ideal choice for a long term responsible contact person for the ontology. An alternative could be the corresponding author on ARO's [most recent publication](https://pubmed.ncbi.nlm.nih.gov/31665441/), Andrew G McArthur (@agmcarthur, mcarthua@mcmaster.ca).

Hopefully both of them can get in touch here and let us know how to proceed.

PR #1644, opened 2021-11-05
Update Bgee contact information to conform with FP11

cthoyt:
All of the Bgee resources are in violation of FP11 by listing a group email address instead of a single responsible person (similarly to #1643).

This PR switches that to Frédéric Bastian (@fbastian), the first and corresponding author on the [most recent Bgee publication](https://academic.oup.com/nar/article/49/D1/D831/5920517).

### DIFF
--- a/ontology/cio.md
+++ b/ontology/cio.md
@@ -2,8 +2,8 @@
 layout: ontology_detail
 id: cio
 contact:
-  email: bgee@sib.swiss
-  label: Frederic Bastian
+  label: Frédéric Bastian
+  email: frederic.bastian@unil.ch
   github: fbastian
 license:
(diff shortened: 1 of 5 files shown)


matentzn (approved):

ontology/cio.md
```diff
@@ -2,8 +2,8 @@
 layout: ontology_detail
 id: cio
 contact:
-  email: bgee@sib.swiss
-  label: Frederic Bastian
+  label: Frédéric Bastian
+  email: frederic.bastian@unil.ch
```
matentzn, line 6: Seems to be missing the github handle..

cthoyt: Interestingly, this one already had the right github handle annotated so it doesn't appear in this PR's diff

PR #1645, opened 2021-11-05
Create disdriv.md

lschriml:
Creating metadata file for DISDRIV ontology

### DIFF
--- a/ontology/disdriv.md
+++ b/ontology/disdriv.md
@@ -10,12 +10,12 @@ twitter: diseaseontology
 domain: disease
 homepage: http://www.disease-ontology.org
 products:
- - id: disdriv.owl
+  - id: disdriv.owl
 browsers:
   - label: DO
     title: DO Browser
     url: http://www.disease-ontology.org/
-taxon: 
+taxon:
   id: NCBITaxon:9606
   label: Homo sapiens
 title: Disease Drivers Ontology
@@ -27,7 +27,7 @@ license:
   label: CC0 1.0
 usages:
   - user: https://www.disease-ontology.org
-    description: Human Disease Ontology 
+    description: Human Disease Ontology
     examples:
       - url: https://www.disease-ontology.org/?id=DOID:0050665
         description: fetal alcohol syndrome, has exposure stressor some alcohol
--- a/ontology/disdriv.md
+++ b/ontology/disdriv.md
@@ -1,0 +1,38 @@
+---
+layout: ontology_detail
+id: disdriv
+contact:
+  email: lynn.schriml@gmail.com
+  label: Lynn Schriml
+  github: lschriml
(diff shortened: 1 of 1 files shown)


matentzn:
Hey @lschriml , so this means you decided against the ecto route? Can I ask why?

lschriml:
Hello @matentzn 
     - I will be using ecto terms, as you devised, where the disease drivers are within the scope of ecto.
     - There are disease drivers that are outside the scope of ecto, they are not environmental exposures, e.g. diet, food deprivation, stress. Thus, I will create DISDRIV, and import ecto for all of the terms that I can.
     
Cheers,
Lynn

matentzn:
@laurenchan @diatomsRcool can you comment on that? Is it true that these will be out of scope?

matentzn:
I think "environmental" really means "anything thats not genetic", right?

mellybelly:
dietary and environmental exposures are a major focus of ECTO and are used in disease axioms.

lschriml:
Defining scope is very helpful, thank you !! 

How about things like stress and emotional anxiety that are internal and not external (environmental)

matentzn:
Just so I know wether this conversation is worth it, let's say all of these are in scope for ECTO - would you still want the DISDRIV namespace?

lschriml:
Yes I do want the namespace.

matentzn (approved):

ontology/disdriv.md
```diff
@@ -0,0 +1,38 @@
...
+  id: NCBITaxon:9606
+  label: Homo sapiens
+title: Disease Drivers Ontology
+build:
+  source_url: https://github.com/DiseaseOntology/DiseaseDriversOntology/tree/main/src/ontology/disdriv.owl
+tracker: https://github.com/DiseaseOntology/DiseaseDriversOntology/issues
+license:
+  url: https://creativecommons.org/publicdomain/zero/1.0/
+  label: CC0 1.0
+usages:
```
matentzn, line 28: I would prefer if we remove the usage, until there is some actual usage..

diatomsRcool: I would say that we are using "environmental" very broadly here to include things like diet and socioeconomics. The issue then becomes our reliance on other ontologies to build ECTO classes because there is no socioeconomics ontology.

laurenchan:
Hello. I was put on here by mistake. I’m not a programmer and won’t be any use to your project.

matentzn:
@laurenchan indeed :) Sorry. @laurenechan was the one we meant.

lschriml:
Hello @matentzn -- I thought usage was a required field.
    I am using disdriv (without a purl) as an import for the DO already.
See: https://disease-ontology.org, owl tree

Can this pull request be merged ? 

Cheers,
Lynn

PR #1647, opened 2021-11-08
Added additional usages for GO

cmungall:


matentzn:
Another nice one could be open targets:
https://platform.opentargets.org/disease/GO_0008283/associations

Not sure about the quality though

matentzn (approved):
I checked all URLs and it looks all good.

matentzn:
Another nice one could be open targets:
https://platform.opentargets.org/disease/GO_0008283/associations

Not sure about the quality though

PR #1648, opened 2021-11-09
Add warning boxes on ontology pages for inactive and orphaned ontologies

### LINKED ISSUE #1312: inactive ontologies are not marked as such on individual ontology pages
http://obofoundry.org/ontology/rnao

nothing on page to indicate inactive

the front table is OK though

cthoyt:
Closes #1312

It seems like a bit sneaky that you can navigate to the page for [FIX](http://www.obofoundry.org/ontology/fix), which is orphaned, but the ontology page does not say so. This PR adds information on both inactive and orphaned ontologies.

- TODO: these should link to pages describing exactly what these two terms "inactive" and "orphaned" mean, see https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1126


matentzn:
Needs a quick discussion at the OFOC call.  I very much agree with it.

Thank you!

matentzn:
@nataled I assigned this to you as chair of the next call, and also put it already on the agenda in the meeting minutes.

nataled:
@cthoyt basically we agreed that the inactive and orphaned ontologies should get a similar message as the obsolete cases. Furthermore, we agreed that a documentation page should be created that describes these statuses. @matentzn agreed to handle this. I'm wondering if a full-fledged page is needed though. Perhaps just a link to a question in the FAQ would suffice?

cthoyt:
> 

Great! It's just that it's not really clear what orphaned and inactive mean (and why they are different from obsolete) so I'd be happy for that to be included in an FAQ and just link there too.

jamesaoverton:
The clearest discussion of these distinctions and the reasoning behind them is here https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1126#issuecomment-592541166. We agreed on the call today that better documentation is needed.

matentzn:
Will deal with all that tomorrow.

matentzn:
Action items:

- [ ] Merge this: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1694
- [ ] Make sure in menu
- [ ] Deploy to life page
- [ ] change text here to include the link

matentzn (approved):

matentzn:
Tada:

Works:

image

Link didnt work as expected, will check

PR #1651, opened 2021-11-14
change license

Lemon-Liu:
change the license to GNU General Public License v3.0

### DIFF
--- a/ontology/iceo.md
+++ b/ontology/iceo.md
@@ -11,8 +11,8 @@ contact:
 products:
   - id: iceo.owl
 license:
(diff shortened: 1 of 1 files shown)


cthoyt:
Are you sure you want to do this? This would mean ICEO no longer conforms to the [OBO Foundry's openness principle (FP-001)](http://www.obofoundry.org/principles/fp-001-open.html). Further, using GNU General Public License v3.0 is more appropriate for code/software than creative work.

Lemon-Liu:
Thanks for the reminder.  I will close this pull request.

PR #1652, opened 2021-11-15
Regenerate files

jamesaoverton:

Issue #1404, opened 2021-01-13
Rework GitHub Team membership and permissions


jamesaoverton:
I would like to reorganize our GitHub Teams and permissions with the primary goal of reducing the number of people with Admin permissions for our repositories, and giving everyone else just the permissions that they need (and are currently using).

We talked about this on yesterday's OBO Foundry Operations Committee (OFOC) call, and had unanimous agreement in principle. I'll spell out the details of my proposal here, and we'll try to get final approval on the next OFOC call in two weeks. Unfortunately it's a bit complex, but I'll do my best to be clear.

GitHub has two tiers of permissions, Organization and Repository, and a few permission levels within each:

- Organization (https://github.com/OBOFoundry)
    - Levels
        - Member
        - Owner
    - Documentation: https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/permission-levels-for-an-organization
- Repository (e.g. https://github.com/OBOFoundry/OBOFoundry.github.io)
    - Levels
        - Read
        - Triage
        - Write
        - Maintain
        - Admin
    - Documentation: https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization

The key concern is that anyone with Owner or Admin permissions can accidentally delete our repositories. While it's possible to recover the main repository contents, it turns out that a lot of other important information cannot be recovered. A secondary concern is that our current Teams are not well organized or described:

https://github.com/orgs/OBOFoundry/teams

When we migrated to GitHub about six years ago, the permission levels provided were quite coarse. At that time we made a long list of Owners and added all the members of the OBO Foundry Operations Committee to the "OBO-Admin" GitHub Team, and gave that team Admin permissions on all our repositories. That team now has more than 20 members. Many of those people may not realize that they have Admin permissions.

These lists are not public, but if you are on the list and logged in to GitHub you'll be able to see your name:

- Owners: https://github.com/orgs/OBOFoundry/people?query=role%3Aowner
- OBO-Admin: https://github.com/orgs/OBOFoundry/teams/obo-admin/members

Since that time, GitHub has implemented a range of more fine-grained permissions, and we've learned more about what permissions people actually need in practise.

I propose these three GitHub Teams:

- OBO Admin
    - description: A small group of active Operations Committee members who have multiple years of in-depth experience with OBO and GitHub
    - organization permissions: Owner
    - repository permissions: Admin
[...]

PR #1657, opened 2021-11-16
Add a GitHub Action to automatically rebuild files

### LINKED ISSUE #1224: Use github actions to generate _config.yml
currently we rely on an obo maintainer to manually run the makefile to compile any changes to ontology/, then commit push results

We should just use github actions for this. It's easy to make a custom workflow that will run the make step.

Note: if we wanted to move away from hosting on gh pages, e.g. static site served off of S3, there are actions for this:

https://github.com/marketplace/actions/build-jekyll

cthoyt:
Closes #1224

This PR adds a new GitHub Action configuration, `build.yml` that does the following things:

1. Installs ODK container
2. Installs the Python requirements from this repository's `requirements.txt` 
3. Runs `make`
4. Commits/pushes the result back to the `master` branch

It automatically runs any time a markdown file is changed in the `ontology/` folder on the master branch. This means that there's no longer a need for a second set of discussions, PRs, etc. every time a metadata file is updated. In theory, the build process is automated and does not itself need review. This also has the added benefit that potential contributors can see their updates incorporated much faster.

This action can also be run on demand, in case some of the tooling changes and you don't want to wait for the next change to a markdown file.

You can see the workflow in action at https://github.com/cthoyt/OBOFoundry.github.io/runs/4226590920?check_suite_focus=true and see the resulting commit it made in https://github.com/cthoyt/OBOFoundry.github.io/commit/663461070fb5889be058863069530e05251d3419. Note it's a pretty boring commit due to the fact that no markdown files were changed. Unfortunately, the RDF dump is not deterministic so this is actually a completely spurious diff. Either way, it demonstrates that everything runs properly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,35 @@
+name: Build derived files
+on:
+  # Allows for on-demand build
+  workflow_dispatch:
+  # Automatically rebuilds when somebody edits a markdown file in
+  # the ontology/ directory
+  push:
+    branches:
+      - master
+    paths:
+      - 'ontology/*.md'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.2.30
+    steps:
+      - uses: actions/checkout@master
+        with:
+          persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
+          fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
+      - name: Create local changes
+        run: |
+          python -m pip install -r requirements.txt
+          make
+      - name: Commit files
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git commit -m "🚀 Automatically rebuilt derived files" -a
+      - name: Push changes
+        uses: ad-m/github-push-action@master
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          branch: ${{ github.ref }}


cthoyt:
Question: does https://github.com/OBOFoundry/OBOFoundry.github.io#generation-of-downstream-artefacts need to be updated? I'm not sure if the jekyll step is necessary or not. It's currently not covered by this PR

matentzn:
Awesome, thanks! I think this may be the right move, but so far we have been more conservative as we (used to) frequently spot errors to metadata in particular during the overall review of ontologies.yml and the various context files. I am right this moment on the fence whether we should commit to master. I would prefer a pull request, but only just. if @cmungall or @jamesaoverton or others on the technical team have different opinions I am also ok to risk it. You may want to use odklite instead of odkfull which is a much smaller container.

cthoyt:
@matentzn can you better describe what kind of errors you could catch by looking at the output of the build that wouldn't be possible to find by looking at the diff / using the CI tests before merging the actual PR that changes the underlying metadata files?

The last 2 PRs I found #1652 and #1349 didn't have any comments related to the regeneration of the files. In #1652, there were actually additional files that got edited - I think it would be good to specifically avoid conflating building files with actual edits

matentzn:
Ok, seems like in the associated [issue](#1224), we agreed on risking it.. Lets roll with it for a while and retreat when there is an issue. Thanks!

For the record, i am thinking of things like when our context is generated, it generates for example a curie map. This curie map has been wrong in the past when people did record the preferred prefix correctly. 

Anyways triffle. What was I thinking. BRAVENESS! AUTOMATION!

matentzn (approved):
Nice one!

jamesaoverton:
Much appreciated. This looks good, but I have one question. In our GitHub repo settings we have `master` as a protected branch, requiring PR approval. Does the GitHub Action ignore the branch protection? I think ignoring the branch protection is the behaviour we want, but I don't know what will actually happen.

jamesaoverton (approved):

jamesaoverton:
Ok, let's see how this goes...

matentzn:
currently not, see #1662

cthoyt:
Screen Shot 2021-11-17 at 19 05 58

I realized in PyKEEN we have it set up so the bot that makes automatic pushes has explicit access rights via its membership in a team. So we can do the same for the GitHub Actions bot in this rep

matentzn:
Will this work with 

- [X] Include administrators?

image

jamesaoverton:
Ok. I implemented some of #1404, so now there are just six Admins. Then I removed the "Include administrators" restriction. Maybe it will work now?

There's also the "Allow specified actors to bypass pull request requirements" option. But how would I specify GitHub Actions is allowed to bypass?

jamesaoverton:
Failed again: https://github.com/OBOFoundry/OBOFoundry.github.io/runs/4253718821?check_suite_focus=true

matentzn:
@cthoyt do we have to:

- Check the option for restrict who can push to matching branches
- Add github actions bot to admin team?

PR #1659, opened 2021-11-16
Fix wrong mixed case prefix message

### LINKED ISSUE #1658: Misleading message "This ontology uses a mixed-case prefix. Identifiers should..."
Example:

http://obofoundry.org/ontology/foodon

image

> This ontology uses a mixed-case prefix. Identifiers and URIs should use "FOODON"

This is confusing as in fact foodon doesn't use a mixed-case prefix

Originally we *only* provided preferredPrefix for exceptions but in #1621 this was added to all

I would like to keep the notification, but have it only show up for non-standard preferred prefixes

matentzn:
Fixes #1658 

Someone needs to double check the jinja, as I am not a wizard. I cant test it as I cant get jeckyll to run in the required way.

### DIFF
--- a/_layouts/ontology_detail.html
+++ b/_layouts/ontology_detail.html
@@ -46,12 +46,14 @@ layout: default
   {% endif %}
 
   {% if page.preferredPrefix %}
+  {% if page.preferredPrefix|upper != page.preferredPrefix %}
   <div>
     <div class="alert alert-info" role="alert">
       This ontology uses a mixed-case prefix. Identifiers and URIs
       should use "{{page.preferredPrefix}}"
     </div>
   </div>
+  {% endif %}
   {% endif %}
 
   <div class="col-md-12">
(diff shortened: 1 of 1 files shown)


_layouts/ontology_detail.html
```diff
@@ -46,13 +46,15 @@ <h1>
   {% endif %}
 
   {% if page.preferredPrefix %}
+  {% if page.preferredPrefix|upper != page.preferredPrefix %}
```
cthoyt, line 49: you could combine this with the previous if statement using `and` for brevity

matentzn: Cheers

matentzn:
I tested this with https://cryptic-cliffs-32040.herokuapp.com/ and it works fine!

cthoyt (approved):

jamesaoverton (approved):

PR #1660, opened 2021-11-17
Add license to EPSO

matentzn:
@phwegner

PR #1661, opened 2021-11-17
Add new tests

cthoyt:
The old tests in the OBO Foundry have a lot to be desired, from both the maintainability, correctness, and utility perspectives. For example, even though the [openness principle (FP-001)](http://www.obofoundry.org/principles/fp-001-open.html) requires both that a license is given and that it is labeled as either `CC BY 3.0`, `CC BY 4.0`, or `CC0 1.0`. The current tests run by make test gives some logging, but it does not act as a testing framework and signal CI to fail. This issue just became apparent when #1646 was accidentally merged without license information, requiring a new PR #1660. Since all reviews require precious time, automating as much of it as possible seems paramount.

This PR adds a new python module containing a single test in a new file (`tests/test_integrity.py`) with an accompanying new CI workflow that addresses this issue as a demonstration. It includes several additional configurations for assessing the code quality of this file and any future python modules put in `tests/` that are run in CI as well in order to support keeping this code high quality and maintainable.

This PR does not change any of the old test suite nor duplicate any of its functionality.

## Details

| File | Purpose |
|-----|----------|
| `tox.ini` | It's the more python-specific version of `make`. It's much more streamlined, readable, and easy to add things to. This specific file has configurations for automatic formatting (i.e., applying `black` and `isort`), for running linting (i.e., `flake8`), and running the tests in the `tests/` directory inside inside their own virtual environments, while also taking care of installation of the necessary dependencies |
| `.flake8` | Configuration for the flake8 linter |
| `pyproject.toml` | Configuration for black and isort. |
| `.github/workflows/tests.yml` | An additional GitHub Actions configuration that runs in parallel with the old tests - it runs each of the things mentioned in the `tox.ini` file |
| `tests/test_integrity.py` | A Python module implementing unit tests using a first-class unit testing framework (i.e., `unittests`). Specifically, it enforces licenses conform to the registry schema. It only has a single dependency (for now) that is automatically installed with tox. It can be run with `tox -e py` from the root. |

### DIFF
--- a/tests/test_integrity.py
+++ b/tests/test_integrity.py
@@ -1,0 +1,97 @@
+# -*- coding: utf-8 -*-
+
+"""Unit tests for metadata integrity.
+
+Run from the root of the repository using ``tox`` with:
+
+.. code-block:: sh
+
+    $ pip install tox
+    $ tox
+"""
+
+import json
+import os
+import unittest
+from io import StringIO
+from pathlib import Path
+from typing import Union
+
+import yaml
+
+HERE = Path(__file__).parent.resolve()
+ROOT = HERE.parent.resolve()
+
+SCHEMA_PATH = ROOT.joinpath("util", "schema", "registry_schema.json")
+SCHEMA = json.loads(SCHEMA_PATH.read_text())
+
+#: These are the valid licenses listed in the registry schema file
+VALID_LICENSE_LABELS = {
+    value
+    for entry in SCHEMA["properties"]["license"]["anyOf"]
+    for value in entry["properties"]["label"]["enum"]
+}
+VALID_LICENSE_URLS = {
+    value
+    for entry in SCHEMA["properties"]["license"]["anyOf"]
+    for value in entry["properties"]["url"]["enum"]
+}
+#: These ontologies have invalid licenses, but they're grandfathered in
+LEGACY_LICENSE_PREFIXES = {
+    "GSSO",
+    "HP",
+    "KISAO",
+    "MAMO",
+    "SBO",
+    "SCDO",
+    "TXPO",
+}
+
+
+def load_all():
+    """Load all metadata files in a dictionary from lowercased OBO Foundry IDs to metadata."""
+    rv = {}
+    for path in ROOT.joinpath("ontology").glob("*.md"):
+        data = load(path)
+        rv[data["id"]] = data
+    return rv
+
+
+def load(path: Union[str, os.PathLike]):
+    """Load metadata from a single markdown file."""
+    with open(path) as file:
+        lines = [line.rstrip("\n") for line in file]
+    idx = min(i for i, line in enumerate(lines[1:], start=1) if line == "---")
+    return yaml.safe_load(StringIO("\n".join(lines[1:idx])))
+
+
+class TestIntegrity(unittest.TestCase):
+    """A test case for checking metadata integrity."""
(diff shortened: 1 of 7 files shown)


jamesaoverton:
This completely changes the structure of the validation system and its dependencies. It will require thorough discussion with the technical working group.

There are reasons for the existing system, including a shared set of requirements used by the PURL system and other major OBO projects that TWG members contribute to, as well as historical reasons.

cthoyt:
I disagree, this doesn't change anything that exists already, but only adds on top of it.

For example, even though the openness principle requires both that a license is given and that it is either CC BY 3.0, CC BY 4.0, or CC0 1.0. The current tests run by `make test` gives some logging, but it does not act as a testing framework and signal CI to fail. This issue just became apparent when #1646 was accidentally merged without license information, requiring a new PR #1660. Since all reviews require precious time, automating as much of it as possible seems paramount.

jamesaoverton:
> I disagree, this doesn't change anything that exists already, but only adds on top of it.

It adds a *second* validation system that works completely differently, with different dependencies. If I need to change the validation, which of these two systems should I change?

Everybody has their preferred tools, but OBO is a big project. The TWG members who have committed to maintaining these systems have different skill sets. People come and go. But the code and data can last for a long time. We all need to look at the big picture.

matentzn:
Am I correct in the assumption that this PR has essentially two things:

1. substantial code listing formatting checking (black/flake8)
2. This `tests/test_integrity.py` which adds some additional checks on the metadata? Like a sort of "additional schema" check or some such?

matentzn:
I am assigning you @cthoyt while we work through the details of this PR

cthoyt:
> Am I correct in the assumption that this PR has essentially two things:
> 
>     1. substantial code listing formatting checking (black/flake8)
> 
>     2. This `tests/test_integrity.py` which adds some additional checks on the metadata? Like a sort of "additional schema" check or some such?

Yes. This PR adds a single test (in `tests/test_integrity.py`) which enforces the correct labels are used for ontology licenses. This is orthogonal to the functionality of the current testing suite, which does not currently cause the CI build to fail when a new file is added with invalid/missing license metadata.

Instead of adding this test to one of the existing modules, it adds it to a new one in a new folder called `tests/`. This PR also adds some linting/format checking configuration that enforces files in this new folder are written in the highest possible quality (without judging any existing code). It adds all of these in a separate CI workflow, which runs in parallel to the existing tests.

matentzn:
I looked this over, and I think I am down with the proposed changes. @jamesaoverton we will work through this PR and any associated risks on our next 1:1(s). 

As always, thank you @cthoyt for your help and enthusiasm!

tests/test_integrity.py
```diff
@@ -0,0 +1,97 @@
...
+    for value in entry["properties"]["label"]["enum"]
+}
+VALID_LICENSE_URLS = {
+    value
+    for entry in SCHEMA["properties"]["license"]["anyOf"]
+    for value in entry["properties"]["url"]["enum"]
+}
+#: These ontologies have invalid licenses, but they're grandfathered in
+LEGACY_LICENSE_PREFIXES = {
+    "GSSO",
```
cmungall, line 41: be good to have a separate yaml for this

matentzn: (this PR is now moved #1684)

PR #1663, opened 2021-11-17
add link to UOM

cmungall:


matentzn (approved):
I am assuming this is fine with the UO team..

Issue #1662, opened 2021-11-17
Branch protection rules prevent auto-push of docs


matentzn:
See https://github.com/OBOFoundry/OBOFoundry.github.io/runs/4241619920?check_suite_focus=true

PR #1666, opened 2021-11-18
Manually updating files while autopush is blocked

matentzn:
While https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1662 blocks auto push

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -912,9 +912,9 @@ ontologies:
   tracker: https://github.com/ApolloDev/apollo-sv/issues
 - activity_status: active
   contact:
-    email: card@mcmaster.ca
-    github: raphenya
-    label: CARD
+    email: mcarthua@mcmaster.ca
+    github: agmcarthur
+    label: Andrew G. McArthur
   description: Antibiotic resistance genes and mutations
   homepage: https://github.com/arpcard/aro
   id: aro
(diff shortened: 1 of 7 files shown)


matentzn:
For the 10th time (sorry - I don't know how I will ever remember) - how exactly is the site deployed? This seems to be metadata only - is there a job that will deploy the markdown page somehow?

balhoff (approved):

jamesaoverton:
We use GitHub Pages with one special hack. On a normal GitHub Pages (i.e. Jekyll) site the `_config.yml` file is very small, and it just configures a theme and maybe a few global variables such as a title string. For example, [ROBOT's `docs/_config.yml`](https://github.com/ontodev/robot/blob/master/docs/_config.yml) just sets the theme.

The hack is that we add all the registry data into our `_config.yml`. This makes all the registry data accessible in a `site.ontologies` global variable that the templates can use. For example, the main table loops over `site.ontologies`: https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/_includes/ontology_table.html#L4

GitHub Pages won't update the `_config.yml` by itself. We use `make` to do that, and to update the various combined metadata files in `registry/`.

matentzn:
Thaaank you!

PR #1679, opened 2021-12-07
Update fypo.md

mah11:
update FYPO contact; add S. japonicus


mah11:
FYPO is now used for both S. pombe and S. japonicus. Is there a way to record that somewhere?

matentzn (approved):

PR #1683, opened 2021-12-08
Adding TOX as a testing framework (under consideration).

matentzn:
For a discussion about this PR see:

https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1661
#1684 

For now, I want to just add the actual test @cthoyt is proposing, but need help #1685


cthoyt:
@matentzn I think this PR can be dismissed as the tox/QA configuration and the specific use case have both been added in external PRs

matentzn:
Dismissing this according to @cthoyt recommendation.

Issue #1684, opened 2021-12-08
Migrating the testing system to tox


matentzn:
@cthoyt has converted me in the last year to using tox (and many other associated tools) to improve my python code 100 fold. He is proposing to migrate our OBO Foundry testing framework to tox, and I am happy to starting thinking along these lines. However, there are some concerns about maintainability (we have very few resources to fix stuff if something breaks). The current system is based on a mix of json schema checking and ad-hoc python code with lots of custom logic - which all by itself is not super easy to extend - but doable. 

I think we should seriously consider updating our pipelines, but before we do so, we need to understand what is needed to migrate everything, who will support it, who will pay for the migration, and what the risks are.

@cthoyt provided a summary of his additions here: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1661

Anyone who feels like they have experiences for and against employing tox for testing please feel free to contribute here. What we do _not_ want is to start adding tox _on top_ of what is already there - if we go for it, then we will want to migrate what is currently there over as well, and have one single pipeline to handle everything.

matentzn:
(one tiny issues is that I dont know how many hours a good programmer would need to toxify the hole pipeline - that's a bit of an issue).

jamesaoverton:
I don't want to sound like an old grump, but I've learned some painful lessons that I don't care to repeat. I've been contributing to OBO for more than a decade now, and somewhere along the line my role changed from only "pushing for improvements" to largely "protecting what we've built". So I end up saying "no" a lot. I really, really hate that. I remember how much it stung (and still does) to hear people push back against my great ideas for improving OBO. But I feel personally responsible for OBO infrastructure, so I want to share my personal opinions.

Chris Mungall wrote the original code for this repo, and many people have helped out, but my team has been responsible for a lot of the work and maintenance. We do not have direct funding for OBO infrastructure. I'm not happy with the current code, but at least it's a known quantity. I would prefer to improve the current code and tooling incrementally.

My team works with Python but also several other languages every day. One reason we use GNU Make is that it's not tied to any one language. We use linters and formatters and testing tools, such as 'flake8' and 'black' and 'pytest' For Python, but we run them from Make, like the rest of our automation tasks. Another reason is that Make is *really* mature -- it's been around longer than I have, it's installed everywhere, and it's not going anywhere.

[...]

PR #1686, opened 2021-12-08
Add minimal `tox.ini` to demonstate future potential code quality checking.

cthoyt:
## Motivation

Maintaining code is hard, and OBO runs on a tight budget, which means that we want to maximize how reliable the code that's already in the ecosystem is while minimizing the effort it takes to read, modify, or improve old code. Additionally, we want to reduce the amount of effort it takes to do code review.

Using programmatic tools to do code quality assurance helps get us towards all of those goals. I realize that #1661 was a bit too much of a jerk in that direction, so this PR is the first of a series that will spoon feed some new ideas to this repo/community.

## What this PR does

This PR is a minimal demonstration for #1684, but I don't think for code quality assurance that it's necessary to consider "moving over" to tox since this is a completely new thing in this repo.

More specifically, this pull request adds a minimal configuration for using `tox` to run the `flake8` code quality checker.

1. It adds a `tox.ini` configuration for [`tox`](https://tox.wiki/en/latest/index.html), which is like `make`, but more python-specific. It's based on the INI config file format and is much more streamlined, readable, and easy to add things to in "environments" that correspond to headers that look like `[testenv:XXX]`
2. Adds an environment `[testenv:flake8]` that takes care of installing and running the [`flake8`](https://flake8.pycqa.org/en/latest/) code quality checker. Right now, the `flake8` tool is only applied to python modules in the `utils/` folder and only the check that triple double quotes are used properly are reported (via the `--select D300` option). This is just a demo so I didn't want to change very many files, but flake8 is [hugely extensible](https://cthoyt.com/2020/04/25/how-to-code-with-me-flake8.html) and could be used to do more in a later PR.
3. Tox can be run with
    ```shell
    $ pip install tox
    $ tox -e lint
    ```
    so I did that and made updates in 3 files that used triple single quotes instead of triple double quotes. This satisfied the flake8 tool
4. Adds a new GitHub Actions check in `.github/workflows/qa.yml` that runs in parallel to the existing tests and is 100% orthogonal. Right now, it effectively installs and runs tox to apply the flake8 checks. Later, it's easy to extend this configuration with additional tox environments for other code quality checks, if desired


matentzn:
Very nice, highly appreciated. There were some concerns on `tox` changing and us having to scramble to fix running code because something in the tox spec changed. Could you tell us a bit about that? If we were to go with your proposal, what are the odds of someone one day waking up and having to change stuff because tox has changed? Maybe my concerns are blue eyed, but a bit of a better sense of the stability of the added dependencies would be good.

cthoyt:
> Very nice, highly appreciated. There were some concerns on `tox` changing and us having to scramble to fix running code because something in the tox spec changed. Could you tell us a bit about that? If we were to go with your proposal, what are the odds of someone one day waking up and having to change stuff because tox has changed? Maybe my concerns are blue eyed, but a bit of a better sense of the stability of the added dependencies would be good.

Same as with make, tox knows that it can't change its schema since people have to rely on this stuff. Tox itself is mature software and seems to have an active community providing bug fixes (see https://tox.wiki/en/latest/changelog.html)

Then again, it's also possible to pin the `tox` version as well as the dependency versions inside for maximum reproducibility.

cmungall:
This is great, I fully support the toxification, I have been moving a lot of other repos to tox and flake8

matentzn:
@cmungall will you commit developer time to future changes caused by issues relating to tox when they arise?

jamesaoverton:
I like `flake8`, `black`, and automated formatting and linting tools in general, even though I get a lot complaints when I insist on them. We can easily run these from `make`.

I don't understand  the benefit `tox` brings. It seems wasteful to use our limited free credits to run a second set of GitHub Actions on every commit.

If my team is responsible for this code base, then this is not the way I want to do things. If other TWG members want to support this approach, I will be delighted to step back. I asked about this on #1684.

matentzn:
@cthoyt can you provide a list of reasons why `tox` should be introduced here? I.e. a clear explanation of what would be the downside of not introducing tox at all - i.e. could we achieve the same result by simply having flake8, black, pytest etc to the Makefile? You think it is easier to maintain? If so, how? Can we get an example? Does it add capabilities _that we need_ that can't be achieved otherwise? No one here is arguing that we need to add the additional tests - but we need to do some work convincing each other that we cant achieve that in the current, albeit fragile, system.

IMO, this is not a matter we can vote on (from an SOP standpoint we could, but I would not take part in such a vote). The OBO technical working group is coordinated by @jamesaoverton, @beckyjackson, @cmungall and me (other [members exist](https://obofoundry.org/docs/Membership.html) but they operate on specific technical issues not related to this one here). For the four of us, such decisions must be, in my opinion, unanimous. I am in favour of your changes, but for me, it's much more important to have @jamesaoverton continued support - which means we need to address his concerns.

cmungall:
> @cmungall will you commit developer time to future changes caused by issues relating to tox when they arise?

Yes, I can. We are standadizing on tox for a lot of our repos.

But of course, I want us all to be happy with all technical decisions. I agree with the general points @matentzn is making.
I've read this ticket and #1684 (sorry for being tardy). I think I have a good sense of arguments in favor of tox, I like it, I like things like being able to easily test different versions. But like James I also don't like the churn in the python ecosystem and have been burned by the choosing the wrong flavor of the month.

To put this in perspective though, if this is our main technical disagreement, then we are doing pretty good! Hey, we all agree python was the best choice for this repo and are glad we didn't do it in java or perl! Think how much pain there would be in backing out of that decision now, and difficulties in balancing the benefits of switching vs maintenance.

To be honest (and sorry if this annoys everyone) but I am not sure I see a huge amount riding on this decision in the short term. I think tox will definitely have benefits but maybe not orders of magnitude. At the same time I don't think changing our own workflows to use a new tool will carry much cognitive overhead or have any drain on resources, especially if others commit to doing the refactor and making changes to CONTRIBUTING.md to help other developers. Worst case scenario, we decide to back out and go back to direct makefile calls, maybe I am missing something but I think it's not such a big lift (and I will commit to doing if Charlie disappears).

I don't think back and forth in these issues is the best way to deal with this, maybe we can be happy with one decision or another in the short term, and then I think we should have a TWG meet early next year. I'd like to plan a few things including properly modularizing some of the code here and releasing on pypi - and I think having a shared idea of long term goals will help us make unanimous decisions in the short term.

jamesaoverton:
> > @cmungall will you commit developer time to future changes caused by issues relating to tox when they arise?
> 
> Yes, I can. We are standadizing on tox for a lot of our repos.

That's all I wanted to hear.

matentzn:
Alright, we are then in a rough agreement. 

Adressing some of the remaining technical concerns:

- [ ] Exhaution of GHA credits: The github actions are super fast - if they will exhaust our credits, we are in a big problem well beyond this repo, because all our repos rely on this.
- [ ] Maintainance moving forward: Seems like @cmungall will commit dev time if issues arise
- [ ] What remains is a to study the advantages of tox over the old way of doing things. [This](https://christophergs.com/python/2020/04/12/python-tox-why-use-it-and-tutorial/) seems a good post explaining the issue.  For example, as already pointed out by @cmungall and @cthoyt it allows us to easily test the infrastructure on many different python versions, which guards as against decay and lets us more seamlessly upgrade out code over time, when dependencies need upgrading. @jamesaoverton if you want to read that blog post, and see where you are and are not convinced, maybe that could bring some clarity.

Please everyone across TWG, add any other technical concerns, lets list them in checklist format and address them one by one.

jamesaoverton:
I already agreed to this.

The checklist of technical concerns should include:

- [ ] Duplication of code and functionality.

In response to checklist above:

- GitHub Actions credits: I was mistaken: ["GitHub Actions usage is free for both public repositories and self-hosted runners."](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) although I expect that will change at some point. Note that credits are per-organization, and we have multiple repos in this organization that need CI. I think it makes more sense to have a single GHA for testing. 99% of our commits do not change the Python code.

- tox vs make: The argument in that blog post seems to be that Python packaging and testing is so bad that you need yet another tool to coordinate the existing tools and manage it safely. This repository is a custom (somewhat hacky) static site generator. It's not a library or application that anyone else should have to run. So I don't see why we need to support multiple Python versions. Upgrading these scripts to new Python versions would take ten minutes every few years, tops. If we're still non-Python tools in the workflow, I would still use Make and I would have just added `flake8` and whatever to the `Makefile`. Also, didn't everyone spend years trying to convince me that we need Docker for everything, to solve exactly the same packaging and testing problems we're talking about again? Hindsight has not shown that to be a clear win.

The only way that OBO has survived this long on zero funding is by keeping things *stupidly*, *embarrassingly* simple. I agree that this PR isn't a big change, but it's obviously just a first step.

I already agreed to this. I've had my say and I'm already repeating myself, so I will step back from this discussion.

cmungall (approved):

PR #1694, opened 2021-12-15
Add Ontology status to docs

matentzn:
According to our discussion, here is a little page we can refer to with the preliminary definitions.

### DIFF
--- a/docs/OntologyStatus.md
+++ b/docs/OntologyStatus.md
@@ -1,0 +1,12 @@
+# Ontology status
+
+Info: this page is non-normative; definitions may change.
+
+An ontology can be active, inactive, orphaned and obsolete
+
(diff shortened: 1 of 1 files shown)


dosumis (approved):

jamesaoverton (approved):

nataled (changes requested):
All in all looks good, but I suggest two minor changes:

1) An ontology can be active, inactive, orphaned and obsolete - change the 'and' to 'or' (these statuses are mutually exclusive)
2) ...for continued discussions on the subject - change 'continued discussions' to 'discussion' (at some point the discussion will stop and we'll need to change it; just saying 'discussion' covers both current and past)

matentzn:
Thank you @nataled done!

nataled (approved):

nataled:
Ready for merging.

PR #1698, opened 2021-12-15
Add Ontology status to navbar

matentzn:


nataled (changes requested):
I'd change it to
 What does it mean for an ontology to be flagged as 'inactive', 'orphaned' or 'obsolete'?

The reason I took out 'active' is because none of the active ones are actually flagged in that way. I nonetheless advocate including it in the answer because it provides a basis upon which the other statuses can be defined and compared.

By the way, though the ontology list page uses 'obsolete', when looking at the individual ontology's page the word used is 'deprecated'. I think the latter should be changed to obsolete.

matentzn:
Thank you @nataled

PR #1700, opened 2021-12-15
Update OntologyStatus.md

matentzn:
It seems that otherwise the links dont get a menu!

https://obofoundry.org/ontology/fix.html
https://obofoundry.org/docs/OntologyStatus.html


jamesaoverton (approved):
Yes, I think `layout: doc` will fix the problem.

nataled (commented):
Approved. (not sure why I'm not seeing the usual interface for that)

PR #1703, opened 2021-12-16
Update and rename epso.md to epio.md

phwegner:
Changed repository and renamed EPSO to EPIO. So we need a new namespace "EPIO" instead of "EPSO" due to redundancy. (EPSO already exists). Sorry for the circumstances. I will also edit the purl file.

### DIFF
--- a/ontology/epio.md
+++ b/ontology/epio.md
@@ -1,6 +1,6 @@
 ---
 layout: ontology_detail
-id: epso
+id: epio
 label: Epilepsy Ontology
 title: Epilepsy Ontology
 description: A application driven Epilepsy Ontology with official terms from the ILAE.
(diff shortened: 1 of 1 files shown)


matentzn:
That is much more complicated. This will probably take months as it will need another round of review and some discussion at the OBO committee..

matentzn:
Can you point us to the place where EPSO exists?

phwegner:
https://github.com/SCAI-BIO/EpilepsyOntology (New)
https://github.com/phwegner/Epilepsyontology (Ols)

phwegner:
But concerning the revision. It is the exact same file but another name and another place.

phwegner:
@matentzn Can we then stick to EPSO until EPIO is approved ?

matentzn:
No I mean, why do we have to change the prefix?

phwegner:
One of our team members notified us that EPSO is already taken by another Ontology. So not in OBO obviously but the abbreviation itself is already in use.

matentzn:
Can you show us which one?

phwegner:
https://bioportal.bioontology.org/ontologies/EPSO

phwegner:
Even though Satya is a collaborator with the OBO EPSO Ontology those are two distinct Ontologies...

matentzn:
What about all the IDs in EPSO, the term idsyy did you change them as well?

phwegner:
We are doing that right now. So in the meantime will we stay with the EPSO namespace?

matentzn:
I think your request may be valid enough to sort it out.. :/ Give us a day to discuss this.

phwegner:
Thank you very much !

matentzn (approved):

matentzn:
We reviewed this case and decided to approve it due to its _very specific_ circumstances. Generally, such requests will not approved, because there are a lot of architectural and conceptual problems that come with it. The specific circumstances for allowing this prefix change are:

- EPIO was only accepted recently and its identifiers are therefore unlikely to be reduced widely
- The conflicting Bioportal ontology is very similar in scope and  likely this will cause a lot of confusion for users of biomedical ontologies as a whole

Please confirm the following to things and I will merge the PR
- [ ] [This](https://bioportal.bioontology.org/ontologies/EPIO) is you
- [ ] You have migrated all your terms from EPSO ids to EPIO Ids

Once you confirm both, we will deal with the rest. Thanks

phwegner:
Thank you for the fast response! 

I can confirm both I just added the correct ids and added the respective release. 
The version in bioportal is ours.

PR #1714, opened 2021-12-21
DRAFT, DO NOT MERGE Create fp-013-notification.md

nataled:
Straw man for new principle

### DIFF
--- a/principles/fp-013-notification.md
+++ b/principles/fp-013-notification.md
@@ -1,0 +1,70 @@
+---
+layout: principle
+id: fp-013-notification
+title: Notification of Changes (principle 13)
+---
+
+Summary
+-------
+
+Ontologies SHOULD announce changes to relevant stakeholders and collaborators ahead of release.
+
+Purpose
+-------
+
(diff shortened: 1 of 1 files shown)


matentzn:
Wow this is some principle.. :) I love this, but this will put a huge burden on ontology developers.. If they use GitHub, maybe we can promote an ongoing diff report that is published at some predefined purl location like /diff.txt or some such.. It will also be massively complex otherwise to automatically test this. I am not against it, just.. worried its quite a tough one to meet!

nataled:
@matentzn as the pull request focuses on wording, I'm going to put your comment on the general ticket for this, and respond there.

matentzn:
Can we add action items here for the OBO Ops Commitee - what needs doing wrt to this Draft PR (just to save a bit of time).

nlharris:
Removing "attn: OFOC call" label for now.

cstoeckert (approved):
Reviewed on EWG call Dec. 6, 2022. Ready for review by Ops committee.

Issue #1722, opened 2022-01-04
obsolete should supersede inactive


cmungall:
Following from

 * #1721 

There is a lovely rainbow of colors here:

https://obofoundry.org/ontology/ehda

![image](https://user-images.githubusercontent.com/50745/148106439-2abcf76c-15bf-411d-a53e-bf23c3b310b6.png)

I think obsolete implies inactive, we should either constrain the schema to allow either but not both, or we should filter at the UI level

The mixed-case indicator is a false positive but that is reported in another issue

Another oddity here is that the replacement for ehda (ehdaa2) is inactive (not not obsolete). I think this is OK and valid but it may be confusing to some. We may also want to make suggestions for active alternatives for ehdaa2 (this would be uberon or fma, the latter which now includes more embryonic terms but is still less complete than ehdaa2). Some of this may be best done in plain old fashioned text.

matentzn:
- [X] inactive and orphaned banners should only show up when not deprecated (https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1738).
- [X] irregular preferred prefix check only if preferred prefix is set (https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1738).
- [ ] Should inactive ontologies be considered?

nlharris:
did anyone make a decision about the last point above?

matentzn:
I think no. 

@cthoyt Do you know if there are ontologies being referred to by consider that are marked as `inactive`?

cthoyt:
```python
import requests
import yaml

URL = "https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/registry/ontologies.yml"


def main():
    records = yaml.safe_load(requests.get(URL).text)["ontologies"]
    prefixes = {
        record["id"]
        for record in records
        if record["activity_status"] in {"inactive"}
    }
    for record in records:
        if record.get("replaced_by") in prefixes:
            print(record["id"])


if __name__ == '__main__':
    main()
```

Gives back:
- ehda
- ehdaa

matentzn:
If you would be so kind to add a test to disallow this and remove the consider relation from these two, that would be very appreciated :) This will close this ticket here as well.

cmungall:
ehda and ehdaa were replaced by ehdaa2. we should not lose this information

cthoyt:
i have no opinion one way or another. The text inside ehdaa2 about alternatives isn't super actionable, though - it reads

[...]

PR #1724, opened 2022-01-04
Added status notes about potential alternatives to EHDAA2

cmungall:
See #1722


cthoyt:
Could you use a structured field like `replaced_by` to point to these preferred ontologies?

cmungall:
We use replaced_by consistent with the same property in IAO/OMO, i.e for drop in replacements. I think it would be misleading to say EHDAA2 is replaced by either uberon or fma.

I suggest we add a field to the schema `consider` consistent with the oboInOwl property, for cases like this

PR #1733, opened 2022-01-11
Add Python code quality check

cthoyt:
Building on the new code quality assurance framework introduced in #1686, this PR does the following:

1. Adds linting configuation in `pyproject.toml` for `black` and `isort`
2. Adds run configuration for linting in `tox.ini`
3. Adds flake8 plugins for ensuring `black` and `isort` were applied
4. Applies `black` and `isort` to all Python files
5. Updates the CONTRIBUTING.md file to tell people to run these linters on their code

## Why all of this code churn?

Right now, it's very tough to read the Python code due to stylistic differences from the many contributors (not to say anyone has bad code style, but just there's an issue when it has a lot of variety in a given project). Using a linter makes the style more consistent, the code more approachable, and ultimately, more maintainable.

PR #1741, opened 2022-01-17
Add missing repositories and mark `repository` as required

### LINKED ISSUE #1730: Require `repository` for new ontologies
Currently, only an issue tracker is required. I think it makes sense to also require new ontologies to provide a repository (e.g., on GitHub (highly preferred), GitLab, BitBucket, etc.)

Implicitly, this will encourage all new ontologies to use a public repository as their main place for curation.

This was motivated by the recent new ontology request for ELO (#1729), which did not include a repository. It seems to be the case that a potential submitter is not aware this is important, because it is not included in the form.

cthoyt:
Closes #1730 

After these three, all non-obsolete, non-orphaned, non-inactive ontologies have a repository. This PR updates the schema and new ontology request page accordingly. This PR is blocked by the PR to mark FMA as inactive, since there isn't an active tracker nor repository associated with FMA.

## Blocked By

- [ ] #1752 or #1458


.github/ISSUE_TEMPLATE/new-ontology-request.md
```diff
@@ -26,6 +26,7 @@ GitHub username:
 
 ## Issue tracker
 
+## Version Controlled Repository
```
cthoyt, line 29: consider updating wording here

matentzn: Suggestion?

ontology/fma.md
```diff
@@ -29,6 +29,7 @@ build:
   insert_ontology_id: true
   method: obo2owl
 tracker: https://sourceforge.net/p/obo/foundational-model-of-anatomy-fma-requests/
+repository: https://sourceforge.net/p/obo/
```
matentzn, line 32: ? You mean as a placeholder?

cthoyt: I don't think this ontology actually has a repository, so I guess this is just a placeholder.

cmungall:
I think better to leave out, as this is misleading to leave the generic
abandoned sourceforge repo there

cthoyt:
So can FMA be considered as abandoned? Somehow it's hosted at http://ontologies.berkeleybop.org/fma.owl (based on https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/fma.yml) but I don't know how to follow this redirect.

cmungall:
Let's take any FMA discussion to this issue: #21

.github/ISSUE_TEMPLATE/new-ontology-request.md
matentzn: Just make PR if you have a suggestion for improvement.

cthoyt:
Alright, I left FMA out. I will wait for #1733 to get merged so I can add a special case to the schema check to skip this error.

matentzn:
I kinda prefer your previous solution with the stub rather than adding some conditionals somewhere in some code unrelated to the fact that FMA has no repo. I even prefer to mark it inactive for the time being.

cthoyt:
I agree but I think a lot of people might be offended by saying FMA is inactive. @cmungall could we do this?

cmungall:
According to @trapperkeeper (who is a better technical contact than onard) way back here:

https://github.com/OBOFoundry/OBOFoundry.github.io/issues/21#issuecomment-157865336

There is a tracker: https://bitbucket.org/uwsig/fma/issues

It has not seen any use, but maybe people would have submitted there if we had included the metadata in OBO.

> I agree but I think a lot of people might be offended by saying FMA is inactive

It's not about offense, it's about accuracy. https://bioportal.bioontology.org/ontologies/FMA - updated <2years ago, this is more active than many OBO ontologies currently marked active, but if our criteria says this makes it inactive we should declare it as such

I have emailed FMA folks for clarity and cc'd you all, let's see what comes of that, but I think a likely resolution will be to create a new category, where we can obsolete a *record* of an ontology without saying the ontology is obsolete.

matentzn (approved):

PR #1742, opened 2022-01-17
Add information for people that wish to register a prefix but are not (yet) ready for OBO Foundry

matentzn:
An FAQ on what to do if your ontology is _not_ eligible for OBO Foundry, but the requestor wants to ensure that the prefix is not easily snatched.

### DIFF
--- a/docs/ReserveNamespace.md
+++ b/docs/ReserveNamespace.md
@@ -1,0 +1,5 @@
+# My ontology is not ready for OBO Foundry, can I reserve a namespace anyway?
+
+If your ontology is not eligible for inclusion in the OBO Foundry you can not directly reserve a namespace in the OBO Foundry. 
+However, if you want to make sure that no one uses your prefix, you can [register your prefix at the Bioregistry](https://github.com/biopragmatics/bioregistry/issues/new/choose).
+The updated [OBO policy on IDs](https://obofoundry.org/id-policy) ensures that new prefix requests must not clash with existing prefixes registered at the Bioregistry.
--- a/docs/ReserveNamespace.md
+++ b/docs/ReserveNamespace.md
@@ -1,3 +1,9 @@
+---
+layout: doc
+id: ReserveNamespace
+title: Reserve Namespace outside of OBO
+---
+
 # My ontology is not ready for OBO Foundry, can I reserve a namespace anyway?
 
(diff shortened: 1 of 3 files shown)


docs/ReserveNamespace.md
```diff
@@ -0,0 +1,5 @@
+# My ontology is not ready for OBO Foundry, can I reserve a namespace anyway?
+
+If your ontology is not eligible for inclusion in the OBO Foundry you can not directly reserve a namespace in the OBO Foundry. 
+However, if you want to make sure that no one uses your prefix, you can [register your prefix at the Bioregistry](https://github.com/biopragmatics/bioregistry/issues/new/choose).
+The updated [OBO policy on IDs](https://obofoundry.org/id-policy) ensures that new prefix requests must not clash with existing prefixes registered at the Bioregistry.
```
cthoyt, line 5: I guess this needs a clause that the only case where this is okay is if you yourself pre-registered the prefix

cthoyt: Otherwise there’s a weird loophole here where if you preregister on bioregistry then you can’t add to obo foundry. I suppose there needs to be some kind of arbitration where you prove you’re the same person and it’s the same resource

matentzn: Lol good point. I will leave that to one of the OBO editorial people to clarify, who are much better at formulating this stuff precisely (@nataled).

nataled: Is it enough that the OBO request simply be made by the same person listed as 'contact' in the bioregistry entry, as verified by the same name and/or github handle? Or is that too high a bar? (I can imagine someone registering the prefix, but leaving the development team before the ontology is ready for the OBO Foundry).

cthoyt: I think it would be good enough to say that the person is also a representative of the resource. If it gets out of sync in obo foundry and the bioregistry who that is, then it’s a good time to make an update. It’s much faster to update bioregistry we well

matentzn: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1742/commits/30acee1f93e6c7f3fde66d7913c1042daecd2f17

cthoyt (approved):

nataled (approved):

PR #1752, opened 2022-01-22
Mark FMA as inactive

cthoyt:
There are a couple lines of discussion currently going with ex-FMA people via an email chain (unfortunately it wasn't really possible to get participation on any common public channels). Most of them are retired or unable/unmotivated/not funded to do some of the things necessary to keep FMA marked as "active". There are no current maintainers and no plans to do maintenance nor participate on a public tracker, therefore this ontology should be marked as inactive. If any of those change, then it could be marked again as active.

### DIFF
--- a/ontology/fma.md
+++ b/ontology/fma.md
@@ -29,7 +29,7 @@ build:
   insert_ontology_id: true
   method: obo2owl
 tracker: https://sourceforge.net/p/obo/foundational-model-of-anatomy-fma-requests/
-activity_status: active
+activity_status: inactive
 preferredPrefix: FMA
 ---
 


lschriml:
From the emails mentioned above,  John Gennari <gennari@uw.edu> - expressed that he wants the ontology to be kept active. And suggested a zoom call to discuss the issues. 
I think it is premature for the OBO Foundry to then move the ontology to inactive. 

Cheers,
Lynn

bpeters42:
I would disagree with Lynn, de facto FMA *is* currently inactive. And there is nothing wrong with us switching that status back to active if someone like John Gennari adopts the ontology, and commits to working on fixing the current issues. That would actually be the best possible outcome, and changing the status publicly (rather than lamenting it behind the scenes) could be a catalyst to make it happen.

cthoyt:
Thanks for the comment @bpeters42. I'd like to echo that there's a difference between people who have interest in a resource maintained and demonstrably doing the maintenance (e.g., there are discussions of a potential update but since FMA has no public issue tracker nor repository, this is not provable)

matentzn:
@cmungall The OBO meeting minutes claim that you will have a meeting with them "in a few weeks" - does this mean we have to wait merging this (as it blocks other PRs we want in)?

matentzn (approved):

matentzn:
It is my interpretation from the email that we can savely move the OBO record of FMA to inactive. @cmungall if you want, we can add some notes to the FMA description describing its complicated real state.

Issue #1668, opened 2021-11-19
Add check: `homepage` should be a required field for active ontologies


cthoyt:
I think some obsolete/inactive/orphaned ontologies don't have a homepage entry, but is it the case that all new ontologies should/must have a `homepage`? The current testing suite doesn't seem to communicate this by failing a test, but rather noting it as logging deep within the build https://github.com/OBOFoundry/OBOFoundry.github.io/runs/4263986283?check_suite_focus=true#step:5:18.

The following Python script identifies EPSO (newly added in #1646) as the only non-obsolote/inactive/orphaned ontology that is missing a `homepage` entry as of 2021-11-19:

```python
import requests
import yaml
from tabulate import tabulate

OBOFOUNDRY_URL = "https://raw.githubusercontent.com/OBOFoundry/OBOFoundry.github.io/master/registry/ontologies.yml"


def main():
    rows = []
    ontologies = yaml.full_load(requests.get(OBOFOUNDRY_URL).text)["ontologies"]
    headers = ["contact.name", "contact.email", "prefix", "repository"]
    for data in ontologies:
        if data.get("is_obsolete") or data.get("activity_status") in {"inactive", "orphaned"}:
            continue
        if "homepage" in data:
            continue
        prefix = data["id"]
        contact = data.get("contact", {})
        github = contact.get("github")
        rows.append((
            contact.get("label", ""),
            contact.get("email", ""),
            f"@{github}" if github else "",
            f"[`{prefix}`](http://www.obofoundry.org/ontology/{prefix})",
            data.get("repository", ""),
        ))
    rows = sorted(rows)
    print(tabulate(rows, headers=headers, tablefmt="github"))


if __name__ == '__main__':
    main()
```

matentzn:
I think it would be a good move to make a homepage required for active ontologies - I cant really see how you can justify there not being one. @phwegner could you add a homepage for EPSO?

cthoyt:
Closed by #1753

PR #1753, opened 2022-01-24
Mark `homepage` as a required property

cthoyt:
As a follow-up to #1668, this PR marks the `homepage` as an error-level required property, such that any future ontologies like EPIO that did not submit a homepage will cause the technical checks to fail and notify reviewers not to merge a new ontology request.

This PR also adds a stub homepage for the following repositories:
1. EPIO by duplicating its GitHub repository
2. FIX by including the ChEBI link, since it's part of ChEBI


matentzn (approved):

matentzn:
Great, thank you @cthoyt :)

PR #1758, opened 2022-01-26
Add exceptions comment to New Ontology Request OBO Dashboard Rule

matentzn:


nataled (changes requested):
change 'requirements are' to 'requirement is' and it should be good to go.

matentzn:
I am assuming you meant `requirement are`: "The only exceptions to this requirement are"

nataled (approved):

matentzn:
Thank you @nataled

PR #1766, opened 2022-01-26
Fix typo in PCO dependency

cthoyt:
Perhaps there should be metadata checks that all of these IDs point towards valid OBO Foundry IDs


matentzn:
Yeah would be good. No one has ever looked _that_ closely at the metadata :D

cthoyt:
When you automatically consume it and have a tool like the bioregistry to apply integrity checks, this stuff pops out of the woodwork

matentzn:
If after you are done with the metadata it isn't the most up to date and consistent open science metadata library I will be genuinely suprised.

PR #1776, opened 2022-01-28
Update schema validation to require `tracker` and `contact.github`

cthoyt:
This PR does two things:

1. Changes the `tracker` level to `error` so it's actually validated. Luckily, all active ontologies have trackers and this requires no change.
2. Adds the "github" field to required in the contact and update the text accordingly. 
3. (cautiously) Add @BRENDA-Enzymes as the GitHub handle for Antje Chang. Unfortunately I have been unsuccessful in getting in contact with them to see if this is their personal handle that they happen to only use for maintenance of BRENDA. This was the only entry that did not have a GitHub handle at this point so at risk of being slightly incorrect, I think it's worth it to enable this check for later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ build/
 /venv
 tmp/
 *.tmp
+.venv
--- a/ontology/bto.md
+++ b/ontology/bto.md
@@ -4,6 +4,7 @@ id: bto
 contact:
   email: a.chang@tu-bs.de
   label: Antje Chang
+  github: BRENDA-Enzymes
 description: A structured controlled vocabulary for the source of an enzyme comprising tissues, cell lines, cell types and cell cultures.
 domain: anatomy
 homepage: http://www.brenda-enzymes.org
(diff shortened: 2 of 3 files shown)


matentzn (approved):
Thanks a ton @cthoyt 

Using Brenda is fine, this is what they were using to work with BTO and afaik they didn't have personal handles. Unfortunately Sandra has left BTO, so I am not sure about its future involvement with OBO.

Lets leave this open until Tuesday night, 1st February, to have people chime in on making `github` handle required. Given that they need to interact with OBOFoundry registry itself using GitHub, I believe it is a no-brainer!

cthoyt:
Given that Locus of Authority (principle 11) states the following:

> A primary contact person MUST be assigned. The name, email address and GitHub username of the contact person MUST be provided when requesting to register with OBO. 

I don't think there will be much dissent.

matentzn:
Oh, didnt realise that it said that! Thx.

PR #1778, opened 2022-01-28
Normalize formatting of registry schema

cthoyt:
This PR doesn't change anything in the JSON file but it does apply a regularized formatting to make it MUCH easier to read


matentzn (approved):
Thanks :)

PR #1799, opened 2022-02-11
Added usages section to pcl.md

shawntanzk:
added usages

### DIFF
--- a/ontology/pcl.md
+++ b/ontology/pcl.md
@@ -36,6 +36,16 @@ dependencies:
   - id: nbo
   - id: chebi
   - id: so
+usages:
+  - user: https://biccn.org/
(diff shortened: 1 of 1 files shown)


shawntanzk:
> This is not an appropriate proof of usage: it should be to a page containing annotations involving a PCL id. A link to OLS is just a sign that the ontology exists.. sorry @shawntanzk

ah no worries, I copied it from PATO which had that. I think that might have to wait for a bit for the app with the annotations on it to be public, just saw the "X" on the dashboard and wanted to plug it :) happy to close this for cleanliness and do this once the app is public

matentzn:
Thank you for your understand.. The PATO usage should really be deleted..

PR #1803, opened 2022-02-18
Update with various guesses

edeutsch:
I feel pretty good about most guesses except build.method. I just left that as is. I don't know what to put there.


matentzn (approved):
I removed the `integration_server` and `build` sections as they are no longer relevant for OBO Foundry, but otherwise this looks great. Thank you!

PR #1808, opened 2022-03-06
Check dependencies' ontology identifiers are valid

### LINKED ISSUE #1769: Implement dependency metadata check
If we want to be all serious about the list of dependencies, I would say we:

- [ ] need to have separate metadata elements for OBO and non-OBO dependencies
- [ ] the non-obo ones should be a valid bioregistry prefix
- [ ] the obo one should be a a valid ID within OBO (no matter if inactive, if the id is in there, it's fine)

Have at it if you want to push this forward!

_Originally posted by @matentzn in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1768#issuecomment-1022266099_

cthoyt:
Closes #1769

This PR does the following:

1. Introduces a test that checks that all of the dependencies for each ontology actually point to OBO ontologies' prefixes. I caught a typo in EPIO as well as a situation in GO where "BridgeOntology" type was used for something that was obviously an internal dependency, so those are excluded from checks for now.
2. Introduces a new location for tests `tests/` where this test gets put
3. Introduces a new environment in the `tox.ini` that runs tests in `tests/` via `pytest`
4. Adds running tests with `tox` into the makefile

Why do steps 2-4? The current setup doesn't rely on using unit testing frameworks for checking data integrity. For new tests, I think it makes sense to finally switch over. This PR makes the compromise that `make` is still used to run these tests. I'm under the impression that @cmungall has committed to supporting modernization of the testing so please don't re-hash some of the old discussions on preferred testing frameworks, development tools, who's committing time to maintaining code and data in OBO Foundry, etc. on this PR.


matentzn (approved):
Looks good to me! Thanks a lot for the fix and test @cthoyt :)

matentzn:
As this PR does not do anything that is not in the spirit of what has already been discussed I motion to merge it asap. I will give it another day: As soon as the sun rises on the horizon of the Aegean Sea on the 12th March, I shall merge this.

cthoyt:
@matentzn friendly reminder to merge

matentzn:
Thank you @cthoyt :)

PR #1820, opened 2022-04-02
Standardize publication references to PubMed

cthoyt:
This pull request standardizes all references to use a PubMed URL in the `publications` and `usages` entries for each ontology. This process resolved a few publisher-specific URLs, found associated PubMed identifiers for DOIs, added https when http was used, and cleaned up various other bits of garbage in URLs.

This PR adds a corresponding unit test to make sure all future publications use this standardized format.

A single exception is made for AGRO, whose reference is to an old BioCreative proceedings which was never assigned a DOI.

Looking forward, it might be a good idea to more carefully annotate these fields using `pubmed` then to have an additional script to bring in full metadata for the publications for display on the OBO site.

### DIFF
--- a/ontology/bspo.md
+++ b/ontology/bspo.md
@@ -24,7 +24,7 @@ build:
   infallible: 1
 tracker: https://github.com/obophenotype/biological-spatial-ontology/issues
 publications:
-  - id: http://www.ncbi.nlm.nih.gov/pubmed/25140222
+  - id: https://www.ncbi.nlm.nih.gov/pubmed/25140222
     title: "Nose to tail, roots to shoots: spatial descriptors for phenotypic diversity in the Biological Spatial Ontology."
 activity_status: active
 repository: https://github.com/obophenotype/biological-spatial-ontology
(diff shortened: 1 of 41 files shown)


cthoyt:
@marieALaporte @celineaubert do you have a more recent publication for AGRO that has a PubMed identifier?

cmungall (approved):

marieALaporte:
Hi @cthoyt No we don't have a more recent publication for AgrO.

cthoyt:
> Hi @cthoyt No we don't have a more recent publication for AgrO.

Okay, thanks for checking. I guess it's fine to leave like this for now.

matentzn (approved):
Man great job again! Thanks so much @cthoyt. I could see how in the future, some people may want to add non-pmid references (arxiv, ceur), but I think it is good practice to try for proper publications here. I am happy to roll with this for now!

cthoyt:
> I could see how in the future, some people may want to add non-pmid references (arxiv, ceur), but I think it is good practice to try for proper publications here. I am happy to roll with this for now!

If/when that happens, I will be happy to make the appropriate improvements to the tests.

PR #1821, opened 2022-04-02
Standardize logo usage

cthoyt:
There were a lot of images floating around in the free text of ontologies' metadata. For example, this causes the Uberon image to be massive and look quite bad (see https://obofoundry.org/ontology/uberon)

This PR standardizes all of those logos using the pre-existing `depicted_by` tag as well as imposes a maximum height constraint on the current format so the larger images don't make the layout look bad. It also does a few minor fixes to the ontology detail.

### DIFF
--- a/_layouts/ontology_detail.html
+++ b/_layouts/ontology_detail.html
@@ -9,10 +9,9 @@ layout: default
     <div class="page-header">
       <h1>
         {% if page.depicted_by %}
-        <img alt="logo" src="{{page.depicted_by}}"/>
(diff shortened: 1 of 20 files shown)


cmungall (approved):

matentzn (approved):
This is great! Much appreciated to bring some orders into the logo situation :) Thank you as always. I didn't check about the removed "tagline" in the HTML page, but I can see that no metadata element has this, so its safe to remove.

PR #1827, opened 2022-04-10
Added the missing reference

egonw:
references https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1826


matentzn (approved):

cthoyt:
@matentzn I would change the text in the box to "references #1826" so it doesn't accidentally close the issue

PR #1841, opened 2022-04-19
Require new ontologies include a link to contribution guidelines

### LINKED ISSUE #1836: Require new ontologies' repositories to include documentation on how to contribute
I was just reading through the repository linked for PROCO in the new ontology request https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1834, and it seems that the ontology has very little information on how the ontology is updated/maintained. Considering that many OBO Foundry ontologies suffer from issues related to longevity and sustainability, not having a document explaining how external contributors can make their own edits and send pull requests seems to be very counterproductive.

GitHub itself has a first-class notion of repositories including a contributions file. For example, in any repository, you can look at the "community page" e.g., https://github.com/proco-ontology/PROCO/community which links to [this guide](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) for writing contribution guidelines.

Screen Shot 2022-04-18 at 10 36 52

Some ways to make this actionable:
1. Add a new field to the metadata linking to the a contributing file (which should be in the root of the repository and can be a plain text file, a markdown file, or a restructured text file)
2. Update the new ontology request form to include a link to this file
3. Figure out how to check new ontology metadata files have this (but disregard old ones). While it's difficult to retroactively apply new rules, it's much easier now than it used to be to check repository properties en masse, as demonstrated in the [OBO Community Health Report](https://cthoyt.com/obo-community-health/)

cthoyt:
Closes #1836

Given the [OBO Foundry Principle 10 (Commitment to Collaboration)](https://obofoundry.org/principles/fp-010-collaboration.html), ontologies should actively lower the bar and encourage participation from the community. For different ontologies, this ranges from participation in community discussion on slack (e.g., UBERON has their own slack workspace), attendance in project-specific meetings (e.g., RO does this), making issues (all ontologies are **REQUIRED** to have a public issue tracker), to reviewing PRs (IMO all ontologies should be required to consider external contributions, but some reject them without consideration). 

Having a CONTRIBUTING.md in an ontology's repository that is further in [a place that GitHub can recognize it](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors) is crucial for enabling community participation since it outlines the expectations for the contributors and the maintainers.

As of https://github.com/INCATools/ontology-development-kit/pull/620, the ODK now has a basic contribution guidelines file built-in, so all of the original blockers towards the issue #1836 that proposed we require contribution guidelines for all new ontologies are now gone.

This PR specifically does the following:

1. Adds a field to the new ontology form to make explicit the URL to the contribution guidelines in the repository
2. Implements (as a follow-up to #2277) a check that is *only* applied to new ontologies that going forwards they will be required to have contribution guidelines before being accepted into OBO Foundry


matentzn:
@cthoyt you say:

> Would be nice to finish https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1841 before admitting new ontologies

I think its good to get the contributing.md requirement in for new ontologies. I will make sure this happens. But before we do, we need a really good blueprint - I will work in the next weeks (ODK), but it will be slow. After the blueprint is done, we will share it as an example.

If we require this contributing.md on ontologies that are already admitted, this will be a never ending process. I think this should be firmly in the "SHOULD" space, and I am also happy to add a warning to the OBO Dashboard in the "collaboration" principle. But an ERROR it wont be, at least not if you ever want this merged.

cthoyt:
As a way to get around the issue of applying new, higher standards to old ontologies (many of which are still called "active" but will not be able/willing to make improvements), I've proposed we start tracking date of admission of ontologies in #1967 such that all new SOPs can be tagged with the date when they go in effect, and only ontologies submitted after that MUST and before that SHOULD follow them.

matentzn:
Ahhhh now I get it! Ok, all in then! I like that!

cthoyt:
After #1969, it will be more simple to say that all ontologies added after some YYYY-MM-DD have to conform to new rules and review of this PR will be easier.

cthoyt:
This issue is now unblocked:

- https://github.com/INCATools/ontology-development-kit/pull/620 added base contribution guidelines to ODK that can be reused in new ontologies
- https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2277 made it possible to add new metadata checks that only apply to new ontologies (it was a valid concern in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1836 that it would not be possible to retroactively get existing OBO Foundry ontologies to make contribution guidelines). Therefore previous references to #1969 have been removed as they would only be a distraction

@matentzn @nlharris  you know my style, I like to bring solutions to OBO Foundry community then ask people to discuss rather than having slow moving discussions that never go anywhere first. Can you help re-orient the solution in this PR and the original discussion in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1836? What kind of discussion do we need to have to start enforcing new ontologies (but again, not existing ones) in OBO Foundry conform to a higher standard?

nlharris:
> What kind of discussion do we need to have to start enforcing new ontologies (but again, not existing ones) in OBO Foundry conform to a higher standard?

See https://github.com/OBOFoundry/OBOFoundry.github.io/discussions/2209 (Should we make passing the dashboard mandatory by January 2025?) (to which the agreement was "yes").
(BTW, I dislike Discussions because it's a separate place that we (including me) mostly forget to check. I think it's fine to use some Issues as discussions.)

cthoyt:
> (BTW, I dislike Discussions because it's a separate place that we (including me) mostly forget to check. I think it's fine to use some Issues as discussions.)

@nlharris It seems that going forward, ontologies should pass OBO Dashboard. However, we should be able to make other quality tests that are not implemented there. Unfortunately, it's the case that OBO Dashboard is not so accessible for contribution, so I'm not going to put any of my effort there.

More importantly, we should do this in a way that we can add new checks without imposing them on already-accepted ontologies. There's way too much grumbling when we start talking about community-wide improvements because people are lacking time/funding/motivation.

nlharris:
Hey, I'm just the messenger.

And I'm not sure why you quoted my grumble about Discussions (by which I specifically meant GitHub Discussions, not lowercase-d discussions!).

cthoyt:
@nlharris oop no somehow I quoted wrong! I wasn't talking about your grumbling!

just let me know if you have any ideas where to bring that up :)

.github/ISSUE_TEMPLATE/new-ontology.yml
```diff
@@ -102,6 +102,14 @@ body:
       placeholder: e.g. https://github.com/obophenotype/uberon/issues
     validations:
       required: true
+  - type: input
+    id: contributing
+    attributes:
+      label: Contribution Guidelines
+      description: What is the URL of contribution guidelines? Ideally, this is a file name CONTRIBUTING or CONTRIBUTING.md in the root of your repository.
```
erik-whiting, line 109: this is a super tiny nitpick so feel free to tell me to kick rocks but can we also add `CONTRIBUTING.rst` as an option since this seems to be gaining popularity with Python people?

tests/test_integrity.py
```diff
@@ -334,3 +334,24 @@ def test_repository_license(self):
...
+            repository = data["repository"]
+            if not repository.startswith("https://github.com"):
+                continue
+            r = repository.removeprefix("https://github.com/").rstrip("/")
+            github_data = self._get_github_data(prefix)
+            default_branch = github_data["default_branch"]
+            paths = [
+                f"https://github.com/{r}/blob/{default_branch}/CONTRIBUTING.md",
+                f"https://github.com/{r}/blob/{default_branch}/docs/CONTRIBUTING.md",
+                f"https://github.com/{r}/blob/{default_branch}/.github/CONTRIBUTING.md",
```
erik-whiting, line 350: This will reject any repository with just a `CONTRIBUTING` or `CONTRIBUTING .rst` correct? is that we want, to enforce `CONTRIBUTING.md` only?

erik-whiting:
I have two questions about these requests to the GitHub API. GitHub rate-limits requests to the API to 60 an hour if you're not using a token, see this section of an API article: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limits-for-requests-from-github-actions
My questions:
1. Are we nearing that threshold? If so, someone with admin access to the repo needs to put a GitHub token in the CI secrets so we can reference it in the test file. If we're nowhere near 60 then never mind me. But yeah, if we hit that threshold in the middle of an ontology validation, it's possible someone could get their contribution wrongly rejected. We would have to wait for the API counter to reset to try again.
2. If we are nearing that threshold, has someone already addressed this issue? I don't have access to the repo secrets so idk. Again, this might not even be an issue, in which case don't mind me.

.github/ISSUE_TEMPLATE/new-ontology.yml
cthoyt: actually, RST is on the downturn in the Python community, but I will add it anyway!

tests/test_integrity.py
cthoyt: If there's no formatting then i am insulted that they care so little for potential contributors

cthoyt:
@erik-whiting since this will only run API calls on branches that have a new ontology, it will at most make 1 API call. I'd consider this negligible, and no follow-up action is required.

.github/ISSUE_TEMPLATE/new-ontology.yml
cthoyt: Done in 9689d084

erik-whiting (approved):
lgtm :shipit:

PR #1844, opened 2022-04-19
Update omrse.md

aellenhicks:
Added publication

### DIFF
--- a/ontology/omrse.md
+++ b/ontology/omrse.md
@@ -27,6 +27,10 @@ build:
 activity_status: active
 repository: https://github.com/ufbmi/OMRSE
 preferredPrefix: OMRSE
+publications:
+  - id: https://jbiomedsem.biomedcentral.com/articles/10.1186/s13326-016-0087-8
+    title: "CThe ontology of medically related social entities: recent developments"
+
 ---
 
 This ontology covers the domain of social entities that are related to health care, such as demographic information (social entities for recording gender (but not sex) and marital status, for example) and the roles of various individuals and organizations (patient, hospital,etc.)
--- a/ontology/omrse.md
+++ b/ontology/omrse.md
@@ -28,7 +28,7 @@ activity_status: active
 repository: https://github.com/ufbmi/OMRSE
 preferredPrefix: OMRSE
 publications:
-  - id: https://jbiomedsem.biomedcentral.com/articles/10.1186/s13326-016-0087-8
+  - id: https://www.ncbi.nlm.nih.gov/pubmed/27406187
     title: "CThe ontology of medically related social entities: recent developments"
 
 ---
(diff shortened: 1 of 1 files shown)


ontology/omrse.md
```diff
@@ -27,6 +27,10 @@ build:
 activity_status: active
 repository: https://github.com/ufbmi/OMRSE
 preferredPrefix: OMRSE
+publications:
+  - id: https://jbiomedsem.biomedcentral.com/articles/10.1186/s13326-016-0087-8
+    title: "CThe ontology of medically related social entities: recent developments"
```
hoganwr, line 32: There appears to be a stray upper-case C at the beginning of the title

ontology/omrse.md
```diff
@@ -27,6 +27,10 @@ build:
...
+
```
hoganwr, line 33: Please also add (I'm in reviewing mode, so I cannot do this myself.  Or if it's more appropriate, I can make edits later)

  - id: http://ceur-ws.org/Vol-833/paper20.pdf
    title: "Representing the reality underlying demographic data"

hoganwr (changes requested):

ontology/omrse.md
```diff
@@ -27,6 +27,10 @@ build:
 activity_status: active
 repository: https://github.com/ufbmi/OMRSE
 preferredPrefix: OMRSE
+publications:
+  - id: https://jbiomedsem.biomedcentral.com/articles/10.1186/s13326-016-0087-8
```
cthoyt, line 31: Please use the PubMed identifier since a URL to a journal isn't really actionable
```suggestion
  - id: https://www.ncbi.nlm.nih.gov/pubmed/27406187
```

ontology/omrse.md
cthoyt: @hoganwr there are currently some debates about whether to / how to add links to PDFs that appear in conferences that don't assign stable identifiers that you can get actionable metadata from in #1846 and #1855 so I would suggest holding off on that. Alternatively, you can put the link to the paper in the free text part of this markdown file.

matentzn: We could consider ceur-ws stable? Its been stable for 15 years at least.. Seems the ceur servers contain a lot of citable content

cthoyt: I said the service doesn't assign stable, actionable identifiers, not that the service itself isn't stable.

matentzn: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1857

matentzn (approved):

matentzn:
@hoganwr I moved your request to a new issue

https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1857

hoganwr (approved):

PR #1845, opened 2022-04-19
Update doid.md

aellenhicks:
Added an additional publication

### DIFF
--- a/ontology/doid.md
+++ b/ontology/doid.md
@@ -38,6 +38,8 @@ build:
   infallible: 1
 tracker: https://github.com/DiseaseOntology/HumanDiseaseOntology/issues
 publications:
+  - id: https://pubmed.ncbi.nlm.nih.gov/28253937/
+    title: "Therapeutic indications and other use-case-driven updates in the drug ontology: anti-malarials, anti-hypertensives, opioid analgesics, and a large term request"
   - id: https://www.ncbi.nlm.nih.gov/pubmed/25348409
     title: "Disease Ontology 2015 update: an expanded and updated database of human diseases for linking biomedical knowledge through disease data"
 license:
--- a/ontology/doid.md
+++ b/ontology/doid.md
@@ -38,7 +38,7 @@ build:
   infallible: 1
 tracker: https://github.com/DiseaseOntology/HumanDiseaseOntology/issues
 publications:
-  - id: https://pubmed.ncbi.nlm.nih.gov/28253937/
+  - id: https://www.ncbi.nlm.nih.gov/pubmed/28253937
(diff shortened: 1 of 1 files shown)


matentzn (approved):

matentzn (approved):

lschriml:
@aellenhicks and @matentzn -- I have updated the DO citation. As a policy, I do not think it appropriate for members of the OBO Foundry who are not the projects editor or PI to update the publications that should be cited for an ontology project.
Instead, a ticket should be opened with the suggested citation. 
Cheers,
Lynn

cthoyt:
@lschriml disagree - such a policy would discourage community members from helping maintain the OBO Foundry. What happens if the ontology's contact person is unresponsive (as a very large number of them have demonstrated to be)? As the GitHub handle of _all_ ontologies' contact people are available, I think a better thing to do would be to flag updates to any given ontology with the contact person so they have the chance to get involved, otherwise our metadata overlords like @matentzn can make an executive decision to merge without them if they're unresponsive for too long, or if the change is not controversial.

matentzn:
Oh, I thought that @aellenhicks was part of your team @lschriml - never thought anyone would ever bother to update the contents of a record otherwise! I think it makes sense to tag the contacts and give them 72 hours to take a look - obviously we will forget that occasionally, but we can try for it. And, if after 72 hours no response is there, we just merge. Sounds like a compromise?

lschriml:
@matentzn --  yes, a good compromise. 
     Agree it is great to keep the ontology files updated.
And to coordinate with the project editor/PI.

Cheers,
Lynn

lschriml:
@cthoyt @matentzn 
   Agreed, we should actively encourage community members in helping maintain the OBO Foundry. 

I think we can be a bit cautious here in how we deliver that help, to offer advice and solutions, at the same 
time to respect the developers of a project and their choices on what to include in their ontology. 

   I am concerned that if we change things in others ontologies, without their knowing about it, that this may discourage people joining the OBO Foundry, or being concerned that they are not involved. Or that developers may feel the OBO Foundry is overstepping. 

  From my own experience, as a PI of a two decades old project, I have responsibilities to my users and to my funders for what is included in the ontology and how it is presented, thus I take in all suggestions, but do review them before integration. 

Perhaps we can come together and formulate guidelines, to be shared with the OBO Foundry members, 
   such as: 
     A suggestion to be followed by a time period for PIs/editors to respond to the suggestion. 
           And clarifying what types of updates the OBO Foundry would involve. 
   As a measured approach, that continues to involve the PI/editor in the OBO Foundry efforts.  


Cheers,
Lynn

lschriml:
@matentzn  -- can do.

PR #1846, opened 2022-04-20
Permit any kind of publication in publication field

matentzn:
According to our OBO operations call on the 19th of April, we (the OBO operations committee) mostly unanimously decided to open the publication field to mean: "This is what we want you to cite when you cite this ontology" as opposed to "this is the peer-reviewed publication we have produced describing the ontology".

### DIFF
--- a/tests/test_integrity.py
+++ b/tests/test_integrity.py
@@ -89,5 +89,3 @@ class TestIntegrity(unittest.TestCase):
         identifier = publication["id"]
         self.assertIsInstance(identifier, str)
         self.assertFalse(identifier.endswith("/"))
-        self.assertTrue(identifier.startswith(uri_prefix), msg=msg)
-        self.assertTrue(identifier[len(uri_prefix):].isnumeric())


cthoyt:
You can already see from #1845 that without any structured checks, this metadata will immediately fall into disarray (again). Here's what I suggest as an alternative to allowing this to go back to open season: anyone who wants to propose a new publication can do it via a pull request. In the few situations like https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1831 where there isn't a real publication (i.e., indexed in a highly regarded resource like PubMed) that accompanies, but only something like a PDF that they want to reference, we can start a discussion where we ask them for a better reference, and if they really can't provide one, we can add their link to an "exception" list. I volunteer to take care of all such requests and maintain the associated code.

If and when the day comes that someone wants to reference a paper in another kinds of resource, like a pre-print server (e.g., arXiv, bioRxiv, medRxiv), we can discuss changing this further.

matentzn:
You would be taking quite a lot on @cthoyt. What if someone dropped a PDF report on Zenodo and wants people to cite that? I understand your sentiment, but its a bit too strict, unless you cover the service of:

X makes pull request, and without debate, you search for the most appropriate indexing site that has that paper, then add a link then extend the QC. I didn't have the impression that anyone was happy for a debate on the matter..

The people on the call wanted to be able to say: cite the pre-print/report on Zenodo etc.

Lets start with #1831 - how would you solve this?

cthoyt:
> You would be taking quite a lot on @cthoyt. What if someone dropped a PDF report on Zenodo and wants people to cite that? I understand your sentiment, but its a bit too strict, unless you cover the service of:
> 
> X makes pull request, and without debate, you search for the most appropriate indexing site that has that paper, then add a link then extend the QC. I didn't have the impression that anyone was happy for a debate on the matter..

Yes, I am mostly saying I will do this. However, I will give debate any time there is a non-useful link, such as to PDFs for non-archival conferences, which aren't actionable. I would recommend these people drop the PDF on Zenodo or arXiv so at least it's assigned an actionable, stable identifier that can be used to get its metadata.

In terms of how much taking on this is, I think it's better to have a more strict check before the metadata goes in rather than having to fix things (i.e., make metadata useful/actionable) _post facto_, which might now even require me to get in touch with the ontology owner _again_ to make updates.

> I didn't have the impression that anyone was happy for a debate on the matter.. 

Not sure what you mean here.

> Lets start with https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1831 - how would you solve this?

I've proposed a concrete update in #1855 to loosen this field to certain kinds of identifiers from pre-print services. This still does not address the case of #1831 since the paper in question there is still not in a place where there is any useful actionable information about it. Alternatively, why can't we ask people who don't have structure citations just to put them in the free-form metadata for the field? Or Alternatively, can we make a list of URLs to "see also" that can have any URL at all inside them, so people can put the less actionable kinds of references (which might also point to twitter threads, github discussions, etc.)

matentzn:
I think this makes sense. Action summary:

- `publication` field means: consider these when citing the ontology. It does not mean: "here is my shiny peer reviewed pub"
- To properly display citable information (like generating citation records etc), we need _some_ kind of structure. A PDF to somewhere on the web is not sufficient. I think this makes sense.
- @cstoeckert would you be willing to upload your paper to a pre-printserver like arxiv, or even Zenodo (which has zero content control) so we can generate a proper citation link?

Thank you for your time on this all!

matentzn:
Closing this in favour of #1855

matentzn:
@cthoyt, am I correct in assuming that semanticscholar does not  neither permanent nor actionable identifiers? #1859

cthoyt:
So this is kind of not correct, it appears that `888ea51dbb138122d39c42f863f2a1ada7976db4` is an identifier for this paper and you can resolve with `https://www.semanticscholar.org/paper/$1` as in https://www.semanticscholar.org/paper/888ea51dbb138122d39c42f863f2a1ada7976db4 (it just redirects to a URL with garbage text in it). I'm not sure how stable this is, especially since it appears to be a hex digest, probably from some kind of hash.

There even appears to be an API for getting the title of a paper described [here](https://www.semanticscholar.org/product/api) such as https://api.semanticscholar.org/graph/v1/paper/888ea51dbb138122d39c42f863f2a1ada7976db4, but it seems pretty limited from a first glance (i.e., it's not documented how to get any other information from this API, like where something was published, when, and who were the authors).

That all being said, I think there's still a big issue with trust - semantic scholar is a third party index. I think it's a pretty reasonable bar to say that we want first-party providers here as well. I'll have to think about a better way to articulate this

cstoeckert:
How about instead of asking for an "id" where a URL can be entered and a pubmed ID is currently expected, we ask for a DOI? Free text could still be provided which can describe how to cite the ontology but only a DOI would show up as a link if there was one. In the case of #1831 there isn't a DOI so no link but there could still be a description: Zheng, J., Cade, J. S., Brunk, B., Roos, D. S., Stoeckert, C. J., et al (2016). Malaria study data integration and information retrieval based on OBO Foundry ontologies. CEUR Workshop Proceedings, 1747.

PR #1855, opened 2022-04-21
Loosen publication validation to allow Zenodo, arXiv, bioRxiv, and medRxiv

cthoyt:
This doesn't allow chemRxiv - unfortunately the fact that the preprint service is run by a big publishing firm was incentive enough for them to bork the "stable" chemRxiv identifiers. Now the only good way for this kind is to use DOI.


matentzn:
Ready for review? Still in draft mode.

cthoyt:
@matentzn yes, I was just testing locally in the mean time but it's probably good enough

matentzn (approved):
I think your proposed compromise is reasonable. Very thankful you keep battling for open science despite resistance! :)

Issue #1848, opened 2022-04-20
Develop SOP for Changing Ontology Metadata


lschriml:
As discussed in #1845 - thank you @matentzn for the feedback. 

I am making a suggestion here, that we develop a SOP for adding/updating metadata for an ontology. 
           Specifically, for updates made by the OBO Foundry and/or individuals who are not part of the a project or the PI.

Issues that arose:
 - what metadata in an ontology.md file is the purview of the ontology editor/PI to edit;
-  which fields/data types can/should the OBO Foundry or non-project people edit. 

I think it is fabulous that we, the OBO Foundry, are helping ontology projects to improve their ontologies, be that via Dashboard reports, or analysis of data representations in their ontology.md file. Moving forward, I think we could use some parameters/guidelines on how we deliver that help, to offer advice and solutions. To more clearly establish boundaries/guidelines for these updates. 
 
Some ideas on moving forward: 
- formulate (and communicate to ontology PIs) a plan to: 
           -  identified problems and intended updates 
           - clarifying what types of updates the OBO Foundry might do
           - establish a protocol to communicate with PIs 
                     - emails in addition to GitHub mentions, as these may be missed
                     - most of us use GitHub daily, but we should consider that not all PIs/editors are checking their GitHub daily 
           - with a set time period for PIs/editors to respond to the suggestion (at least a week)
           - add some specifically protected fields that only the project PI/editor would update
           - if an individual has suggestions for a project, it is to be conveyed via a Github ticket, the ontology.md should not be 
              updated otherwise
Cheers,
Lynn

matentzn:
@lschriml would you like to prepare a list of metadata items that you think 
- _must_ be signed off by the contact/pi (id, name, etc) (no merge without permission)
- _should_ be signed off by the contact/pi (publication, etc) (request for permission but merge after 7 days period even without permission) 
- _could_ be signed off by contact/pi (changes to metadata structure that don't affect meaning, typos) (no request for review, merge after 72 hours)

I am against the proposal of the email message, but could be convinced otherwise. For SHOULD and COULD, but could be convinced for an email in the MUST section.

cthoyt:
If there are any fields that _must_ be signed off by the contact, then I think it would be good to in tandem outline what is the time frame the contact person has for this kind of request and what the consequences are for not engaging (e.g., marking the ontology as orphaned)

matentzn:
[...]

PR #1866, opened 2022-04-24
Fix MRO publication

beckyjackson:
As requested by @rvita - the publication for MRO is incorrect, it should be [An ontology for major histocompatibility restriction](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4709943/).

### DIFF
--- a/ontology/mro.md
+++ b/ontology/mro.md
@@ -26,8 +26,8 @@ usages:
       - url: https://www.iedb.org/assay/1357035
         description: "Epitope ID: 59611 based on reference 1003499"
     publications:
-      - id: https://www.ncbi.nlm.nih.gov/pubmed/30357391
-        title: "The Immune Epitope Database (IEDB): 2018 update"
+      - id: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4709943/
+        title: "An ontology for major histocompatibility restriction"
 activity_status: active
 repository: https://github.com/IEDB/MRO
 preferredPrefix: MRO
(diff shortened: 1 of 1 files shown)


beckyjackson:
Can we allow `https://pubmed.ncbi.nlm.nih.gov/` for PMIDs? It's strange to me that this fails.

cthoyt:
Technically, yes, but this would sort of defeat the point of standardizing these URLs. I think a better step forward would be to open this dictionary up to allow just putting in the pmid key itself, then downstram consumption could figure out how to format that into a URI.

ontology/mro.md
```diff
@@ -26,8 +26,8 @@ usages:
       - url: https://www.iedb.org/assay/1357035
         description: "Epitope ID: 59611 based on reference 1003499"
     publications:
-      - id: https://www.ncbi.nlm.nih.gov/pubmed/30357391
-        title: "The Immune Epitope Database (IEDB): 2018 update"
+      - id: https://pubmed.ncbi.nlm.nih.gov/26759709
```
cthoyt, line 29:
```suggestion
      - id: https://www.ncbi.nlm.nih.gov/pubmed/26759709
 ```

beckyjackson:
+1 to putting the PMID in instead

matentzn (approved):

matentzn:
@bpeters42 is listed as contact, given you are all part of the same time I assume this does not require a specific review by the contact. See also https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1848 for making this more transparent.

PR #1868, opened 2022-04-27
Update contacts

cooperl09:
Updating contact email and adding second contact


cthoyt:
@cooperl09 the obo principles stipulate that there is only a single responsible contact person for each ontology

PR #1885, opened 2022-05-05
Added SWO publication as per #1826

allysonlister:
Please check and approve if I have done the addition of the SWO publication correctly, thank you!


cthoyt:
@allysonlister thank you! the only issue is the spacing is incorrect (should be 2 instead of 3) and there are trailing spaces as noted by the failing CI workflow in https://github.com/OBOFoundry/OBOFoundry.github.io/runs/6302952073?check_suite_focus=true#step%3A5%3A11=

allysonlister:
Thanks for that @cthoyt ! (I also appreciate the ping that I found in my inbox this morning with regards to this).

I think I've solved the spacing issues now :)

cthoyt (approved):

matentzn (approved):

PR #1890, opened 2022-05-05
Update labo.md

pfabry:
Publication added

### DIFF
--- a/ontology/labo.md
+++ b/ontology/labo.md
@@ -30,6 +30,9 @@ dependencies:
   - id: opmi
 products:
   - id: labo.owl
+publications: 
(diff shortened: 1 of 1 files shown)


pfabry:
Hi Charles,

Unfortunately, it seems I can't get an actionable link from ICBO proceedings. The PDF linked is an ICBO article from 2019. Could a semantic scholar link do the trick ?

pfabry:
Thanks. 
On a side note, when I update the .md file I systematically get a fail. Is it normal behavior or is there something I'm doing wrong ? I have a third ontology to update and I want to be sure I'm doing the right thing.

cthoyt:
Yes - you should be careful to avoid trailing whitespace and use the proper indentation.

matentzn (approved):

PR #1910, opened 2022-05-08
Annotate several ontologies' publications

cthoyt:
These were all based on suggestions from https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1826 - each commit is annotated with a link to the comment.

### DIFF
--- a/ontology/fbcv.md
+++ b/ontology/fbcv.md
@@ -35,6 +35,7 @@ usages:
         description: "alleles and constructs annotated to bang sensitive in FlyBase"
 activity_status: active
 repository: https://github.com/FlyBase/flybase-controlled-vocabulary
+publications: []
 ---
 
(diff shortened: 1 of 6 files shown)


ontology/fbcv.md
```diff
@@ -35,6 +35,7 @@ usages:
         description: "alleles and constructs annotated to bang sensitive in FlyBase"
 activity_status: active
 repository: https://github.com/FlyBase/flybase-controlled-vocabulary
+publications: []
```
matentzn, line 38: What is the point of documenting this? Just to say that there is no primary publication? Why is this useful?

cthoyt: Having an empty list versus a missing value allows us to check if it’s been annotated explicitly that there’s no publication (yet)

matentzn (approved):

PR #1922, opened 2022-05-19
Updating CL obo foundry page

shawntanzk:
Updating CL OBO foundry page in preparation for NAR update paper

Will not merge until @dosumis goes through

### DIFF
--- a/ontology/cl.md
+++ b/ontology/cl.md
@@ -73,19 +73,19 @@ usages:
     type: annotation
     description: HuBMAP develops tools to create an open, global atlas of the human body at the cellular level. The Cell Ontology is used in annotating cell types in the tools developed.
     publications:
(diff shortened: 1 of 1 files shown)


shawntanzk:
Fixes https://github.com/obophenotype/cell-ontology/issues/1591

matentzn (changes requested):
As per convention, we can only accept usages if they point to a page that reveal a concrete usage of a CL term.. else usages are only documented in the description section. Can you provide concrete links with examples? See Mondo md file.

shawntanzk:
whoops sorry about that, yeah there should be specific examples, will add them :)

matentzn (approved):
Thank you @shawntanzk 

Please fix errors and we are good to go.

PR #1928, opened 2022-05-23
Centralize info on participation

nlharris:
Currently distributed among menus or linked from other pages.

Not sure whether < li > will work in the web page, and I can't check that in the preview, so I'll have to wait for someone to merge this PR and then fix the formatting if needed.

### DIFF
--- a/docs/participate.md
+++ b/docs/participate.md
@@ -1,0 +1,16 @@
+---
+layout: doc
+id: Participate
+title: How to participate in the OBO Foundry
+---
+
+The OBO Foundry is currently a volunteer-run organization. There are many ways to get involved:  
+ <ul>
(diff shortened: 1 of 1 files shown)


docs/participate.md
```diff
@@ -0,0 +1,16 @@
+---
+layout: doc
+id: Participate
+title: How to participate in the OBO Foundry
+---
+
+The OBO Foundry is currently a volunteer-run organization. There are many ways to get involved:  
+ <ul>
```
cmungall, line 8: in general, markdown is preferred over HTML as more maintainable, but we can do as a separate PR

cmungall (approved):

docs/participate.md
nlharris: ok, i was being lazy; i copied it from a .html page and was hoping not to have to hand-edit it into md. but i can if that is preferred!

nlharris: @cmungall i converted it to .md

cmungall (approved):

matentzn (approved):

PR #1931, opened 2022-05-24
Showing edit buttons only if page has an id

### LINKED ISSUE #1925: Edit button not working on publications.html
The edit button on https://obofoundry.org/registry/publications.html goes to
https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/docs/.md which is wrong

I'm not sure where/how these edit button URLs are generated.

matentzn:
fixes #1925

### DIFF
--- a/_layouts/check.html
+++ b/_layouts/check.html
@@ -28,7 +28,7 @@ layout: default
         </li>
         {% endfor %}
         </ul>
-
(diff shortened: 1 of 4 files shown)


nlharris (approved):
seems reasonable...

PR #1933, opened 2022-05-24
Update fp-004-versioning.md

nataled:
I suspect this will need further refinement, but here's attempt 1.

### DIFF
--- a/principles/fp-004-versioning.md
+++ b/principles/fp-004-versioning.md
@@ -17,17 +17,17 @@ OBO projects share their ontologies using files in OWL or OBO format (see [Princ
 Recommendations and Requirements
 -------
 
-In addition to an IRI specifying the current release (see [Principle 3](https://obofoundry.org/principles/fp-003-uris.html)), each official release MUST have a unique [version IRI](https://www.w3.org/TR/owl2-syntax/#Ontology_IRI_and_Version_IRI) that resolves to the specific ontology artifact indicated. Consumers can then use the version IRI to uniquely identify which official release of the ontology they used, and to retrieve unaltered copies of the file(s). The versionIRI is constructed using (1) an ontology _base prefix_ (2) a version _identifier_ and (3) a path to a particular version _product_ or artifact.
+In addition to an IRI specifying the current release (see [Principle 3](https://obofoundry.org/principles/fp-003-uris.html)), each official release MUST have a unique version IRI that resolves to the specific ontology artifact indicated. Consumers can then use the version IRI to uniquely identify which official release of the ontology they used, and to retrieve unaltered copies of the file(s). A _version IRI_ is a full path that MUST resolve to the particular version of the ontology artifact. Both the version IRI and the corresponding artifact MUST contain an identical _version identifier_ string.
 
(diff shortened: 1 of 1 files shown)


cthoyt:
For an assessment of the current situation related to how well ontologies are doing this, see https://github.com/cthoyt/oquat/tree/main/docs/versions#obo-analysis

matentzn (approved):
This is excellent. Love it.

- I approve, but you could optionally add a link in both cases you mention semantic versioning to https://semver.org/.
- Also I note from a previous edit that the quotes you use have the wrong encoding, i.e. `“YYYY-MM-DD“` instead of `"YYYY-MM-DD"`.

matentzn:
@cthoyt I am wondering if your tools (obo health, version checker etc) should be listed here:

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/OntologyTools.md

Do you think its appropriate/fits the intention of that page etc?

nlharris:
Note also this page, which overlaps with the one Nico mentioned above: https://obofoundry.org/resources (https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/resources.md)

Should that be merged with https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/OntologyTools.md ?

matentzn:
Thanks @nlharris forgot about the resources page, maybe @cthoyt to add to resources page instead!

If someone wants to make a PR merging the two be my guest, I think its ok either way. I am happy keeping them separate!

cthoyt:
@matentzn where would you list the OBO Dashboard? I think these would be good to list in similar places

matentzn:
resources

nataled:
Quick notes/questions:
1) @matentzn The OBO Dashboard is not currently listed in either resources or ontology tools; it is listed under the 'ontologies' dropdown (which I always found a bit odd). Wasn't sure if you thought it was already under resources, or if you'd like it put there.
2) It appears that all the 'Edit' buttons are now gone from all pages (they were visible as of yesterday).
3) @cthoyt I notice that at least one ontology (PR) does not show up on your versioning check list. Any idea what would cause that?

matentzn:
1: @cthoyt has done it now. Will show up soon.
2: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1936

nataled:
@matentzn I'm fixing the quotes issue you mentioned. I'm declining the link to semver because that description seems tailored to programs rather than documents, and it isn't clear how to map the suggestions.

nlharris:
> It appears that all the 'Edit' buttons are now gone from all pages (they were visible as of yesterday).

Nico's fix should have removed the non-functional Edit buttons and left those that worked. Maybe none of them were functional?

matentzn:
There was a bug, its fixed now: https://obofoundry.org/principles/fp-001-open.html

nataled:
Those non-functional Edit buttons were not caused by any programming bug so far as I could tell. At least in the cases I saw, they were caused by an incorrect file pointer in the 'id' line (for example, instead of 'id: fp-004-versioning' that line might have been 'id: fp-004'). I fixed all those for the principles some months ago, but I did not look at all pages.

matentzn:
Yeah, @nlharris observed that bug as well - I simply now do not show the edit buttons where the `id` line does not exist.

PR #1934, opened 2022-05-24
Clarify publication page: How to cite OBO Foundry

nlharris:
apparently this script overwrites https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/registry/publications.md (which i manually updated yesterday). (should we add a comment to that effect to publications.md so future-me or others don't waste time editing that again?)


matentzn:
@nlharris I added a markdown comment to clarify that the file should not be edited and where it is managed.

matentzn (approved):

Issue #1939, opened 2022-05-25
Check pages for missing 'Edit' button


nataled:
Properly functioning Edit buttons require an id line in the .md file, like these:

layout: ontology_detail
id: pr

or 

layout: principle
id: fp-003-uris

Some pages lack this line, and thus are not easily editable. I've gone through all the principles pages and verified that these are as they should be, but we should review all Foundry pages to make sure the pages we want editable are properly formatted.

nlharris:
See also: #1925 #1931

nlharris:
The edit button on https://obofoundry.org/resources points to https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/docs/resources.md which is wrong -- it should point to https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/resources.md

The id in resources.md is simply "resources". What should it say to make the edit button go to the right place?

nataled:
Unless I'm mistaken, there's naught to do except move the file to where it is expected.

nlharris:
I don't want to do that; then everything else that points to that page will be pointing to the wrong place.

Instead, I think the "fix" here is to take out the id so that it doesn't have an Edit button. Is that ok?

nlharris:
BTW, re
> Some pages lack this line, and thus are not easily editable.

I think by far the majority of page edits are done by insiders who know where the pages can be found in GH. Those who can't find them can always open a ticket.

nataled:
@nlharris what you say is true, but trying to find the pages isn't obvious (as you now know). It's a barrier that doesn't need to be there. IMO, if a page is one that isn't intended to be completely static, it should have an Edit button.

PR #1943, opened 2022-05-26
remove "id" line to avoid having Edit button that goes to the wrong URL

nlharris:
to fix this issue I noted in #1939:
> The edit button on https://obofoundry.org/resources points to https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/docs/resources.md which is wrong -- it should point to https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/resources.md

Darren noted:
> Unless I'm mistaken, there's naught to do except move the file to where it is expected.

I responded:
> I don't want to do that; then everything else that points to that page will be pointing to the wrong place. Instead, I think the "fix" here is to take out the id so that it doesn't have an Edit button. Is that ok?

### DIFF
--- a/resources.md
+++ b/resources.md
@@ -1,6 +1,5 @@
 ---
 layout: doc
-id: resources
 title: Ontology Tools and Resources
 ---
 ## Ontology Browsers


matentzn (approved):
Seems a bit annoying that we cant edit it directly but makes sense as a quick fix.

Issue #1942, opened 2022-05-26
Mark ERO and OGI as obsolete?


matentzn:
https://obofoundry.org/ontology/ero.html
https://obofoundry.org/ontology/ogi.html

http://purl.obolibrary.org/obo/ero.owl
http://purl.obolibrary.org/obo/ogi.owl

PURL does not work at all anymore.

Can someone contact owners?

rvita:
I contacted Marc Ciriello and Juliane Schneider about this 5/26/22

matentzn:
Thank you @rvita

matentzn:
Thank you @rvita, see #1944. Let me know if you learn anything regarding OGI!

rvita:
I was never able to find a contact person for OGI. Does anyone know Lin Yu?

cthoyt:
CC @linikujp they're trying to get in touch

cthoyt:
@rvita I also added all relevant contact information in https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1953 if you want to reach out on non-public channels

linikujp:
Thank you @cthoyt and @rvita , I left the comment and questions at #1953 . Please let me know.

nlharris:
What was decided about this?

matentzn:
It's finished! OGI incorporated into OGSF and ERO is obsoleted, closing.

PR #1953, opened 2022-05-30
Update OGI metadata

cthoyt:
1. Add Asiyah Yu Lin as contact @linikujp since on https://code.google.com/archive/p/ontology-for-genetic-interval/ she's listed as the first author for the publications
2. Switch status from orphaned to inactive (maybe also consider obsolete, as proposed in #1942, since this is on a defunct platform and doesn't enable standard OBO-required community interaction)

### DIFF
--- a/ontology/ogi.md
+++ b/ontology/ogi.md
@@ -3,11 +3,16 @@ layout: ontology_detail
 id: ogi
 description: An ontology that formalizes the genomic element by defining an upper class genetic interval
 homepage: https://code.google.com/p/ontology-for-genetic-interval/
+contact:
+  label: Asiyah Yu Lin
+  email: linikujp@gmail.com
+  github: linikujp
+  orcid: 0000-0002-5379-5359
 products:
   - id: ogi.owl
 title: Ontology for genetic interval
 tracker: https://code.google.com/p/ontology-for-genetic-interval/issues/list
-activity_status: orphaned
+activity_status: inactive
 ---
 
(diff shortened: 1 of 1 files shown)


matentzn:
So are you saying that the whole ontology was merged into https://obofoundry.org/ontology/ogsf.html?

If so, we should obsolete it and redirect users to OGSF instead. It is very important that an ID can always be clearly associated with one ontology. For example OGI:123 should either belong to OGI or OGSF, not both. Would you say that all off OGI is now managed by the OGSF, and that requests and issues should be raised on the OGSF repo? If so, I suggest we obsolete it and add "replaced by" OGSF. Users should always know where to find a term!

Are you ok to be listed as contact either way?

linikujp:
Yes. I will be the contact.

matentzn:
@linikujp can you comment on my question?

ontology/ogi.md
```diff
@@ -3,11 +3,16 @@ layout: ontology_detail
...
+  label: Asiyah Yu Lin
+  email: linikujp@gmail.com
+  github: linikujp
+  orcid: 0000-0002-5379-5359
 products:
   - id: ogi.owl
 title: Ontology for genetic interval
 tracker: https://code.google.com/p/ontology-for-genetic-interval/issues/list
-activity_status: orphaned
+activity_status: inactive
```
matentzn, line 15: change to "obsolete" and add a replaced_by link to ogsf (as per email discussions with @linikujp)

Keep contact information the way you have it here!

matentzn (changes requested):

ontology/ogi.md
cthoyt: done, but obsolete is its own field unrelated to activity status so note that it is now marked as inactive

matentzn:
Excellent. I will merge this on Monday. Thank you @cthoyt and @linikujp

linikujp:
Yes. I can be the contact.
And Yes the OGI is merged into OGSF.

PR #1970, opened 2022-06-18
Improve accessibility of home page

cthoyt:
This PR updates the aria-label attribute on buttons to better describe what the buttons do. This benefits disabled users who might be accessing the page by a screen reader.

Thanks @xkons for making me more aware of this stuff!

### DIFF
--- a/_includes/ontology_table.html
+++ b/_includes/ontology_table.html
@@ -52,14 +52,14 @@ Download table as: [ <a href="/registry/ontologies.yml">YAML</a> | <a href="/reg
       </td>
       <td>
         <a href="ontology/{{ont.id}}.html">
-          <button type="button" class="btn btn-default" aria-label="Left Align" title="More details">
+          <button type="button" class="btn btn-default" aria-label="More details for {{ ont.title }}" title="More details">
             <span class="glyphicon glyphicon-info-sign" aria-hidden="true"></span>
           </button>
         </a>
       </td>
       <td>
         <a href="{{ont.homepage}}">
-          <button type="button" class="btn btn-default" aria-label="Left Align" title="Project home">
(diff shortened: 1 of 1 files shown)


konstantintieber:
That's the spirit! Looking at the HTML, I think there's even more to be gained by removing the buttons inside the anchors altogether and putting the `aria-label` on the anchors.

This is of course a little more work, as I guess the buttons are there in the first place to get their styling. So in order to keep the current look of those links, there probably needs to be another CSS class applied to the anchor, which replicates the button style.

Buttons and anchors should generally be used "either or". I learned this one from @scottaohara: Anchors go places, buttons do things.

This stackoverflow answer includes an excerpt of the HTML spec why doing this is also not valid HTML: https://stackoverflow.com/a/6393863

_includes/ontology_table.html
```diff
@@ -77,7 +77,7 @@
       <td>
         {% if ont.contact %}
         <a href="mailto:{{ont.contact.email}}">
-          <button type="button" class="btn btn-default" aria-label="Left Align" title="Email ontology devs">
+          <button type="button" class="btn btn-default" aria-label="Send an email to {{ont.contact.label}}, the contact person for {{ ont.title }}" title="Email ontology devs">
```
matentzn, line 80: What happens if there is no contact? Pages are still rendered for obsolete ontologies?

cthoyt: There’s an if statement above that checks this

matentzn: oops. thx

matentzn (approved):
Excellent, thank you @cthoyt. Soon you will have more commits on this repo then me :D

PR #1971, opened 2022-06-20
Improve accessibility of homepage by switching buttons to links

cthoyt:
As a follow-up to #1970, this PR switches the buttons on the homepage to just be links. Luckily, bootstrap has the `btn` class that lets links look just like buttons, so there's no visual change from this PR.


cthoyt:
Screen Shot 2022-06-20 at 16 52 13

matentzn (approved):

Issue #1968, opened 2022-06-17
Provide guidelines and quality control on consistent short ontology descriptions


cmungall:
Examining the description field https://obofoundry.org/ we see a variety of different lengths, styles

- An ontology for
- A controlled vocabulary for
- A structured controlled vocabulary for
- Standard terms for
- A community ontology for
- An integrated ontology for
- An ontological representation of
- An ontology for the standardization of terminology for ..
- FOO is an ontology for
- FOO (The Foo Obvious Ontology) is an ontology for
- FOO describes
- <verb phrase immediately describing what the ontology does>
- 

This variance is potentially misleading, some users may be confused about the difference between "vocabulary" and "ontology" and think there is a meaningful difference when in fact the description is just reifying the start date of the ontology.

We should have a style guide for these, use consistent terminology, exclude adjectives that do not meaningfully separate ontologies ("integrated", "community", "structured", "formal", "standardized", "controlled" - these are all fuzzy, subjective, and ultimately meaning-free terms that all OBO ontologies should aspire to), keep meaningful adjectives ("upper", "application", taxonomic scope, etc).

We *may* want to keep the controlled vocabulary vs ontology distinction for the psi ontologies, since there is a meaningful difference in how these ontologies are expected to behave, but I think many of the ontologies that are currently described as vocabularies (primarily model organism ontologies) are as much ontologies as anything else

The ontology acronym or name should not be repeated in the description

After this, the descriptions vary wildly in conciseness, with some including random information about related projects, mini-essays, definitions of acronyms, parenthetical information, lists of other ontologies that are vaguely related or the ontology developer is friends with, ...

Sometimes species names are italicized, sometimes not.

We should have a style guide for the descriptions, this will help orient users who currently are presented with an overwhelming table and confusing descriptions. We should of course use reasonable judgment here - some ontologies such as OBI are a bit harder to concisely summarize than say an anatomy ontology, so some additional verbiage is useful and required.

I suggest the following template that allows for concise description of both content/scope and intended use:

- An [single optional meaningful adjective] ontology representing ... [, used for/in ....]

We can have templates for particular kinds of ontologies, for example:

Application ontologies:

- `An application ontology for [short description of purpose] covering [brief summary of contents]`

Anatomy ontologies:

- `An ontology representing the (gross anatomy | gross anatomy and cell types | gross and cellular anatomy) [and life stages] of <Formal taxon name> [ (common name) ]`

Avoid "development" to mean life stages

[...]

PR #1966, opened 2022-06-17
De-duplicate ontology descriptions

cthoyt:
Closes #1965 by removing the long-form text in 16 ontologies that matches the short form text and institutes a data integrity test to stop this from happening again. The following ontologies are affected:

1. AISM @JCGiron
2. APO @srengel
3. CHEMINF @egonw
4. CTO @akodamullil
5. FBbi @wawong
6. HAO @mjy
7. MAMO @gambardella
8. MPATH @PaulNSchofield
9. OARCS @mjy
10. OHMI @yongqunh
11. OPMI @yongqunh
12. RS @shurjenw
13. SPD @martinjramirez
14. WBbt @raymond91125
15. WBls @chris-grove
16. WBPhenotype @chris-grove

To all of the ontology responsible people, please respond ASAP if there's any reason this change is problematic. Again, note that in each of these 16 ontologies, the text was _exactly_ the same in the `description` field in the structured info and in the long form markdown area. I'd invite all of you to write longer descriptions for your ontologies in separate PRs as well, as many of them were quite short.

Issue #1965, opened 2022-06-17
Some ontologies have duplicate short and long descriptions


cthoyt:
The frontmatter can include a `description` field. 16 ontologies in OBO Foundry contain exactly the same text in the `description` field as the long-form markdown that follows the frontmatter. Potentially more contain text that's trivially different. We should do two things:

1. Remove the long-form text that duplicates the short-form text
2. Provide a test to make sure this doesn't happen again

matentzn:
These two fields have different purposes on the website, but I agree, there is no point to render the same thing twice, see here:

https://obofoundry.org/ontology/aism.html

Good effort!

Just one thing about SOP moving forward - this does not count exactly as a mass change, but I think it would be a good idea to get some feedback on issues before tagging more than 5 people to review it. In the first instance, the Technical Working Group should agree with a change to the metadata, then we can put it forward to the community. I know it "slows you down", but I think its a better process - because asking people to interact cost "social capital", which we don't have much off.. :)

nlharris:
See proposal from @cmungall in #1968

lschriml:
If the two descriptions are exactly the same, then I agree we can remove the long term text.
If the two differ, even trivially, then they should remain. 
It should be up to the ontology developers to decide how their ontologies are described.
Lynn

PR #1972, opened 2022-06-20
Truncate long short descriptions

cthoyt:
Related to #1966, #1965, and #1968, this PR adds some code to truncate the ontology short descriptions to 140 characters. Just like tweets, this is a pretty reasonable length and makes the homepage much easier to read. E.g., from the top ontologies, only PO's description is truncated (and the details link is already there to expand to the full ontology page):

Screen Shot 2022-06-20 at 15 26 57


matentzn:
This table is going to disappear soon! In the meantime, I am fine with anything. Have you tried this?

matentzn (approved):

PR #1980, opened 2022-06-23
Ensure shacl prefix map does not exclude obsolete ontologies from the prefix map

### LINKED ISSUE #1976: OBO prefix map should include prefixes of obsolete ontologies
I think OBO prefix map should contain prefixes of obsolete ontologies. Sometimes, like [here](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1974/files) one ontology is merged into another, with identifiers and all - it would still be useful to be able to loop these up.

@jamesaoverton objects to that?

matentzn:
fixes #1976

@jamesaoverton in the end this was only the shacl map, the obo_context was correct.

### DIFF
--- a/util/make-shacl-prefixes.py
+++ b/util/make-shacl-prefixes.py
@@ -43,8 +43,9 @@ def main(args):
     print(" sh:declare")
     sep = ""
     for ont in data["ontologies"]:
-        if ont.get("is_obsolete", False):
-            continue
+        #if ont.get("is_obsolete", False):
+        #    continue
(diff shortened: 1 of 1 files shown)


jamesaoverton (approved):

jamesaoverton:
This looks fine. I'm more concerned about how `registry/obo_context.jsonld` is generated, since I think that file is more widely used.

matentzn:
Yes me too. The OBO context already does the right thing! So there was no need to worry here. Thanks! 

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/registry/obo_context.jsonld#L160

PR #1985, opened 2022-07-05
Standardize YAML in ontology metadata frontmatter

cthoyt:
This PR does the following:

- [x] Introduces a Python script `util/standardize_metadata.py` that sorts and properly indents all of the metadata in all ontology metadata files' frontmatter. This script has no dependencies and can be run with `python util/standardize_metadata.py`. It's also included in the `lint` environment in `tox`
- [x] Adds a unit test to ensure that this metadata is sorted
- [x] Runs the standardization script and creates a big diff (can be run readily after the concept of this PR is reviewed and accepted)

This PR **does not** change any of the metadata itself. It is just reformatting it in a standard, sorted way.

In 6242c7f, I committed the automatic update to `aao.md` so you can see a bit more concretely what this "standardization" does. Here, you can see it sorts the keys alphabetically, canonicalizes quoting, and standardizes spacing in lists.

## Why do this?

It makes it much easier to programatically edit these files without creating massive diffs (after the massive diff in this PR, at least). There are several ways we might want to automatically maintain these metadata, such as the list of dependencies for each ontology.


util/standardize_metadata.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+
+"""This script sorts the ordering of the metadata in each ontology markdown file.
+
+Run with: ``python standardize_metadata.py``.
```
matentzn, line 5: Can we add an action that runs this script on any PR if a yaml file is changed, and automatically pushes the new order?

ontology/aao.md
```diff
@@ -1,17 +1,17 @@
 ---
-layout: ontology_detail
-id: aao
-title: Amphibian gross anatomy
```
matentzn, line 4: layout, id and title should be first (in that order), activity_status last.. The rest can be done alphabetically like you propose.

util/standardize_metadata.py
cthoyt: That's possible for branches off the origin. Not sure if you can arbitrarily have a bot pushing to forks. However, it's possible for maintainers to do it so I give it 75% likely to be okay

ontology/aao.md
cthoyt: The sorting is done by the YAML Dumper class, I'm not sure if the sort can be customized. I'll look into it, but no promises. Even if this isn't possible, I still think the consistent ordering is more valuable than the priorization

matentzn: I use a lot of "head" when working with these files locally, and also it helps me to quickly navigate the YAML file. I wouldn't say it for a cosmetic preference. The activity status isn't important, so you could split the YAML and dump twice.

cthoyt: ^ excellent thinking

cthoyt: Done in c748947

util/standardize_metadata.py
cthoyt: Not sure how the best way to go about testing this is, though. For now, you can run `tox -e lint` to run the linter for the yaml locally.

matentzn (approved):
We agreed to merge this for now on the condition that we find a way to update PRs with the reformatted YAML files according to the new sorting pipeline.

matentzn:
I will merge this PR on Saturday 9th July unless there is a veto.

PR #1988, opened 2022-07-12
Create permissible-metadata-content

nataled:
NOTE: This file begins the process of collecting information needed to fill out YAML files containing ontology metadata. The JSON schema is used for this purpose because it (presumably) contains all the fields that can be filled in. It will be prettied up as we go along.


faq/permissible-metadata-content.md
```diff
@@ -0,0 +1,15 @@
...
+
+## What can I put in the 'publications' field?
+Allowed number of entries: *any*<br>
+Each entry under this field should have one 'id' subfield and one 'title' subfield with content as indicated:
+- id: \<cited reference> The cited reference must take the form of a URI (e.g., https://...") that points to one of the following permitted resources: PubMed, PubMed Central, and the Digital Object Identifier (DOI) service. Note that preprint servers such as arXiv, bioRxiv, medRxiv, and ChemRxiv--as well as alternative hosting platforms such as Zenodo and FigShare--can still be used, but not directly. For example, instead of a direct link to Zenodo such as  https://zenodo.org/record/6685957, the equivalent DOI record https://doi.org/10.5281/zenodo.6685957 should be used. Links to Semantic Scholar (https://www.semanticscholar.org) or CEUR-WS (http://ceur-ws.org/) are not permitted. As an alternative, the Zenodo platform can be used to upload content that does not have a home in any of the permitted resources.
+
+- title: <title of cited reference> Technically this is a free-text field, but best practice is to use the title of the publication cited under the 'id' subfield.
```
cthoyt, line 15: wouldn't bother too much with this since we can automate getting the appropriate values for this

faq/permissible-metadata-content.md
```diff
@@ -0,0 +1,15 @@
...
+
+NOTE: In the instructions below, items in angle brackets--including the brackets <> themselves--should be replaced by the appropriate content.
+
+## What can I put in the 'publications' field?
+Allowed number of entries: *any*<br>
+Each entry under this field should have one 'id' subfield and one 'title' subfield with content as indicated:
+- id: \<cited reference> The cited reference must take the form of a URI (e.g., https://...") that points to one of the following permitted resources: PubMed, PubMed Central, and the Digital Object Identifier (DOI) service. Note that preprint servers such as arXiv, bioRxiv, medRxiv, and ChemRxiv--as well as alternative hosting platforms such as Zenodo and FigShare--can still be used, but not directly. For example, instead of a direct link to Zenodo such as  https://zenodo.org/record/6685957, the equivalent DOI record https://doi.org/10.5281/zenodo.6685957 should be used. Links to Semantic Scholar (https://www.semanticscholar.org) or CEUR-WS (http://ceur-ws.org/) are not permitted. As an alternative, the Zenodo platform can be used to upload content that does not have a home in any of the permitted resources.
```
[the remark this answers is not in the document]
nataled: Two reasons for this: (1) In general we avoid making value judgements in our explanations ("this site is good", "this site is bad"); and (2) the rationale itself doesn't provide the typical user with actionable information. What I mean is...well, let's take the Principles as examples. The provided rationales ('purposes' using the terminology there) are clearly understood and in themselves offer guidance. For this FAQ, the user would have to know that X is first-party while Y is not, and would have to reconcile the fact that DOIs are actually never first party. They'd also need to have some inkling that X has programmatically-accessible metadata while Y does not (which is the real issue, I believe). As we are aiming for simplicity, the EWG decided that the rationale cannot easily be understood, reconciled, or acted upon without a lot of background information, so opted to remove it.

faq/permissible-metadata-content.md
nataled: Since this is a subfield currently specified in the metadata file, for completeness it must be described. Once the automation is implemented, presumably the subfield will be removed and the description will be removed at the same time. Same situation as for the CURIEs; we cannot describe something not yet implemented. It's the price one pays for getting something quickly rather than waiting for full implementation.

matentzn (approved):

PR #1995, opened 2022-07-27
Add what to do if NOR liaison does not appear to 3 consecutive OFOC calls.

matentzn:
As discussed during yesterdays call.


nataled (changes requested):
Fix typos and wording:

"statement of wether they intend to continue being the liason."   --->
"statement confirming the ability to continue as liaison."

nataled (changes requested):
Sorry for this, but I missed other typos the first time around and needed to make other changes, so here's the whole part to begin after "will be assigned.":

1. At the next OBO Foundry Operations Committee conference call (hereafter, "Operations call"), a liaison is selected to be responsible for the issue. This liaison becomes familiar with the new ontology and rallies the appropriate people to provide feedback.
1. At the next Operations call after that one, the liaison presents the NOR to the OBO Foundry Operations Committee and answers questions. In most cases, the information provided will be sufficient to either grant or refuse the prefix request. In some cases, the committee may choose to postpone its decision to require some clarification and fixes from the submitter.
The liaison MUST be present at the Operations call in order for the NOR case to be discussed. If the liaison does not participate for 2 consecutive Operations calls, the chair of the second call emails the liaison to request a statement confirming the ability to continue as liaison. If the liaison does not participate in 3 consecutive Operations calls and did not respond to the email above, a new liaison is assigned during that third call.

nataled (approved):

PR #1996, opened 2022-07-28
Add permissible metadata instructions to navbar

matentzn:


nataled:
Some issues here:
1) Though I understand the reason why this could be considered a policy item, to me it seems more appropriate to put it with the other metadata-related FAQs under 'Ontology display'. Perhaps a better solution would be to create another section for 'Ontology metadata' and move 'How do I edit...' and 'What is the build object...' there, along with this new FAQ question.

2) It is our intention to add more information on other fields to the new page. It will not only be about publications. We thought the FAQ question would be something like "What content goes in the metadata file?" or similar.

3) That being said, considering there is only 'publications' at the moment, the question as written is fine. Almost. It lacks context. I suggest "What do I put in the 'publications' field of the metadata file?" as the displayed question for now. This does two things: (1) indicates what precisely is being addressed; and (2) reframes the issue as one of information rather than one of restriction (changing 'can' to 'do')

nlharris (approved):

nataled (approved):
Approved, though I think the question about flagging should be under 'display'. Will re-evaluate after changes take effect.

nataled:
@matentzn  Yes, I think the question about flagging should go under 'display' rather than metadata. Technically all these require editing the metadata, but the metadata section are for questions specifically mentioning 'metadata'. 

@nlharris One other change: there should be a question mark after 'How do I cite the OBO Foundry'

Issue #1963, opened 2022-06-16
Update Members page


nlharris:
https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/Membership.md

Add new members (Hande, Bill H, Shawn). Any old members to move to alumni?

Check affiliations and links. Add ORCID links where readily available.

See also issue #1358

nlharris:
See spreadsheet called "OBO Foundry Operations Committee members" in obo ops gdrive.

nlharris:
Somehow, when I did this update, I managed to leave off Hande and Bill H! Not sure how that happened.

PR #2005, opened 2022-08-02
Update members

nlharris:
as per discussion at OFOC meeting on 2022-07-26 and spreadsheet called "OBO Foundry Operations Committee members" (don't want to list the URL publicly here).
Added new members Shawn Tan and Hector Guzman-Orozco; moved some past members to Alumni; updated some affiliations; fixed some typos.

### DIFF
--- a/docs/Membership.md
+++ b/docs/Membership.md
@@ -23,45 +23,47 @@ Membership in working groups: T=technical working group, O=outreach working grou
 Current members of the OBO Foundry Operations Committee are (in alphabetical order of surname):
 
  * Jim Balhoff (T), RENCI, University of North Carolina, Chapel Hill, NC, USA
- * Mathias Brochhausen (O), University of Arkansas for Medical Sciences, Little Rock, AR, USA
  * Matt Brush (T), Unversity of Colorado Anschutz Medical Campus, Auroa, CO, USA
  * Pier Luigi Buttigieg (E, O), Alfred Wegener Institute, Helmholtz Centre for Polar and Marine Research, Bremerhaven, Germany
  * Seth Carbon (T), Lawrence Berkeley National Laboratory, Berkeley, CA, USA
- * [Melanie Courtot](http://purl.org/net/mcourtot) (O, T), EMBL-EBI, Cambridge, UK 
  * Alexander Diehl (T), University at Buffalo, Buffalo, NY, USA
  * [Damion Dooley](https://orcid.org/0000-0002-8844-9165) (O), Centre for Infectious Disease Genomics and One Health, Simon Fraser University, BC, Canada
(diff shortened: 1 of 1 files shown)


nlharris:
Oh yeah: fixes #1963

cthoyt:
Can you please add relevant start and end dates?

docs/Membership.md
```diff
@@ -23,45 +23,47 @@ Membership in working groups: T=technical working group, O=outreach working grou
...
- * [Jie Zheng](http://cbil.upenn.edu/profile-staff_bio/39) (T, O), University of Pennsylvania, Philadelphia, USA
+ * [Jie Zheng](http://cbil.upenn.edu/profile-staff_bio/39) (T, O), University of Pennsylvania, Philadelphia, PA, USA
  
 New members: follow the instructions on the [onboarding doc](https://docs.google.com/document/d/1MKhNTjZjGx6Ls72dybIV2ajYtbqtwP7O4lwxN2v3RBA/edit#heading=h.10q6n5qc13dp)
 
 Alumni:
 
  * Michael Ashburner
- * Colin Batchelor (E)
- * Eric Douglass (T)
```
cthoyt, line 62: why remove this information about what committees people served on? should that be a secret?

nlharris: I feel it was unreliable and doesn't add much, especially for alumni. And only some of them even had that info.

nlharris: > Can you please add relevant start and end dates?

Not easily, no.

cthoyt: at least for the people we're changing now we know the start and end dates. it would be nice to keep track of that going forwards

nlharris: Good idea to keep track. We know the start dates for the new additions. End dates are generally vague; I removed (alumnified) people who hadn't been active in some time, so it would be difficult to assign a specific end date. A removal date, sure.

cthoyt: I don't think you need to remove alumni, ever. This is a valuable history for the community. Only transfer people from active to alumni need to get given an "end date"

nlharris: I didn't remove any alumni (at least not on purpose) -- I think those two showed up on the 'deleted' side because I deleted the (possibly spurious) committee labels (E) etc.

cthoyt: Oh sorry I misread what you wrote. Glad we agree :)

nlharris: Cool. So can you approve this PR even though it's not 100% perfect? 🙂

cthoyt (approved):

matentzn (approved):

nlharris:
Thanks for the reviews. I'm not privileged enough to merge this.

PR #2007, opened 2022-08-02
Create bug-or-question.md

nlharris:
New category of issue template.
Note that I also suggest getting rid of ontology-review-request.md, since we no longer support human review of already-accepted ontologies.


matentzn (approved):
This is a very general issue template, but why not.

PR #2008, opened 2022-08-02
Delete ontology-review-request.md

nlharris:
I suggest deleting this as an issue template since we are not currently supporting human review of already-accepted ontologies, and there's another mechanism for submitting new ontologies for consideration.


matentzn (approved):
I approve this but others may disagree: @cstoeckert @wdduncan @lschriml

matentzn:
I guess the reason for removing this is because it could be confusing to people making a new submission? I think deleting may have to be preceded by collectively deciding we are not supporting reviews anymore. I agree with this for empirical reasons, but I think it may be cool to create a new format for that. Perhaps @cstoeckert review doc supersedes this part!

wdduncan (approved):

wdduncan:
I also approve. We can re-implement the template if needed.

wdduncan:
P.S. I don't have write access. So, my review doesn't count :)

cstoeckert:
The document I'm working on is tailored specifically for ontologies wanting to join the OBOF and would need to be rewritten to be appropriate for already accepted ones. I agree with @matentzn that "deleting may have to be preceded by collectively deciding we are not supporting reviews anymore" of ontologies already in the OBOF library. We may have already made that decision as @nlharris notes we have stopped doing those reviews.

nlharris:
> I agree with @matentzn that "deleting may have to be preceded by collectively deciding we are not supporting reviews anymore" of ontologies already in the OBOF library. We may have already made that decision as @nlharris notes we have stopped doing those reviews.

Right, this was my understanding -- that we had collectively agreed that we're no longer doing these reviews, so we don't want to mislead community members by inviting them to request such a review.

Issue #1861, opened 2022-04-22
Review criteria for admitting an ontology to the OBOF


cstoeckert:
This issue is to provide a draft stating criteria for ontologies to meet in order to be added to the OBO Foundry (OBOF). The topic has been discussed many times during operations committee calls and a request to start documenting the criteria was made at the April 19, 2022 call. 

Historical context: 
The OBOF has “reviewed” and library ontologies. The former are typically established reference ontologies and involved manual review for meeting OBOF principles. Library ontologies were admitted if committed to following OBOF principles and were not obviously in conflict with existing OBOF ontologies. The reviews and decisions to admit were done on a case by case basis. There have been cases of not admitting ontologies that were simply repeats of existing resources or not scientifically based. In the past couple years there has been a movement away from this approach motivated by a desire to get away from the reviewed versus library status and the introduction of the Dashboard to programmatically check meeting principles. 

Current review practice:
Progress has been made towards standardization of requirements to become part of the OBOF. An issue needs to created at https://github.com/OBOFoundry/OBOFoundry.github.io/issues and a detailed template (https://github.com/OBOFoundry/OBOFoundry.github.io/issues/new?assignees=&labels=new+ontology&template=new-ontology.yml&title=Request+for+new+ontology+%5BNAME%5D) needs to be satisfactorily filled out. The template includes a pre-registration checklist that essentially requires the submitter to agree to OBOF principles to check all the boxes. The process includes passing the provisional Dashboard (https://obofoundry.org/obo-nor.github.io/dashboard/index.html). It is possible to still have an ontology submitter agree to everything and the ontology pass the Dashboard but the ontology not actually be logically consistent, scientifically accurate, or follow the principles as intended. A review of some kind is made as part of discussion for admittance by the OBOF operations committee.

OBOF operations committee discussion points to be applied in future reviews:
1. We don’t want to admit ontologies whose use of imported terms or creation of new ones are problematic (e.g., inappropriately place imported classes, create/apply object properties that don’t make sense). Such logical inconsistencies will prevent interoperability with other OBOF ontologies and cause confusion if others try to use it. 

2. We don’t want to admit ontologies that claim to cover a domain but whose contributions are problematic (e.g., don’t really cover the area claimed, make assertions that are inaccurate). Such ontologies will be scientifically untrustworthy for others to use and will make it hard for other ontologies to provide better coverage of the domain. 

[...]

PR #2009, opened 2022-08-04
Reviewer SOP for new ontology requests

cstoeckert:
Two SOPs have been added to the page. One covers instructions for reviewers of new ontology requests and addresses https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1861 
The other addition is boiler plate for an email letting submitters know of acceptance and next steps. 

It is expected that further improvements will be needed to the text. This PR is meant to provide a mechanism for commenting on the text.


docs/SOP.md
```diff
@@ -11,6 +11,8 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
 
 ## SOPs
 - [New Ontology Requests](#NOR)
+- [Reviewing Ontologies for OBO Membership](#ROOM)
```
cthoyt, line 14: anchors automatically get generated based on the titles, so you could do
```suggestion
- [Reviewing Ontologies for OBO Membership](#reviewing-ontologies-for-obo-membership)
```

cstoeckert: Thanks for tip!

matentzn: I like fixed anchors better, just so we don't have to think about what happens when we change a title..

matentzn:
Excellent, does this SOP incorporate the feedback of the governance team?

cstoeckert:
@matentzn yes

nataled:
A concern regarding "Ontology scope". This part seems to mix multiple ideas, none of which is the scope as referred to in the Principles (which refers specifically to the scope of the content).  My take:
- "The new ontology must present some use cases where it will be relevant to life-sciences." This is P9.
- "Was the ontology developed using expert input or trusted scientific sources representative of the consensus in its target domain of knowledge?" This is (somewhat) P16.
- "If the ontology was developed for a very specific purpose or community, representation and consensus need not be broad; however, this scope should be clearly stated." This is a different type of scope, one that refers not so much to content but to target audience, or perhaps to the application vs reference categorization.

The final sentence of point 2 refers to scope as that term is used in the principles.

Point 3: What does 'accurate' mean here? Are the following two questions intended to provide guidance as to what it means? If so, adding a "That is" to the beginning of the first of the two followup questions will help clarify that. Also, considering the discussion around additional axioms (whether they should be allowed or not), it raises the question of whether this should be mentioned. Is the question intended to refer to additional axioms FOR the imported term, or USING the imported term? The latter is uncontroversial, but if that's what's intended, the wording should be revised to clarify.

Points 4 & 5 would benefit from an example of the correct way to say what's in those incorrect examples.

For the moment I suggest one revision in wording:

From: The new ontology must present some use cases where it will be relevant to life-sciences.
To: The new ontology must present use cases demonstrating its relevance to the life sciences.

There will definitely be more revisions as the EWG goes over it, but for now I intend to approve to facilitate further review.

nataled (approved):

cstoeckert:
Hi @nataled Thanks for the review! Did you really mean to merge? I was expecting to get more comments and make some edits before merging.
OK with you if I revert to allow that?

cstoeckert:
Reverted this so still collecting comments. Don't know how to move this from merged back to open. Looks like I need to make a new PR. Will do once others get a chance to review.

docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+Are imported terms in appropriate hierarchies? That is, has the import of the term preserved its upper-level alignment?
+Are any additional axioms used for these terms correct in both a technical (e.g. passes reasoning) and substantive sense?
+4. Basic review of axiomatic patterns:
+Are existential restrictions used correctly? Typical mistakes include “R some (A and B and C)” to mean “(R some A and R some B and R some C)”
+Are axioms generally highly complex? If so, we should review a handful to ensure they are as intended.
+5. Appropriate use of [object properties](https://www.w3.org/TR/2004/REC-owl-semantics-20040210/#owl_ObjectProperty). Examples of incorrect usage include those based on some interpretation of the label of the object property but not actually fitting the property definition or domain and range. A typical example of incorrect usage is R some (A and B and C) to mean R some A and R some B and R some C.
+6. Responsiveness to fixing changes. A willingness to fix any identified issues during the review must be demonstrated. Issues expected to be addressed should be added in TODO mode in the GitHub issue. The time limit for addressing these is 2 months.
```
nlharris, line 43: What is "TODO mode"?

docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+
+<a name="OAE"></a> 
+### Ontology Acceptance Email 
+
+Once a new ontology has been accepted, the following should be used as boiler plate for letting them know and what are the next steps they should take.
```
nlharris, line 48: > the following should be used as boiler plate for letting them know

the following should be used as a template for letting the ontology owners know

docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+
+"Thank you again for your ontology submission to the OBO Foundry. We are happy to inform you that your ontology (YOURID) has been accepted following discussion in the OBO Operations Committee meeting, DD-MM-YYYY. Before we can add it to the OBO ontology registry you need to complete the following steps.
+
+Create a metadata record for your ontology to be included in the registry
+1. Create a new file in this directory,  https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/ontology, called YOURID.md (there is an “Add file” button in the top right):
```
nlharris, line 53: is there an image following the :? If not, remove the :

nlharris (commented):
I think this looks good (I had only a few minor comments) but will review again in context once it's merged.

docs/SOP.md
cstoeckert: agree with change

docs/SOP.md
cstoeckert: agree with change

docs/SOP.md
cstoeckert: using GitHub checklist syntax (`- [ ] TODO`)

cstoeckert:
This PR has been replaced with https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2017

docs/SOP.md
sabrinatoro, line 43: Some might argue that some issues reported are not "actual" issues but modeling choices. Maybe add something to acknowledge this. For example : 
A willingness to fix any identified issues **and/or discuss modeling choices** during the review must be demonstrated.

Also, I would be a bit more generous about the time limit since some issues might take more than 2 months to address. So maybe change to:
"the time limit for addressing these is 2 months; if a longer period is needed, this should be communicated.

sabrinatoro (commented):
I have 2 suggestions, otherwise, it looks good.
Thank you!

PR #2013, opened 2022-08-05
Fix dividers in FAQ dropdown

cthoyt:
Now looks like this:
Screen Shot 2022-08-05 at 12 03 45


matentzn (approved):
Awesome man thank you!!

PR #2017, opened 2022-08-10
Review of new ontology request SOP

cstoeckert:
Addressed minor wording changes in PR#2009. That PR was merged and then reverted to give more time for review and get go ahead from operations committee.


docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+1. Create a new file in this directory,  https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/ontology, called YOURID.md (there is an “Add file” button in the top right).
+2. Obtain the already curated metadata that relates to your ontology from https://github.com/OBOFoundry/obo-nor.github.io/blob/master/dashboard-config.yml (you'll need to scroll down to locate yours; it will have your prefix in the " - id:" field.) 
+3. Create a pull request to add the metadata record. This pull request should include a link to this issue (the New Ontology Request issue).
+
+Here is an example record for the PATO ontology: https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/pato.md?plain=1
+
+Your metadata will be reviewed and merged by a member of the OBO Foundry Operations Committee. Permissable content for fields is being documented [here](https://obofoundry.org/faq/permissible-metadata-content.html).
+
+Create a PURL registry entry for your ontology.
+1. Go https://github.com/OBOFoundry/purl.obolibrary.org/tree/master/config, click “Add file” and add a file named YOURID.yml.
```
nlharris, line 62: Go -> Go to

docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+2. Add the desired configuration.
+3. Make a pull request with a link to this issue
+See here for an example: https://github.com/OBOFoundry/purl.obolibrary.org/blob/master/config/pato.yml "
```
nlharris, line 65: for an example -> for an example of a PURL yml file:

(as written, it looks like it's an example of a PR)

nlharris (approved):
Looks good to me! I found one typo and one place where wording could be made clearer.

matentzn:
Converted to draft PR, please set to ready for review if its ready for merging and I will review

docs/SOP.md
cstoeckert: done

docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+1. Ontology scope. The new ontology must present use cases demonstrating its relevance to the life sciences.  Was the ontology developed using expert input or trusted scientific sources representative of the consensus in its target domain of knowledge? If the ontology was developed for a very specific purpose or community, representation and consensus need not be broad; however, this scope should be clearly stated.
+2. Terms with the new ontology prefix. The new terms provided by the ontology should not already be available in an available OBO Foundry ontology. If there is an overlap of terms, justification should be provided. There must not be a term with the same label available. There should not be another OBO Foundry ontology whose scope covers those terms and if there is there should be some evidence of contacting that OBO Foundry ontology.
```
matentzn, line 35:
```suggestion
2. Terms with the new ontology prefix. There MUST NOT be a term with the same label available in another OBO Foundry Ontology. There SHOULD NOT be another OBO Foundry ontology whose scope covers any of the new terms. If there is an overlap of terms with another OBO Foundry ontology, justification MUST be provided. An example of an acceptable justification is that the other OBO Foundry ontology was contacted and the request for adding the terms in scope for that ontology where rejected. In this case, evidence of this contact must be provided to the reviewer on request.
```

matentzn: This paragraph is so important, I decided to give it a shot to rewrite it before merging. It does not change the meaning, but I think adds clarity on what exceptions are permissible.

matentzn: @cstoeckert

cstoeckert: Thank you for reviewing this @matentzn The current text was written in response to the feedback from the Governance Task Team (see the google doc for more on this): 
"Therefore we suggest that this point to be change to 1) use less strong wording (i.e. avoid “must not”), and 2) express that this practice should be avoided, and, when used, should be justified. Also, clear guidance should be given to reviewers to understand how this can be checked (how would one know what term to check?)." 
I think some of your edits go against that. Your suggested justification example could be added though:
"An example of an acceptable justification is that the other OBO Foundry ontology was contacted and the request for adding the terms in scope for that ontology was rejected. In this case, evidence of this contact must be provided to the reviewer on request."

cthoyt: there's a lot of MUSTs now, what's the recourse when these aren't followed? Kicked out of OBO?

cstoeckert: @cthoyt This SOP is for review of ontologies asking to join the OBO Foundry so the recourse is a recommendation against admission. If there are "must"s that you think would be better as "should"s please point them out. Thanks!

cthoyt: no, I'm all for the MUSTs! But what happens when an ontology follows the rules to get in but then stops following the rules? For example, I know that my old group is trying to get many ontologies accepted into OBO and it's their _modus operandi_ to abandon things once they can say to the bosses "look we are legitimate because OBO said so"

cstoeckert: That is definitely an issue but to be covered in a different SOP. :-)

matentzn: I will have this rechecked by the Governance team.

sabrinatoro: There are a lot of "MUST" and "MUST NOT" in the new version, which doesn't give space for exceptions (which we already have a lot in already approved obo ontologies. I know this is a different issue, but one could make the argument). I think the message should be: This must be avoided at all costs, but if it must happen, (good) justification(s) is(are) required.
- "There MUST NOT be a term with the same meaning available in another OBO Foundry Ontology. "
I agree. Basically, we don't want 2 identical terms (ie 2 terms with the same concepts), whether or not the label is the same
- "Where meaning is ambiguous, a clear justification for the interpretation of both terms MUST be given (labels, definitions, axioms)." 
This is a dangerous sentence... It assumes that one would have to argue for the interpretation of a term from an ontology that is not theirs. 

Proposal: 
There MUST NOT be a term with the same meaning available in another OBO Foundry Ontology, ie there must not be a term referring to a concept that already exists in another OBO Foundry Ontology (whether or not the label is identical). There SHOULD NOT be another OBO Foundry ontology whose scope covers any of the new terms. In the event that these conditions cannot be fulfilled, justification(s) must be provided. Such justification(s) include: 
- the demonstration that these terms are actually not the same (this happens when term meaning/concept is ambiguous); or
- the other OBO Foundry ontology was contacted and rejected the request for adding new terms in scope for that ontology. 

(what about terms with same labels? should we add this too? )

docs/SOP.md
```diff
@@ -23,6 +25,45 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+Are any additional axioms used for these terms correct in both a technical (e.g. passes reasoning) and substantive sense?
+4. Basic review of axiomatic patterns:
+Are existential restrictions used correctly? Typical mistakes include “R some (A and B and C)” to mean “(R some A and R some B and R some C)”
+Are axioms generally highly complex? If so, we should review a handful to ensure they are as intended.
+5. Appropriate use of [object properties](https://www.w3.org/TR/2004/REC-owl-semantics-20040210/#owl_ObjectProperty). Examples of incorrect usage include those based on some interpretation of the label of the object property but not actually fitting the property definition or domain and range. A typical example of incorrect usage is R some (A and B and C) to mean R some A and R some B and R some C.
+6. Responsiveness to fixing changes. A willingness to fix any identified issues during the review must be demonstrated. Issues expected to be addressed should be added using GitHub checklist syntax (`- [ ] TODO`) in the GitHub issue. The time limit for addressing these is 2 months.
```
sabrinatoro, line 43: Some might argue that some issues reported are not "actual" issues but modeling choices. Maybe add something to acknowledge this. For example :
A willingness to fix any identified issues **and/or discuss modeling choices** during the review must be demonstrated.

Also, I would be a bit more generous about the time limit since some issues might take more than 2 months to address. So maybe change to:
"the time limit for addressing these is 2 months **; a longer period should be requested if needed.**

sabrinatoro (commented):
I have 2 comments. Let me know if you have any questions.
@matentzn

docs/SOP.md
matentzn: I like this. I updated my proposal to reflect exactly your suggestion @sabrinatoro. Can you confirm it is ok now?

We don't need to say anything about labels, because labels are only a proxy for meaning anyways.

docs/SOP.md
matentzn: @sabrinatoro I like this, please use `suggestion mode` to make a suggestion (on mac, start writing a comment then hit cmd+g)

docs/SOP.md
sabrinatoro: @matentzn It's perfect! Thank you!

docs/SOP.md
sabrinatoro, line 43:
```suggestion
6. Responsiveness to fixing changes. A willingness to fix any identified issues during the review must be demonstrated. Issues expected to be addressed should be added using GitHub checklist syntax (`- [ ] TODO`) in the GitHub issue. The time limit for addressing these is 2 months; a longer period should be requested if needed.
```

sabrinatoro: @matentzn I think I did it (?)

matentzn (approved):

matentzn (approved):

docs/SOP.md
```diff
@@ -23,8 +25,46 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
...
+1. Ontology scope. The new ontology must present use cases demonstrating its relevance to the life sciences.  Was the ontology developed using expert input or trusted scientific sources representative of the consensus in its target domain of knowledge? If the ontology was developed for a very specific purpose or community, representation and consensus need not be broad; however, this scope should be clearly stated.
+2. Terms with the new ontology prefix. The new terms provided by the ontology should not already be available in an available OBO Foundry ontology. If there is an overlap of terms, justification should be provided. There must not be a term with the same label available. There should not be another OBO Foundry ontology whose scope covers those terms and if there is there should be some evidence of contacting that OBO Foundry ontology.
```
matentzn, line 35:
```suggestion
2. Terms with the new ontology prefix. There MUST NOT be a term with the same meaning available in another OBO Foundry ontology, ie there must not be a term referring to a concept that already exists in another OBO Foundry ontology (whether or not the label is identical). There SHOULD NOT be another OBO Foundry ontology whose scope covers any of the new terms. In the event that these conditions cannot be fulfilled, justification(s) MUST be provided. Such justification(s) include:
```

matentzn (approved):

PR #2015, opened 2022-08-09
add new "OBO Dashboard" section to FAQ menu

nlharris:
Along with #2014, fixes #1930

(sorry this is on a separate branch; i tried briefly to remember how to add (from the web interface) another PR and decided it wasn't worth spending more time on that)

PR #2019, opened 2022-08-10
New FAQ page

### LINKED ISSUE #2016: Create seperate FAQ page
The drop down menu on the site is getting too long.

nlharris:
replaces PR #2015 (which added the new dashboard section to the FAQ menu).
fixes #2016

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -1,9 +1,7 @@
 ---
 layout: faq
-title: FAQs
+title: OBO Foundry Frequently Asked Questions
 ---
-
-## OBO Foundry Frequently Asked Questions (FAQs)
 
 ### Policy
 - <a href="/principles/fp-000-summary.html">What are the OBO Foundry Principles?</a>
--- a/faq/index.md
+++ b/faq/index.md
@@ -1,0 +1,47 @@
+---
+layout: faq
+title: FAQs
+---
+
+## OBO Foundry Frequently Asked Questions (FAQs)
+
+### Policy
+- <a href="/principles/fp-000-summary.html">What are the OBO Foundry Principles?</a>
+- <a href="/id-policy.html">What is the OBO ID Policy?</a>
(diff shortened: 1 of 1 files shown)


cthoyt:
How would a user know to navigate to this page? Shouldn't this PR also remove the entire drop-down list from the navbar?

nlharris:
i was going to do that part after this page is merged so that i can check that it looks ok.

faq/index.md
```diff
@@ -0,0 +1,47 @@
+---
+layout: faq
+title: FAQs
+---
+
+## OBO Foundry Frequently Asked Questions (FAQs)
```
cthoyt, line 6: this is taken care of by the `title` in the front matter and is therefore redundant

faq/index.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Policy
+- <a href="/principles/fp-000-summary.html">What are the OBO Foundry Principles?</a>
```
cthoyt, line 9: would prefer using markdown style links where possible

nlharris: totally agree but i converted the menu to this page using a few regexps. to rewrite it in md would require either more extensive scripting or tedious editing. or i guess i could try an html-to-md converter, if you feel strongly about it.

faq/index.md
nlharris: thanks, fixed

faq/index.md
cthoyt: it's not a big deal but markdown syntax is much more straightfowards and less error-prone.

nlharris: if i had created the page from scratch i would have used md. it didn't seem worth it to me to spend a lot of time to convert the html to md -- but if you would like to do that, please feel free! 🙂

matentzn (approved):
This is great, thank you @nlharris and @cthoyt!

PR #2022, opened 2022-08-16
delete obsolete faq.html

### LINKED ISSUE #2018: archive faq.html
https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/faq.html (which makes https://obofoundry.org/faq.html) exists but is obsolete and not linked anywhere. (When I searched for occurrences of faq.html, I found on https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/faq/how-do-i-edit-content.md, which has a link to a faq.html but not this one, and the one it points to doesn't exist.)

Can we just trashcan faq.html without losing it from the history?

nlharris:
fixes #2018


matentzn (approved):
Thank you @nlharris!

PR #2030, opened 2022-08-18
Change ontobee URL

### LINKED ISSUE #2027: Use https version of Ontobee.org
At the top of the obofoundry.or webpage: 
https://obofoundry.org/
there is a search bar. 

current it uses http version of ontobee. for example, when searching for term 'vaccine', it goes to the insecure webpage:  
http://www.ontobee.org/search?ontology=&submit=Search+terms&keywords=vaccine

Since ontobee.org has recently updated to use https instead of http, I would suggest changing
http://www.ontobee.org/
to: 
https://ontobee.org/
such as: 
https://ontobee.org/search?ontology=&submit=Search+terms&keywords=vaccine

Thanks, Oliver

matentzn:
Fixes #2027

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-ontology.yml
+++ b/.github/ISSUE_TEMPLATE/new-ontology.yml
@@ -256,13 +256,13 @@ body:
           required: false
         - label: My ontology has a specific release file with a version IRI and a `dc:license` annotation, serialised in RDF/XML.
           required: false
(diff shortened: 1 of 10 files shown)


yongqunh (commented):
I checked, and they look all good. Thanks! 
- Oliver

yongqunh (approved):

nlharris (approved):
at a quick glance, this looks fine to me, though i can't verify that you caught all the instances of http://[www].ontobee.

matentzn:
I didn't catch all of them, about 15 ontology pages have links to the old ontobee.. since these will continue to work, it's fine. People can fix their own metadata when it is necessary!

yongqunh:
Looks great! Thanks!

PR #2031, opened 2022-08-18
Update navbar

cthoyt:
This gets rid of the redundant navbar and updates the footer to look slightly nicer. This also deletes unused `footer.html` and `navigation.html` templates which seem like they're artifacts from the template originally used to create this site. Here are some screenshots:
Screen Shot 2022-08-18 at 23 20 34
Screen Shot 2022-08-18 at 23 20 26


nlharris (approved):
Looks good, thanks for doing this.

matentzn (approved):

Issue #1994, opened 2022-07-27
Improvements for new front page table 


matentzn:
This is a sink for all the feedback I decided not to force for the first version:

- [x] @nlharris: If the down-arrow for sort changed to an up-arrow for sorting the other way, that would be extra cool (but not worth spending a lot of developer time on).
- [x] @bpeters42: I would uncheck the 'group by domain' field in the default load view and start off with the flat list. (@matentzn disagrees, as the group by domain is part of his great plan to avoid argument about how the table should be ordered by default)
- [x] @nlharris: Space usage: I would prefer if each ontology used up as little vertical space as possible, to maximize the number of ontologies you can see on one page w/o scrolling. The descriptions often take multiple lines; meanwhile, there's a lot of horizontal padding esp. on the Quick Access column that could be reduced and donated to the Description column to help it wrap less. (@matentzn disagrees, he thinks that modern web design gives more space, and he likes it the way it is. We can have a vote!)
- [x] @cthoyt: [140 characters (for description) might be too long given this layout, some of the descriptions get up to 6 lines. This also needs a right-strip of spaces since the ellipses is applied inconsistently.](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1978#discussion_r933984190)
- [x] @cthoyt: Consider [removing redundant (i) button.](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1978/files/2dff091924a33f306e1b1654fcf2d7b109c9bf96#r933982472)  - @matentzn agrees.
- [x] @cthoyt: Consider trying to align the icons a bit better (publication is optional, so it can cause some non-alignment across rows)
- [x] @matentzn: Search bar does not search title field

cthoyt:
it's going to be a massive diff but I think it's time to upgrade bootstrap and stop vendoring dependencies, too.

nlharris:
I believe all of the listed improvements have been implemented, and therefore this issue can be retired in triumph! 🙌

cthoyt:
In the immortal words of daft punk

> our work is never over

but for sure it's a good time to close this one as everything is all done

PR #2032, opened 2022-08-19
Table should search title field

matentzn:
See #1994 

Right now, the table search does not consider the Title field. It should - I hope this PR does that?

cc @despean

### DIFF
--- a/_includes/js/custom.js
+++ b/_includes/js/custom.js
@@ -360,8 +360,12 @@ jQuery(document).ready(function() {
                 if (row.description === undefined) {
                     row.description = ""
                 }
+                if (row.title === undefined) {
+                    row.title = ""
+                }
                 return (row.id.toLowerCase().includes(term) ||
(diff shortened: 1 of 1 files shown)


cthoyt:
Code change makes sense. In practice, most ontologies duplicated their titles in their descriptions so this was a little tricky to reproduce.

cthoyt (approved):

_includes/js/custom.js
```diff
@@ -360,8 +360,12 @@ jQuery(document).ready(function() {
                 if (row.description === undefined) {
                     row.description = ""
                 }
+                if (row.title === undefined) {
```
cthoyt, line 363: this code could assume that all titles exist since this is a pretty consistent requirement of OBO Foundry

matentzn: Just sticking with the pattern that was already there, but yes, I agree. It should :D

nlharris (approved):

PR #2035, opened 2022-08-20
Deactivate sort by foundry status

matentzn:
Now we have a front page which breaks up the site by domains by default:

image

This PR will deactivate sort by foundry status so we don't get these odd sorts in the default view (this is foundry status first, then alphabetical):

image

I will leave this open until Wednesday 26th September.


shawntanzk (approved):
OFOC meeting call agreed with this

nataled (approved):

PR #2038, opened 2022-08-21
Remove stray debug command

erik-whiting:
Hi! If you visit the [homepage](https://obofoundry.org/) in Chrome with dev tools open, the loading will stop on this stray debug command. The `debugger` word in JavaScript is equivalent to a breakpoint, but they only work if a browser's dev tools equivalent is active at the time the `debugger` line is reached.

Most people will not likely be affected by this bug except for developers who may have dev tools open while looking at the source or messing around in the js console. Anyway, removing this line fixes the problem, let me know if you have any questions!


matentzn (approved):
Excellent @erik-whiting thank you very much for this!

PR #2040, opened 2022-08-21
Toggle arrow direction when sorted by column

erik-whiting:
See #1994 

When sorting a table by a specific column, toggle the direction of the arrow in that column appropriately.

I did this by passing the chevron element to the `setColumnHeaderSort` method and setting its class to the desired `glyphicon` class. Do note that for "descending" the arrow points up and vice versa. This doesn't make linguistic sense, but when you look at it, it looks right (at least in my opinion).

Let me know if you'd like me to change anything!

### DIFF
--- a/_includes/js/sorttable.js
+++ b/_includes/js/sorttable.js
@@ -55,7 +55,7 @@ class SortableTable {
    * Takes the index of the column to be sorted parses sort parameters from html
    * @param {string|number} columnIndex
    */
-  setColumnHeaderSort(columnIndex) {
+  setColumnHeaderSort(columnIndex, chevron) {
     if (typeof columnIndex === 'string') {
       columnIndex = parseInt(columnIndex);
     }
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Oh cool! Thanks again! There seem to be General debates around this; https://ux.stackexchange.com/questions/37564/use-up-or-down-arrow-to-represent-sort-ascending-at-table-header

I am happy to go with your intuition for now and see what others think! 

Thank you!!

balhoff:
I would prefer they point the opposite direction. 😛 That would be inline with most interfaces I use.

erik-whiting:
@balhoff  Just to be clear, you're saying you prefer:
🔼 
a
b
c
d

and
🔽 
d
c
b
a

is t hat correct?

balhoff:
Yes, that's what I personally prefer. I'm happy to go along with consensus though (and I'm not suggesting it's of major importance). A couple random examples:

- https://www.ibm.com/docs/en/zos/2.4.0?topic=zosmf-sorting-tables
- https://datatables.net/examples/styling/order-column.html

erik-whiting:
a made a PR switching the arrow direction if it's decided to be the more favorable design 😄  https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2059

PR #2042, opened 2022-08-21
Update summary page

cthoyt:
Some invalid HTML (i.e., an unclosed div) made the summary page look wrong after adding the new footer in #2031. This PR does the following to the ontology detail page:

1. Removes superfluous divs
2. Removes the unused disqus code
3. Adds a bioregistry link

| before | after |
|:------:|:------:|
| Screen Shot 2022-08-26 at 12 58 08 | Screen Shot 2022-08-26 at 12 58 19 |

## Dependencies

- [x] #2049 
- [x] #2052


matentzn:
Awesome thank you @cthoyt - much appreciated!

This PR is one huge diff - could you

1. Make a PR that changes nothing other than formatting (I merge it in without much question), and then 
2. another with the functional changes? 

Else I wont be able to really review this..

matentzn:
@cthoyt merged the blocker

cthoyt:
@matentzn now this PR only contains functional changes but it's also been prettified so the diff remains big. I've added a side-by-side comparison in the top as well as in this comment so you can see there are no major style changes, besides removing an unnecessary grey box

| before | after |
|:------:|:------:|
| Screen Shot 2022-08-26 at 12 58 08 | Screen Shot 2022-08-26 at 12 58 19 |

matentzn (approved):
Ok @cthoyt I have reviewed about 20% of the PR, and I am well enough convinced its a valuable change, and the change corresponds to your description. Awesome! Thank you as always!

As an aside, the next time we add functionality like adding a button, please make an issue first so we give people a chance to object. I am using a bunch of credits from my role now as TWG Metadata General to basically make this decision on behalf of everyone else, but doing this more than once or twice per quarter will diminish my ability to influence bigger issues in the future, because people will lose trust in me if I bypass decisions like that.

PR #2052, opened 2022-08-26
Lint HTML and markdown with prettier

cthoyt:
This PR runs `npx prettier --write .` to reformat HTML and markdown in a principled, deterministic way. This will make it easier to make updates that have smaller diffs, such as #2042. It adds a note on how to do this in the README-sitedev.md

It also adds a `.prettierignore` file that lists some file types and directories to ignore.

This PR **does not** change how anything looks on the website. However, in 3 cases (see 5ec2d01) it fixes some invalid HTML.

### DIFF
--- a/README-sitedev.md
+++ b/README-sitedev.md
@@ -150,5 +150,5 @@ We use bootstrap 3, so far no themes.
 HTML code is kept nice looking with [`prettier`](https://prettier.io):
 
 ```shell
-npx prettier --write _includes/ _layouts/
+npx prettier --write .
 ```
--- a/README-sitedev.md
+++ b/README-sitedev.md
@@ -147,8 +147,6 @@ We use bootstrap 3, so far no themes.
 
 ### Code quality
 
-HTML code is kept nice looking with [`prettier`](https://prettier.io):
-
-```shell
-npx prettier --write .
-```
+1. Install the Node Package Manager (NPM) following [these instructions](https://docs.npmjs.com/downloading-and-installing-node-js-and-npm)
+2. Install [`prettier`](https://prettier.io) with NPM using `npm install --save-dev --save-exact prettier`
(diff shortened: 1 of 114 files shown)


_includes/navbar.html
```diff
@@ -1,102 +1,236 @@
...
+      <li class="dropdown">
+        <a
+          href="#"
+          class="dropdown-toggle"
+          data-toggle="dropdown"
+          role="button"
+          aria-haspopup="true"
+          aria-expanded="false"
+          >Citation <span class="caret"></span
+        ></a>
```
matentzn, line 159: Is this how the new linters like it? Looks a bit odd, do you know the motivation for it?

README-sitedev.md
```diff
@@ -145,3 +145,10 @@ current wiki, but other opinions welcome.
 
 We use bootstrap 3, so far no themes.
 
+### Code quality
+
+HTML code is kept nice looking with [`prettier`](https://prettier.io):
+
+```shell
+npx prettier --write .
```
matentzn, line 153: Can we get a bit more detail here? Dont we have to install npx prettier first?

Also, how do you see this getting kept up to date? Any chance for a github action that runs this once in a while? Or whenever and md/html file is changed?

matentzn (approved):
Man this is awesome, I was always wondering about markdown linters, and here is my answer. I checked every single file with an 87% rigour, and it looked great. I have two questions and they are not blockers on merging this, but I will wait for a moment for you to answer before merging.

_includes/navbar.html
cthoyt: I guess so. We don't have to argue with what the linter does, assuming it's deterministic, even if we don't really like it ourselves.

matentzn: No problem, I agree! Just curious.

README-sitedev.md
cthoyt: I just update the installation docs. There are indeed github actions to automatically run this, but I didn't want to make a first PR too complicated

matentzn:
Awesome!

Issue #2051, opened 2022-08-26
Create shield.io badges for ontologies with non-standard licenses


matentzn:
Just for optical reasons. Instead of:

<img width="1264" alt="image" src="https://user-images.githubusercontent.com/7070631/186834331-6367793e-0a1a-4b5e-9283-787fdb799b20.png">

We do 

[![License](https://img.shields.io/badge/License-HPO-green)](https://hpo.jax.org/app/license)
[![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
[![License: Artistic-2.0](https://img.shields.io/badge/License-Artistic%202.0-0298c3.svg)](https://opensource.org/licenses/Artistic-2.0)
[![License: GPL v3](https://img.shields.io/badge/License-GPL%20v3-blue.svg)](https://www.gnu.org/licenses/gpl-3.0)

erik-whiting:
Hello! Question for you. I've made--I think--the necessary updates to make this happen, however I'm pretty sure I need to run the methods in `utils/extract-metadata.py` to update the site in general, is this true? I'm using Windows, so using the Makefile isn't super straightforward, any advice for generating the rest of the needed changes?

For reference, I've added the following lines to `util/extract-metadata.py#decorate_metadata` method:
```python
  elif lurl.find("hpo.jax.org/app/license"):
      logo = "https://hpo.jax.org/app/license"
  elif lurl.find("apache.org/licenses/LICENSE-2.0"):
      logo = "https://opensource.org/licenses/Apache-2.0"
  elif lurl.find("opensource.org/licenses/Artistic-2.0"):
      logo = "https://opensource.org/licenses/Artistic-2.0"
  elif lurl.find("gnu.org/licenses/gpl-3.0.en.html"):
```

cthoyt:
@erik-whiting if you make a PR I can answer the question. The extract metadata script is run on any change to one of the ontology files by GitHub actions but it’s also possible to run locally with just ‘make’

erik-whiting:
Hey @cthoyt, here's the PR: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2062 let me know if it works the way I expect. Thanks!

PR #2062, opened 2022-08-28
Add links to shield.io badges

erik-whiting:
This PR aims to address the issue in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2051

I don't actually know if this will behave the way I expect since I can't figure out how to generate the new metadata. Hopefully someone can generate the changes and see if it works.


cthoyt (changes requested):
Here's how to prepare the new data (this does what's expected - updates `_config.yml`):

```shell
vf new -p python3.7 obo37  # or whatever to make an old venv, the code is super old
python -m pip install -r requirements.txt
make
```

You can run the website locally with docker (don't even bother installing jekyll yourself, it's a total mess):

```shell
docker run --rm --volume="$PWD:/srv/jekyll" -p 4000:4000 -it jekyll/jekyll:latest jekyll serve
```

However, the content of this PR doesn't make any sense. The logos are supposed to be URLs to images that represent the licenses, not URLs to pages describing the licenses themselves. 

A better way to address the ugliness of non-standard licenses is probably to do some conditionals, somewhere in the new javascript logic. Tbh, I don't think we should be investing ~any~ a lot of time or effort supporting non-standard licenses. For example, the HPO license is restrictive w.r.t. reuse and is awfully antithetical to the OBO principles. Similar issues exist for the other ones as well.

erik-whiting:
Thanks for reviewing this. I mentioned in the linked issue that I'm using Windows and `make` is not really cooperating with me; neither with `venv` nor `virtualfish`. Since I'm having a hard time using the tools and my first attempt "doesn't make any sense" I'm probably out of my depth trying to address the issue so I'll close the PR. Thanks for your time!

PR #2063, opened 2022-08-28
Update ENVO.md

cthoyt:
Following https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2065, https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2066, #2067, and https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2070, this PR updates ENVO. It needs some input from @pbuttigieg on which of these should be the "preferred" citation. The two that were deleted from the body of ENVO.md are exactly the same two that appear in the structured metadata.


matentzn (changes requested):
Another awesome drop, thanks so much! 

If you want me to merge this quickly, you will have to revert eco.md, envo.md and fbbt.md changes. The problem with these three is that the citation information conveys another piece of information not covered by the citation record: "preferred citation". I cannot unilaterally permit this change, as this is not merely a "minor shift" of identical information content from one place to another (like in all the other cases), but involves a loss of information as well. 

@Clare72 @pbuttigieg @mgiglio99 would have to give explicit consent to this change (which could take a few days).

For all the other changes, I have proposed this [amendment to our metadata change SOP](https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2065) which I hope to get accepted quickly. I would probably break the rules a tiny bit as well, because the changes should have zero controversies: You are simply moving the publication metadata which is currently informally present in the ontology description into the structured publication metadata field.

@lschriml I would value your thoughts on this issue because you had previously suggested the new and much-needed SOP, and tell me if you agree with me on #2065.

matentzn:
## Metadata permission change granted

As per [SOP](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/SOP.md), we give a 4 week warning for the metadata change. @cthoyt, I am pretty convinced this [is not necessary](#2065) for any but the first 3.

- [ ] ECO (@mgiglio99)
- [ ] ENVO (@pbuttigieg)
- [ ] FBbt (@Clare72)
- [ ] PORO (@bobthacker)
- [ ] CIO, HOM (@fbastian)
- [ ] INO (@yongqunh)
- [ ] MIAPA (@hlapp)
- [ ] OHPI (@e4ong1031)
- [ ] OMP (@jimhu-tamu)

cthoyt:
I'll split this PR up so we can merge the style updates while we wait on these

Clare72:
As @matentzn says, I think it would be good to be able to still specify the preferred citation somehow. Perhaps for now this could done be by appending " (preferred citation)" to the publication title?
i.e.
```
publications:
- id: https://www.ncbi.nlm.nih.gov/pubmed/24139062
  title: The Drosophila anatomy ontology (preferred citation)
```

matentzn:
I think this is hacky - a boolean flag "preferred: true" is the correct way to deal with this IMO.

Clare72:
sure - as long as there is some way to indicate which reference is preferred that will show up in the display.

cthoyt:
@Clare72 @matentzn see #2070 for adding preferred citation badge.

Clare72:
I am happy with this solution

ontology/envo.md
```diff
@@ -52,6 +52,7 @@ products:
 publications:
 - id: https://www.ncbi.nlm.nih.gov/pubmed/24330602
```
pbuttigieg, line 53:
```suggestion
- id: https://dx.doi.org/10.1186/2041-1480-4-43
```

ontology/envo.md
```diff
@@ -52,6 +52,7 @@ products:
...
   title: 'The environment ontology: contextualising biological and biomedical entities'
+  preferred: true
 - id: https://www.ncbi.nlm.nih.gov/pubmed/27664130
```
pbuttigieg, line 56:
```suggestion
- id: https://dx.doi.org/10.1186/s13326-016-0097-6
```

pbuttigieg (changes requested):
In addtion (I can't seem to comment / suggest changes on the other lines)

- [ ] Are we still using Travis? I thought we're now using GH Actions since Travis changed its billing policies
- [x] [License] label: CC0
- [ ] [License] url: https://creativecommons.org/share-your-work/public-domain/cc0/
- [ ] There are other subsets, should we list them all?
- [ ] email_cc: pier.buttigieg@awi.de
- [x] email: pier.buttigieg@awi.de
- [x] description: An ontology of environmental systems, components, and processes.
- [x] [Jobs] id: https://github.com/EnvironmentOntology/envo/blob/master/.github/workflows/qc.yml

On the job ID, I'm not sure if that's canonical. There are run IDs of the form:  https://github.com/EnvironmentOntology/envo/runs/7909529777

Some of the fields above seem to appear more than once.

cthoyt:
@pbuttigieg I'm happy to change the identifiers from pubmed to DOI, but is there a meaningful difference? Besides that, this PR is only about which one is "preferred" - do you have a comment on that? otherwise I'd suggest we discuss other changes in a different place, perhaps you can make a PR for those yourself.

ontology/envo.md
matentzn: dx.doi.org is not an allowed prefix, can we change to doi.org? (works just was well)

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/10dd17d93ef4537cbab02e35a3c867e1ab100c36/tests/test_integrity.py#L23

https://doi.org/10.1186/2041-1480-4-43

matentzn:
I made the requested changes to the metadata, and also opened a new issue to figure out of the jobs field with CI information is really of relevance to OBO: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2085

matentzn (approved):

matentzn:
@pbuttigieg I "dismissed" your review (in the GitHub UI) because I want to merge this - and it wouldn't let me with your "request changes" status. All your points are addressed though! If you have any problems with any of the changes let me know!

PR #2064, opened 2022-08-28
Improve home page column spacing

cthoyt:
This PR improves the spacing of columns on the home page by removing all of the `col-sm-X` classes and making the button group have a flex display. It also removes the redundant link to the ontology detail page in each row in the quick access buttons. Now it looks like:

Screen Shot 2022-08-28 at 23 15 26


matentzn (approved):
AWESOME!

PR #2067, opened 2022-08-29
Update non-controversial citation information

cthoyt:
This PR is a slimmed down version of #2063 based on suggestion from https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2063#pullrequestreview-1088320114 that only contains non-controversial changes.

## Depends On

- [x] #2065


matentzn (approved):
As a senior TWG member, in accordance with the recently merged amendments to the metadata change SOP #2065, I judge all changes in this PR lossless. 

Thank you @cthoyt for driving this forward!

Issue #2043, opened 2022-08-23
Documentation for Chairing OFOC call


shawntanzk:
We need documentation on chairing calls, @matentzn wrote the following: 

# How to prepare an OBO Ops call :
## Before the call:
1. Prepare a stub in the OBO Ops running agenda with all the regular items (you will find instructions in the running minutes itself). This include items like discussing new ontology requests, adding new members to the mailing list if any, etc.
1. Looking at OFOC call at OBO foundry repo. Pick one or two open issues you deem important and put them towards the end of the agenda. It is good to have these there. (This time, the main issue of importance is contuing our reviewer SOP).
1. Send an email to obo-operations (not obo-discuss!) to say: “I am doing this weeks meeting, here is a provisional agenda, feel free to add stuff!”
## During the call:
1. Wait for 8-10 people to arrive. Start the latest 4-5 minutes after the hour regardless of how many people are present.
1. Share your screen, say “hello y’all” and drive through the agenda. (You dont have to show your face, but some people do. This is up to you).
1. The important thing is to keep on track. If a discussion takes too long (in the sense of, you wont finish the rest of the agenda), you politely summarise the discussion and say: “After the meeting I will make an issue or a comment on the issue with the next steps”. We dont have to finish everything in a given sitting. If you are unsure: “Unfortunately we need to move on now, can I ask X (primary participant in the discussion) to summarise what needs to happen next, so we can move to the next item of the agenda?“.
1. Try to add a few raw notes to the agenda, capturing decisions etc. Do not hesitate to have people waiting for you taking notes: if we dont take notes with decisions, the discussion will resurface which means, more time will be lost.
1. Close exactly on the hour or earlier, no later, interrupt if necessary. Otherwise people will be forced to hop off in an impolite manner, which is not nice. Say thank you to the participants.
## After the call:
1. Sift through the agenda, streamline/format it a bit if it helps with clarity, make sure all issues are made on GitHub, all comments are made as discussed.

nataled:
The pull request is https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2074.

@cstoeckert note that I had to change the section headers from using three hash marks to using two so that the subsection headers displayed a reasonable size. You might have to make the corresponding change to your branch for consistency.

nataled:
[...]

PR #2074, opened 2022-08-30
Update SOP doc with Ops Chair info

nataled:
Addresses https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2043

### DIFF
--- a/docs/SOP.md
+++ b/docs/SOP.md
@@ -6,14 +6,15 @@ title: Standard Operating Procedures
 
 This document contains standard operating procedures (SOPs) for the OBO Foundry Operations Committee, and is intended for internal use only.
 
-## SOPs
+# SOPs
 
 - [New Ontology Requests](#NOR)
 - [Changing ontology metadata in the registry](#META)
+- [Chairing an OBO Operations Committee meeting call](#OPS_CHAIR)
 
 <a name="NOR"></a>
 
-### New Ontology Requests (NOR)
+## New Ontology Requests (NOR)
 
 1. When receiving a new ontology request (NOR), the OBO dashboard administrator should thank the submitter for their submission.
 1. The OBO dashboard administrator adds the new submission to the NOR dashboard, which is deployed at https://obofoundry.github.io/obo-nor.github.io/
@@ -24,7 +25,7 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry 
 
 <a name="META"></a>
 
-### Changing ontology metadata in the registry
+## Changing ontology metadata in the registry
 
 In general, the metadata record of an ontology in the OBO Foundry metadata registry ([example](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/ontology/go.md)) is managed and curated by the ontology team that is responsible for the respective ontology. However, as an open data organisation, the OBO Foundry does accept proposals by any member of the community to change this metadata. Such change proposal can include fixing typos, adding a tag, adding a publication where it was missed. The following SOP exists to ensure that these changes are not performed without the knowledge of the responsible ontology team.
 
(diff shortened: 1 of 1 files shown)


docs/SOP.md
```diff
@@ -39,3 +40,48 @@ In general, the metadata record of an ontology in the OBO Foundry metadata regis
...
+
+     1. Check for [pending members for obo-discuss](https://groups.google.com/g/obo-discuss/pending-members) 
+     2. Get volunteers to sign up to lead upcoming meetings (if needed)
+     3. Review [new ontology requests](https://github.com/OBOFoundry/OBOFoundry.github.io/labels/new%20ontology)
+     4. Report from Editorial Working Group (EWG) (Darren)
+     5. Report from Technical Working Group (TWG) (Nico)
+     6. Review additional [open issues](https://github.com/OBOFoundry/OBOFoundry.github.io/labels/attn%3A%20OFOC%20call) and [pull requests](https://github.com/OBOFoundry/OBOFoundry.github.io/pulls?q=is%3Apr+is%3Aopen+label%3A%22attn%3A+OFOC+call%22) that are labeled “attn: OFOC call”
+
+2. Check the issues labeled [“attn:OFOC call”](https://github.com/OBOFoundry/OBOFoundry.github.io/labels/attn%3A%20OFOC%20call) found at the OBO foundry github repository. Pick one or two open issues you deem important and put them towards the end of the stub agenda, after the Working Group reports and before the review of additional open issues.
```
matentzn, line 58: Quotes look like they are wrongly formatted. Should be `"`, not `“`.

docs/SOP.md
```diff
@@ -6,14 +6,15 @@ title: Standard Operating Procedures
 
 This document contains standard operating procedures (SOPs) for the OBO Foundry Operations Committee, and is intended for internal use only.
 
-## SOPs
+# SOPs
```
matentzn, line 9: We just made this change the other way around, because `#` (headline 1) is  reserved to the "site title" of a page. So we should keep this as it was.

docs/SOP.md
matentzn: Happens in a few other places

matentzn (changes requested):
Awesome, thank you for doing that! I made a few suggestions!

docs/SOP.md
```diff
@@ -39,3 +40,48 @@ In general, the metadata record of an ontology in the OBO Foundry metadata regis
...
+"I am chairing this week’s OFOC meeting.<br>
+
+Please review
+[new ontology requests](https://github.com/OBOFoundry/OBOFoundry.github.io/labels/new%20ontology),
+[open pull requests](https://github.com/OBOFoundry/OBOFoundry.github.io/pulls), and
+[recent tracker items](https://github.com/OBOFoundry/OBOFoundry.github.io/issues?q=is%3Aopen).<br>
+
+Please also add agenda items to https://docs.google.com/document/d/1aka4i6R89i04IYPS7CyzItQPOyb3IgtW4m75G475qcc/edit.<br>
+
+Zoom link: https://lbnl.zoom.us/j/95769353231?pwd=Y3ZtVTVVZnZEQW1GSkxZK0NwdVIydz09"
```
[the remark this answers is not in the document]
nataled: In that case we should probably just leave the entire text of the email in the agenda document?

docs/SOP.md
nataled: We edited the original after copying to a Google doc to facilitate group editing. These are a holdover of that process. Will correct them.

docs/SOP.md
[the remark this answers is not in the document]
nataled: Nevermind, I used a different tactic.

nataled (commented):
File revised.

nataled:
Merged, but it looks like the headers have reverted to what I had them as and not to what you wanted them changed tow. Was that intended?

Also, the agenda URL is missing, replaced by a tag of some sort, so it no longer works.

matentzn:
Yeah, I removed to URL - just to not create overhead to remember to update the SOP whenever we change URL..

The rest of the nesting is correct. Thanks!

nataled:
By removing the URL, the email will be rather useless, as the recipient won't be able to find the Zoom link either. It will be far easier to revise the SOP if the agenda link changes than it will be for each chair to grab links from multiple places to compose the email.

PR #2075, opened 2022-08-31
Update SOP.md

matentzn:


nataled (commented):
Can't say I'm in favor of this. We either have a fully fleshed out set of instructions here, or just revert back to having the full set in the agenda, which will be much easier to deal with. I suggest keeping the URL in the SOP, and do as we've done in the past regarding the rolling agenda: when it changes, we add a line pointing to the current set. Then, even if the URL in the SOP is old, meeting attendees will still get to where they need to go, and there will be a reminder of sorts to change it.

matentzn:
Ok! Check again, I trust your judgment on this.

nataled (approved):

PR #2077, opened 2022-09-01
add Slack community to the "Participate" list near the top of the OBO…

### LINKED ISSUE #2046: add Slack community to the "Participate" list at the top of the OBO Foundry home page
The OBO Community Slack channel is listed on the Participate drop down menu in the header. That's great but it would also be good to highlight this really great community resource as one of the bullet items under Participate on the Home Page. Currently that list has:
    Join the OBO mailing list
    OBO Foundry Operations and Working Groups
    Submit bug reports or suggestions for improvement via GitHub
    Submit your ontology to be considered for inclusion in the OBO Foundry
which is different from the Participate drop down menu.
I suggest putting it after the mail list. Or If we want to save room/ not add more white space, we could add it to the mail list as in "Join the OBO mailing list and the OBO Community Slack channel" with different links for both of course.

nlharris:
… Foundry home page

fixes #2046

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,9 +22,8 @@ title: The OBO Foundry
   <h4>Participate</h4>
   <ul>
     <li>
-      <a href="https://groups.google.com/forum/#!forum/obo-discuss"
-        >Join the OBO mailing list</a
-      >
+      Join the <a href="https://groups.google.com/forum/#!forum/obo-discuss">OBO mailing list</a>
+      and the <a href="https://obo-communitygroup.slack.com/archives/C01DP18L5GW">OBO Community Slack workspace</a>
     </li>
     <li>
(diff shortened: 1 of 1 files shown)


cstoeckert (commented):
Thanks @nlharris this is exactly what I asked for but I'm now wondering whether we should be pointing people specifically at the community channel which is important but kind of limited or more generally to the obo-community group to showcase all the exciting available channels for participation. What do you think?

nlharris:
Should we link to https://obo-communitygroup.slack.com/ instead?

cstoeckert (approved):
@nlharris Thanks for making the change!

nlharris:
I can't merge; someone else will need to.

matentzn (approved):

PR #2081, opened 2022-09-05
Add missing products to several ontologies

cthoyt:
I wrote a script that tries to fill these in by re-writing the file extensions then running a HEAD request with redirects to see if the file exists. I then went through and **manually checked** that actual files exist for all of the changes proposed here.

Rationale: I want JSON exports of everything, and some were available and I was wasting tons of time converting OWL/OBO with ROBOT - much better to have it annotated directly so these can be automatically consumed.


matentzn:
I think this is a good change, but not sure yet how I will go about getting permission for this.. I will ask at the call later today.

matentzn:
As far as I can judge this issue will be a bit longer running. @cthoyt if you want to move forward with this can you tag all contacts involved in checklist format so we can elicit permissions from them?

cthoyt:
> did your manual check also check the version? Do we know for sure that this PR would not be linking to a one-off stale derived form of the ontology? I suspect you did check this, but I'm just checking you checked to be sure :-)

I didn't check this, but if one of the ones that I added are stale, it's because the content was already stale, and this should be addressed in a different way not in scope of this PR.

> As far as I can judge this issue will be a bit longer running. @cthoyt if you want to move forward with this can you tag all contacts involved in checklist format so we can elicit permissions from them?

No, I don't want to move forwards on this. I have run out of motivation and patience to tag people from ontologies and wait on them. I made the corresponding updates in Bioregistry at https://github.com/biopragmatics/bioregistry/commit/0351498644b45ef7d7e34b61aa4cd0ad978e3389 and will leave you all to figure out a way to push the ontology community forwards without me.

Issue #2084, opened 2022-09-08
Add link to SOP


nlharris:
- [x] Link to [SOP](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/SOP.md) from some menu

- [x] Replace “instructions to chairs” in the meeting notes doc with a link to the SOP (so we don’t have to do updates in two places)

nlharris:
I have a concern about info that is exposed in the SOP; I wrote a message in Slack about it.

matentzn:
This link should be closed to the public! Isn't it?

nlharris:
It was not closed until @balhoff made it so a few hours ago.

nlharris:
Regarding the second thing ("Replace “instructions to chairs” in the meeting notes doc with a link to the SOP (so we don’t have to do updates in two places)"), what I'd like to do is create a fresh obo ops meeting doc that can be shared publicly and make this change in that new doc. I would then change the OBO Ops calendar event to point to the new doc. (Normally I would prefer to keep the URL of the meeting notes doc unchanged and move out the old notes to a new doc, but in this case, that would enable the general public to see the removed notes via doc history.)

Are we ready for this? @matentzn @bpeters42 @cmungall

matentzn:
Yes, the current minutes have been overflowing anyways. New minutes I think is fine.

nlharris:
Chris wants it to be in a new drive, and we decided not to do that yet. I did move the 2021 minutes into a new doc, and left the 2022 minutes in the current doc for now.

PR #2087, opened 2022-09-12
add link to SOP

nlharris:
also added a summary (from the SOP page itself) but if that seems like a bad idea (because it could get stale if the SOP page is changed), then i'm happy to leave that out.

Addresses #2084

### DIFF
--- a/docs/OperationsCommittee.md
+++ b/docs/OperationsCommittee.md
@@ -17,12 +17,6 @@ See a description of our activities in our [mission statement](MissionStatement.
 
 ## OBO Operations Committee SOP
 [Standard operating procedures (SOPs)](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/SOP.md) for the OBO Foundry Operations Committee can be found in our GitHub repo.
-That document describes:
-- New Ontology Requests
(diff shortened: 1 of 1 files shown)


docs/OperationsCommittee.md
```diff
@@ -15,6 +15,15 @@ title: Operations Committee
...
 See a description of our activities in our [mission statement](MissionStatement.html).
 
+## OBO Operations Committee SOP
+[Standard operating procedures (SOPs)](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/docs/SOP.md) for the OBO Foundry Operations Committee can be found in our GitHub repo.
+That document describes:
+- New Ontology Requests
+- Reviewing Ontologies for OBO Membership
+- Ontology acceptance email
+- Changing ontology metadata in the registry
+- Chairing an OBO Operations Committee meeting call
```
matentzn, line 25: We would have to keep this list up to date whenever SOP.md changes.. Also would be great to add SOP to menu somewhere.

nlharris: I see it now in the About menu -- did you just add it?

matentzn (approved):

PR #1958, opened 2022-06-01
create vbo.md

sabrinatoro:
This is related to this issue https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1883

PR #2091, opened 2022-09-16
Add automated check for OBO Library PURL configuration

cthoyt:
As mentioned in https://github.com/OBOFoundry/OBOFoundry.github.io/pull/1958#issuecomment-1249287870, all new ontologies should have a working configuration on the OBO Library PURL service at https://github.com/OBOFoundry/purl.obolibrary.org before merging any new ontology metadata files.

This PR adds this check, which formats the prefix for the ontology into the appropriate URL and checks if it can be retrieved with a HTTP 200.

## Blocked by

- [x] @sabrinatoro you should create an appropriate configuration, noted in https://github.com/OBOFoundry/purl.obolibrary.org/issues/869

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,6 +8,7 @@ version = 0.0.1-dev
 [options]
 install_requires =
     pyyaml
+    requests
 
 # Random options
 zip_safe = false
--- a/tests/test_integrity.py
+++ b/tests/test_integrity.py
@@ -6,6 +6,7 @@ from io import StringIO
 from pathlib import Path
 from typing import Set
 
+import requests
 import yaml
 
 from obofoundry.standardize_metadata import ModifiedDumper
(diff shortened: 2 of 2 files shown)


matentzn:
I think this makes sense, but for some reason I remember that there is a dependency the other way around (purl-->metadata). @jamesaoverton can you confirm/unconfirm my memory? I somehow remember you telling me 2 years ago that we first merge metadata than purl configs, but I don't remember why or even if you said that..

jamesaoverton:
Since the discussion takes place here, I have been waiting for the metadata file to be merged here before I merged the PURL config. As long as the process is clear, either way works.

cthoyt:
An alternative is to add a required metadata field that's the pull request number for adding the metadata configuration to the other repo. Up to you

jamesaoverton (approved):
I'm happy with this. From now on the PURL PR for a new ontology will have to be merged before the registry entry PR, which makes sense.

matentzn:
@cthoyt PR is currently failing because of the python requirements that are missing. 

Also in the error log I am noticing something entirely unrelated:

```
INFO	NOMEN funded_by: 'url' is a required property
INFO	UBERON funded_by: 'url' is a required property
INFO	ECO funded_by: 'url' is a required property
```

Shouldnt this fail with all your work you put into standardising funding information?

matentzn:
Thanks a ton :)

PR #2097, opened 2022-09-23
Use prepend when importing `all.js`

### LINKED ISSUE #2088: Bug: menus don't work from some pages
I'm not sure if this is a new bug, but I just noticed it.

From the home page, all the menus in the top bar work, but if you go to some pages, the pulldown menus stop working. ("Resources", which is just a pointer to a single URL, still works.)

For example, from https://obofoundry.org/about-OBO-Foundry.html the menus work, but from https://obofoundry.org/faq/index.html they don't (all top-level menu items just go to https://obofoundry.org/faq/index.html#, and the pulldown menus don't pull down). Other pages where the menu is broken include https://obofoundry.org/principles/fp-000-summary.html# and https://obofoundry.org/docs/SOP.html. Another page where the menu works is https://obofoundry.org/resources

erik-whiting:
I am fairly certain this will fix #2088


erik-whiting:
Hey @nlharris and @matentzn I think this will fix the issue with the menus not dropping down

matentzn (approved):
Let's just try - can't get worse

PR #2099, opened 2022-09-26
Try to fix dropdowns again

erik-whiting:
Hey @matentzn here's another shot at it. I have a better feeling about this one, but I don't know.

Basically, I just moved `all.js` to `assets/js` which seems more like what Jekyll is expecting. I also used a slightly different syntax for linking the JavaScript file. I'm more hopeful about this approach because I am referencing the file much more explicitly. But I've been wrong before 😬 I wanted to get this change to you before ICBO starts.

Once you merge this, I will immediately make a revert commit so that you can roll this back if it doesn't work or if it breaks something.

Thanks! 🤞


matentzn (approved):
Let's try it ;)

PR #2102, opened 2022-09-26
Update ohd.md

wdduncan:
Add citation for the OHD.

### DIFF
--- a/ontology/ohd.md
+++ b/ontology/ohd.md
@@ -12,6 +12,7 @@ contact:
   orcid: 0000-0001-9625-1899
 description: The Oral Health and Disease Ontology was created, initially, to represent the content of dental practice health records.
 domain: health
+cite: Structuring, reuse and analysis of electronic dental data using the Oral Health and Disease Ontology
 homepage: https://purl.obolibrary.org/obo/ohd/home
 license:
   label: CC BY 4.0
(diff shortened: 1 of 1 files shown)


wdduncan:
I don't understand why this failed :(

cthoyt:
Here's an example of how this should look in the diff:

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/68f02e3f9ad8d7ef302067aabd9c07af03f80620/ontology/bto.md?plain=1#L26-L28

Further, you shouldn't add redundant information of the citation in the free text part of the markdown file.

ontology/ohd.md
```diff
@@ -12,7 +12,9 @@ contact:
   orcid: 0000-0001-9625-1899
 description: The Oral Health and Disease Ontology was created, initially, to represent the content of dental practice health records.
 domain: health
-cite: Structuring, reuse and analysis of electronic dental data using the Oral Health and Disease Ontology
+publications: 
+  - id: https://pubmed.ncbi.nlm.nih.gov/32819435/
```
cthoyt: Trailing slashes are not standard 
```suggestion
  - id: https://pubmed.ncbi.nlm.nih.gov/32819435
```

cthoyt:
@wdduncan are you comfortable with clicking the "details" button on the failing tests and reading through the output? Maybe there are ways we can make that more helpful

wdduncan:
@cthoyt Sometimes I can make sense of the details. Sometimes I can't. In this case, I can't.

wdduncan:
@matentzn In my browser both URLs work for the paper. Did the first one/original one not work for you?

wdduncan:
@cthoyt @matentzn Do you know if this error is referring to the `title` of the paper?
```
AssertionError: 'layout: ontology_detail\nid: ohd\ntitle:[951 chars]tive' != "layout: ontology_detail\nid: ohd\ntitle:[953 chars]tive"
E               Diff is 1208 characters long. Set self.maxDiff to None to see it. : 
```

matentzn (approved):

PR #2109, opened 2022-09-28
Update gaz.md

lschriml:
adding domain to GAZ


matentzn (approved):
Thank you @lschriml !

Issue #2107, opened 2022-09-28
Ontologies without domain not displayed on homepage


allenbaron:
https://obofoundry.org/ does not display ontologies without domains even when all checkboxes are unchecked.
<img width="1216" alt="image" src="https://user-images.githubusercontent.com/21225409/192851217-6e6b45ab-d6eb-4c0f-ab96-cfdeacf29ddd.png">

Examples of missing ontologies by prefix: aao, fix, gaz

nlharris:
Hey @allenbaron thanks for reporting this (and thanks for giving specific examples of the issue -- always helpful!).

erik-whiting:
Gave this one a try: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2114

cthoyt:
Alternate solution to making a programmatic solution: require _all_ ontologies are annotated with a domain, do the annotation, and add a unit test.

cthoyt:
All ontologies are required to have domains as of #2115, so this can be closed.

PR #2114, opened 2022-09-30
Don't filter ontologies with no domain when not grouping by domain

erik-whiting:
With these changes, the homepage will show ontologies that don't have a domain if the "Group by Domain" checkbox is unchecked. I believe this is what the reporter of #2107 wants.

Something to think about: Do you want to put a "No Domain" grouping when looking at ontologies grouped by domain? This will be a little bit more work but if it's something that would be helpful to your users, I'd be happy to take a shot at it.

### DIFF
--- a/_includes/js/custom.js
+++ b/_includes/js/custom.js
@@ -404,10 +404,22 @@ jQuery(document).ready(function() {
      */
     function apply_all_filters(data) {
         let selectedDomain = $("#dd-domains").children("option:selected").val();
-        let res = data["ontologies"].filter(x => x["domain"] !== undefined);
-        let dt = res.filter(x => x["domain"].includes(selectedDomain));
-        let dt2 = Search($("#searchVal"), dt);
-        applyFilters(dt2)
+        if (selectedDomain) {
+            /**
+             * Filter out ontologies with no `domain` attribute and apply
+             * the rest of the user-supplied filters to the data.
+             */
+            let res = data["ontologies"].filter(x => x["domain"] !== undefined);
+            let dt = res.filter(x => x["domain"].includes(selectedDomain));
+            let dt2 = Search($("#searchVal"), dt);
+            applyFilters(dt2)
+        } else {
+            /**
+             * If no filters are supplied, just send back the raw data,
+             * including ontologies that don't have a `domain` attribute.
+             */
+            applyFilters(data["ontologies"]);
+        }
     }
 // obtain json data using fetch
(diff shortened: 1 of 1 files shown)


cthoyt:
@erik-whiting sorry for the double post, but here I think a curation solution would be better than doing more programming

matentzn:
@cthoyt - I see it a bit differently. This PR solves a problem we currently have with the code - your are talking about "using the opportunity of this bug to enhance the metadata".. 

Here is my reasoning:

1. @erik-whiting took the time to investigate the code to figure out how to display ontologies in the "no domain" view. It's important that we make such contributions if they correctly address the issue easy to do and not add additional churn by saying: it would be better to solve this other problem (incomplete metadata) instead. @erik-whiting is unlikely to solve the metadata + QC issue now (correct me if I am wrong @erik-whiting).
2. We have no dedicated support for fixing the metadata now, and it can be fixed entirely independently of this PR.
3. Some OBO ontologies are not shown on the frontpage which is bad. 

My suggestion is to merge this PR (Thank you @erik-whiting, again!) and if adding domains for all these ontologies is a concern, make a separate issue about that. Let me know what you think!

cthoyt:
@matentzn okay yes you have a good point. Sorry, I should have started by saying thank you @erik-whiting for figuring this out! Let's move forwards the way you suggested (though I would ask kindly that Erik put LOTS of code comments explaining what this javascript is doing, and why, before accepting this PR)

matentzn (changes requested):
Excellent @erik-whiting thank you! Can you add some inline comments just above your changes in the code what their part of the code is doing and why it had to change? That would be fabulous! Thanks @cthoyt for the review :D

erik-whiting:
> @erik-whiting took the time to investigate the code to figure out how to display ontologies in the "no domain" view

Thanks, but just so you know, you don't have to accept my PR just because I made one. If there's a more preferable solution, I won't be offended if this PR is rejected.

> @erik-whiting is unlikely to solve the metadata + QC issue now

This is true, I wouldn't know where to start

> Can you add some inline comments just above your changes

Yes, I can do this after work today

matentzn:
> Thanks, but just so you know, you don't have to accept my PR just because I made one. 

Hahha don't worry. It will be accepted because it solves a problem :) we would never accept a PR if we didn't believe that!

matentzn (approved):

matentzn:
Thank you @erik-whiting

PR #2115, opened 2022-09-30
Curate domains for all remaining ontologies

cthoyt:
It took about 25 minutes to go through and give a good shot at annotating all of the remaining ontologies that didn't have one. I just read through the descriptions, some of them that had replacement annotations were even easier since I could just grab the domain from the updated one.

By definition all of the remaining ontologies are orphaned, deprecated, or obsolete, so there is no bureaucratic need to ask anyone if we can update these metadata, but any suggestions are welcome.

### DIFF
--- a/ontology/ev.md
+++ b/ontology/ev.md
@@ -5,6 +5,7 @@ title: eVOC (Expressed Sequence Annotation for Humans)
 contact:
   email: evoc@sanbi.ac.za
   label: eVOC mailing list
+domain: chemistry and biochemistry
 homepage: http://www.evocontology.org/
 is_obsolete: true
 activity_status: inactive
--- a/ontology/ev.md
+++ b/ontology/ev.md
@@ -5,7 +5,7 @@ title: eVOC (Expressed Sequence Annotation for Humans)
 contact:
   email: evoc@sanbi.ac.za
   label: eVOC mailing list
-domain: chemistry and biochemistry
+domain: anatomy and development
 homepage: http://www.evocontology.org/
 is_obsolete: true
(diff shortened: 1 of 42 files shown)


ontology/ev.md
```diff
@@ -5,6 +5,7 @@ title: eVOC (Expressed Sequence Annotation for Humans)
 contact:
   email: evoc@sanbi.ac.za
   label: eVOC mailing list
+domain: chemistry and biochemistry
```
cmungall, line 8: Should be anatomy and development

ontology/ogi.md
```diff
@@ -8,6 +8,7 @@ contact:
   label: Asiyah Yu Lin
   orcid: 0000-0002-5379-5359
 description: An ontology that formalizes the genomic element by defining an upper class genetic interval
+domain: investigations
```
cmungall, line 11: Should be the same as SO

cmungall (approved):

ontology/ev.md
cthoyt:
```suggestion
domain: anatomy and development
```

ontology/ogi.md
cthoyt:
```suggestion
domain: chemistry and biochemistry
```

Should OGSF also be updated then?

cthoyt:
I made the changes suggested by Chris. this should be ready to merge

matentzn (approved):
Excellent detective work and great effort. I have reviewed every single ontology, and while some of them are a bit ambiguous, I also couldn't find a better domain to fit them in.

PR #2117, opened 2022-10-01
Structure membership information

cthoyt:
This pull request standardizes the information about membership into yaml files to make it more simple to maintain. It also adds the requirement that the ORCID identifier of each person is annotated, I went through and grabbed all of them.

This PR also makes minor changes to the layout of various WG pages for consistency (no information loss).

## To Do (but not necessary for merging)

Get ORCID identifiers for the following alumni (help wanted):

- [ ] Eric Douglass (@dougli1sqrd is this you?)
- [ ] Gareth Owen (this person? https://www.researchgate.net/scientific-contributions/Gareth-Owen-44854599 / https://mobile.twitter.com/GarethOwenLloyd?)
- [ ] Carlo Tornial (this person? https://scholar.google.com/citations?user=pYexO7AAAAAJ&hl=en?)

## Screenshots

Screen Shot 2022-10-01 at 13 46 43

Screen Shot 2022-10-01 at 14 24 18


_data/alumni.yml
```diff
@@ -0,0 +1,17 @@
+members:
+- name: Michael Ashburner
```
cthoyt, line 2: ideally we could also get departure dates for alumni, but I don't think that's feasible @nlharris

matentzn (commented):
@cthoyt also always - this is amazing. Having the memberships in this much more formal way is a great idea: it gives us a bit more framework to communicate expectations to OBO Ops members. 

The only bigger thing that I disagree with is that the OBO Ops member list is now replaced by the other three sublists. I think there should be a single table with all OBO ops members, easily accessible, and, indeed, with photos even (I collected them all for ICBO, could upload them).

A smaller thing is the redundancy of having to curate the same affiliation and orcid twice if people are part of multiple subgroups. Also, we are about to add other groups, such as governance.

Do address all of the above at once, I recommend to have a single yaml file with all OBO Ops current members, and a list item: `working_groups` or some such which lists which working groups they are part of. The individual pages can then use this to decide if a person should be listed on page (liquid template conditional).

cthoyt:
Done

Screen Shot 2022-10-01 at 16 33 17

_data/operations.yml
```diff
@@ -0,0 +1,172 @@
...
+  affiliation: Lawrence Berkeley National Laboratory, Berkeley, CA, USA
+  orcid: "0000-0001-8244-1536"
+  groups:
+  - technical
+- name: Alexander Diehl
+  affiliation: University at Buffalo, Buffalo, NY, USA
+  orcid: 0000-0001-9990-8331
+  groups:
+  - technical
+
```
matentzn, line 68:
```suggestion
```

matentzn (commented):
This is absolutely fantastic. How exactly are these pages updated? Does this somehow happen automatically when you, say change, the opertions.yml file?

cthoyt:
> This is absolutely fantastic. How exactly are these pages updated? Does this somehow happen automatically when you, say change, the opertions.yml file?

Yes, data inside the `_data` directory is used during building the static site. So anytime you commit a change to the `operations.yml` file, it will trigger a rebuild of the static site, and automatically reflect the changes there.

matentzn (approved):
I am very happy to take this for a ride, but your own QC is failing :D

cthoyt:
gotta eat your own dog food sometimes

cthoyt:
Unfortunately I got excited about using high-power typing and imported something that wasn't available on py36. But it's crazy to run tests on that, so I upgraded the testing pipeline to use py39 and it's all good now.

matentzn (approved):

PR #2126, opened 2022-10-02
Map operations committee metadata to wikidata and github

### LINKED ISSUE #2125: Add GitHub handles to new OBO Metadata Pages
@cthoyt gifted a great new overview page for OBO Ops members https://obofoundry.org/docs/Membership.html

I find myself looking for Ops GitHub handles all the time, and I would like to propose to add them to 

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/_data/operations.yml

I tagged this as good first issue, please do _not do this_ if your name is Charlie or you have completed more than 2 issues on this repo here in your life!

cthoyt:
Closes #2125

This PR adds a script that queries the SPARQL endpoint of Wikidata for ORCID identifiers of operations committee to get their wikidata identifier and GitHub. Originally, this was optional, but there were only 3 or 4 members who didn't have the data in their wikidata already so I just added it then made it required.

### DIFF
--- a/src/obofoundry/utils.py
+++ b/src/obofoundry/utils.py
@@ -4,14 +4,11 @@ from pathlib import Path
 
 import yaml
 
+from obofoundry.constants import ONTOLOGY_DIRECTORY
+
 __all__ = [
     "get_data",
-    "ONTOLOGY_DIRECTORY",
 ]
-
-HERE = Path(__file__).parent.resolve()
-ROOT = HERE.parent.parent.resolve()
-ONTOLOGY_DIRECTORY = ROOT.joinpath("ontology").resolve()
 
 
 def get_data():
--- a/Makefile
+++ b/Makefile
@@ -150,6 +150,7 @@ tox:
 
 # Use SPARQL queries to Wikidata to enrich the
 # OBO operations committee metadata file
+.PHONY: update-operations-metadata
 update-operations-metadata:
 	python -m obofoundry update-operations-metadata
 
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,15 @@ prettify: $(ONTS)
 tox:
 	tox -e py
 
+##########################
+## Metadata Maintenance ##
+##########################
+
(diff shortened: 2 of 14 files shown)


src/obofoundry/utils.py
```diff
@@ -4,15 +4,12 @@
 
 import yaml
 
+from obofoundry.constants import ONTOLOGY_DIRECTORY
+
 __all__ = [
     "get_data",
-    "ONTOLOGY_DIRECTORY",
```
matentzn, line 9: What is the significance of this specific removal here?

matentzn (changes requested):
Wow this is so cool! Tiny change request:

- [ ] Add PHONY make target to `Makefile` to run the `update_operations_metadata.py. The Makefile also serves as a sort of documentation of all the different scripts that can be run in OBO, else we will lose track of all of these and wont ever run them again!

src/obofoundry/utils.py
cthoyt: made this a constant that could be reused, therefore it doesn't need to get exported from `utils.py`

cthoyt:
> Add PHONY make target to Makefile to run the `update_operations_metadata.py. 

Will do. What does `PHONY` mean, btw?

matentzn:
If you add this:

```
.PHONY: refresh-ofoc-wikidata
refresh-ofoc-wikidata:
    python script.py
```

Then you tell `make` that your goal does not result in the creation of a file. The main consequence of this is that the goal is ignored during make caching (i.e. it is always going to be run whether or not its dependencies were updated).

src/obofoundry/utils.py
matentzn: cool

matentzn (approved):
Excellent, thank you for this!

matentzn:
Shall I merge?

cthoyt:
Okay thanks. Done in 7b02664 and 6f2c33c

PR #2128, opened 2022-10-02
Simplify maintenance of publication list

cthoyt:
Since the `publications.md` file is turned into HTML during static site generation, this PR switches out the auto-generated functionality to use Jekyll's "liquid" templating instead. This creates a lot less code and removes the need for maintenance.

@jamesaoverton this removes some code that you wrote, so I just wanted to check that this is okay with you and that there aren't some other reasons it was done this way that I was not aware of.

## Screenshot

(things look exactly the same with this update)

Screen Shot 2022-10-02 at 15 03 56


jamesaoverton (approved):
I have not recollection of writing that code, and no attachment to it. :+1:

matentzn (approved):
Excellent, thanks a ton. All looks right!

Issue #2130, opened 2022-10-03
Table search on homepage no longer working


allenbaron:
Input to the "Search Table" box on https://obofoundry.org/ no longer works.
This bug is not fixed by unchecking the "Group By Domain", "Hide Inactive", or "Hide Obsolete" boxes.

<img width="600" alt="Screen Shot 2022-10-03 at 11 04 45 AM" src="https://user-images.githubusercontent.com/21225409/193611114-8bc88356-c1fe-4ec0-b0df-0b75c6311a88.png">

Could this be related to the fix for #2107?

matentzn:
@erik-whiting would you be able to look at that?

erik-whiting:
@matentzn yes, I'll check tonight after work and school

nlharris:
Wow, you have work AND school and you're still helping us? You're amazing, @erik-whiting! Thank you so much!

nlharris:
And thank you, @allenbaron, for reporting this bug!

nlharris:
In case it's helpful, I found that the search isn't totally broken. If the ontology you're searching for has a domain and you restrict to that domain, it shows up.
<img width="243" alt="Screen Shot 2022-10-03 at 10 22 38 AM" src="https://user-images.githubusercontent.com/348704/193639950-7e42601e-5fe6-495b-830d-10bfe283702e.png">
<img width="309" alt="Screen Shot 2022-10-03 at 10 22 00 AM" src="https://user-images.githubusercontent.com/348704/193639951-ccc55e86-12f2-4a24-9d09-c84da436597f.png">

erik-whiting:
this makes me think it's related to my recent PR here: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2114
you can probably revert that commit to stop the bleeding if you can't wait for me to get to it

erik-whiting:
This PR (https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2133) reverts the changes I made to make domainless ontologies show up on the homepage. However, this PR (https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2115) adds a domain to tall the ontologies that didn't have one. I think the best thing to do here is just revert the breaking commit, since #2107 is solved by #2115

erik-whiting:
can this be closed or are there still problems?

nlharris:
Search seems to be working again; ok with me to close.

PR #2133, opened 2022-10-04
Revert "Don't filter ontologies with no domain when not grouping by domain"

erik-whiting:
This reverts commit 78ff5ee88bb565dbd2018a4dfda18b2f32a92044

This should fix the problem with searching documented in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2130


matentzn (approved):
@erik-whiting  thank you! :)

PR #2135, opened 2022-10-04
Add note about OFOC members stepping down

jamesaoverton:
As discussed on the OFOC call today, we need an SOP for members stepping down and being removed from the various mailing lists and resources. Since the onboarding instructions are in the "OBO Operations Onboarding" Google Doc, I added some "offboarding" instructions there as well.


cthoyt:
Please also include in the SOP to mark them appropriately in the alumni.yml as well as add a "departed" date

jamesaoverton:
Thanks @cthoyt, I added that to the Google Doc.

matentzn (approved):

PR #2150, opened 2022-10-07
Update display of alumni

### LINKED ISSUE #2134: Add remaining ORCID identifiers for Alumni
The following three alumni of the OBO Operations committee are missing the ORCID identifiers in https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/_data/alumni.yml. It would be great to add them!

- Eric Douglass (https://orcid.org/0000-0001-6946-1807, thanks @balhoff for finding this)
- Gareth Owen (https://orcid.org/0000-0001-7265-5776, thanks @balhoff for finding this)
- Carlo Tornial (https://orcid.org/0000-0002-3734-1859, thanks @wdduncan for finding this)

## ~Step 1: Internet Sleuthing and Reaching Out~ Done!

It's an internet detective mission to figure out who people are, especially if you haven't been an insider for the last 10+ years. The job here is to figure out who these people are by searching around google scholar, ORCID, by asking people who have been around for a long time to get email addresses, etc. (usually the best way is to [join the OBO Foundry slack workspace](https://join.slack.com/t/obo-communitygroup/shared_invite/zt-1ghuih22a-Sxo4bpSd167ik~q8UB9MFw), and ask someone like Chris Mungall via DM)

If these people don't have ORCIDs, then it would be good to ask them directly to make one.

cthoyt:
This PR applies similar metadata standards to the alumni list and improves its display. It also closes #2134 in the process (sorry, potential first time contributors) based on the suggestions from @balhoff and @wdduncan (thank you!!!).

Screen Shot 2022-10-07 at 18 33 48

### DIFF
--- a/_data/alumni.yml
+++ b/_data/alumni.yml
@@ -1,35 +1,63 @@
 members:
-- name: Michael Ashburner
+- github: ghost
+  name: Michael Ashburner
+  note: Retired before having a GitHub was standard
   orcid: 0000-0002-6962-2807
-- name: Colin Batchelor
+  wikidata: Q6828277
+- github: batchelorc
+  name: Colin Batchelor
   orcid: 0000-0001-5985-7429
-- name: Mathias Brochhausen
+  wikidata: Q28923515
+- github: mbrochhausen
+  name: Mathias Brochhausen
   orcid: 0000-0003-1834-3856
-- name: Melanie Courtot
+  wikidata: Q57161944
+- github: mcourtot
   link: http://purl.org/net/mcourtot
+  name: Melanie Courtot
   orcid: 0000-0002-9551-6370
-- name: Eric Douglass
-- name: Janna Hastings
+  wikidata: Q23809262
+- github: dougli1sqrd
+  name: Eric Douglass
+  orcid: 0000-0001-6946-1807
+  wikidata: Q114403167
+- github: jannahastings
+  name: Janna Hastings
   orcid: 0000-0002-3469-4923
-- name: Simon Jupp
+  wikidata: Q27902110
+- github: simonjupp
+  name: Simon Jupp
   note: OBO Industry Liaison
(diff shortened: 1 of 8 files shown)


nicolevasilevsky (approved):
looks nice! thanks @cthoyt

cthoyt:
Ghost is the GitHub equivalent of “missing account”

cmungall:
Can we not just leave blank/null?

cthoyt:
In many places in the OBO Foundry, I've found myself arguing for the strictest possible schemata - this makes it better possible to validate data and provide the community with objective, automatically validated guidelines on how to do things. Rather than making it possible to make values optional, I find in almost all situations it's better to either:

1. Spend the time and do the detective work to find the right information until there are no blanks/nulls
2. in literally this single instance, there actually isn't a github for this person, so we will use a sentinel value
 
Further, I find the fact that @ghost isn't an obvious sentinel value like an empty string to be particularly endearing, because I am always on the lookout for people who want to skirt the curation rules and abuse a less strict schema.

matentzn:
I agree with the general sentiment @cthoyt, but curating gh handles for long gone alumni.. for that it seems excessive. The goal is not to gather as much information as possible, it is to gather the right data for the job - and requiring GH handles for Alumni seems a bit over the top - I would favour these tags to be optional. If anyone could type the string "ghost" as opposed to `null`, it seems nothing more than a fancy way to say its optional...

cthoyt:
Alright I will change this tag to optional.

cthoyt:
After removing requirement for github, the display remains unchanged:

Screen Shot 2022-10-09 at 14 30 05

@matentzn I hope we can merge this now :)

matentzn (approved):

matentzn:
Looks amazing, thank you!

PR #2160, opened 2022-10-24
Update responsibilities of Registry Metadata Steward

matentzn:

PR #2161, opened 2022-10-24
Add data integrity tests for roles.yml

cthoyt:
Similarly to the OBO operations data files, this PR adds a first gentle test for metadata in the `roles.yml` markup. This would be helpful to identify issues with YAML formatting like I suggested fixes for in https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2160#discussion_r1003530903

### DIFF
--- a/tests/test_roles.py
+++ b/tests/test_roles.py
@@ -1,0 +1,48 @@
+"""Test data integrity, beyond what's possible with the JSON schema."""
+
+import json
+import unittest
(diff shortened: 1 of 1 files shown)


cthoyt:
@matentzn it's updated. now there is no diff in `roles.yml` since the change was also included in another PR that got merged

matentzn (approved):
Excellent! Thank you!

Issue #2166, opened 2022-10-27
CSS Issues after Bootstrap 5 migration


matentzn:
- [x] links are all blue, bold and underlined - this looks a bit ugly
- [x] ordered and unordered lists should be indented a bit deeper under the headline
- [x] ids in table should not be bold/underlined and should probably be black?

cc @erik-whiting 

<img width="828" alt="image" src="https://user-images.githubusercontent.com/7070631/198368896-98b072ff-faaf-401c-8105-c5805245937b.png">

matentzn:
@erik-whiting can you send me an email? I wanted to send you a few invites and questions :)

erik-whiting:
@matentzn how do I find your email address?

erik-whiting:
also, what do you want the links to look like (besides not blue, bold, and underlined)?

matentzn:
I don't really know what's best here in terms of CSS.. Something more.. Modern? I only know that blue underlined Bold looks 90ies :) maybe black and not underlined? I don't know. Maybe you have better ideas?

cthoyt:
@matentzn I totally get some of these stylistic concerns, but I see a lot of value in just using the default Bootstrap styles. This means we don't have to add in tons of custom styles, which create massive overhead in understanding and later maintaining the site (trust me, figuring out the existing site was a huge headache).

erik-whiting:
https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2168

matentzn:
@cthoyt normally I would agree with you on these kinds of issues, but even for me, a person with very little aesthetics, these issues noted here are too ugly - be they standard or not.. I think a little bit of CSS customizability is justifiable in this case.. I am very surprised the default bootstrap styles are so... old.. :P

jamesaoverton:
In general, I agree with @cthoyt that we should save ourselves time and effort, and learn to live with the Bootstrap defaults. The one that kills me is the underlined links. Bootstrap 4 did not underline links, but they changed that in Bootstrap 5 (according to <https://stackoverflow.com/q/65480036>), and it looks so awful to me. So I guess I'm a hypocrite.

matentzn:
@erik-whiting and @cthoyt thank you for your great minimal solution to this issue!

PR #2168, opened 2022-10-28
Add a (small) style.css file

erik-whiting:
These are some of the changes requested in #2166, but I have some notes.

First, I kept the original (pre Bootstrap 5 upgrade) color of the links. So while they're not bright-1990s-hyperlink-blue, they're still blue-ish. Let me know if you want a different color base or different shade. Also, regarding changing the ontology IDs to black, I decided not to do that and keep them the same color as the links. I did this because 1: that's how it was before the Bootstrap 5 upgrade and 2: if they're black, it's not obvious that they can be clicked. If you really want those hyperlinks to be black though, just let me know and I'll make it so. Here's what part of the front page looks like:
image

Also, notice the indentation of the bulleted links just below the "Learn about ..." bit. They are not too dramatically indented but take a look at the bulleted lists on a different page:

image

Let me know if the indentation is too dramatic here.

Finally, wrt @cthoyt concern about maintenance overhead: I kept the CSS file as small and well-commented as possible. Luckily, there is no JavaScript involved in these changes and though the last block of the `styel.css` file is slightly more complicated than a basic straightforward CSS file, I don't think it's that bad and it's worth the slight upgrade in the site's appearance (in my opinion of course).

Also, for whatever it's worth, I don't mind being the primary person "responsible" for maintaining the CSS file here if complexity is a huge concern. Despite my best efforts, I've written a lot of CSS in the last five or so years and this file is actually pretty basic by web-dev standards, I'm willing to be the one to unwind any complexity it might introduce.


erik-whiting:
why are my commits like this

erik-whiting:
@matentzn @cthoyt PR for some of the CSS things, would love to hear y'all's thoughts

cthoyt:
FWIW I think that this small CSS file is fine, hopefully this doesn't get out of hand. I don't have any arguments around changing the link style, Bootstrap is probably smart enough not to let this sheet mess up its other buttons, etc.

However, I figured out a non CSS solution for the indentation in #2169 so I think that would be much preferred

_layouts/default.html
```diff
@@ -4,6 +4,7 @@
 <head>
     <title>{{ site.name }}</title>
     {%- include head.html -%}
+    <link rel="stylesheet" href="/assets/css/style.css">
```
cthoyt, line 7: why is this not in the `head.html` along with the other stylesheets?

assets/css/style.css
```diff
@@ -0,0 +1,31 @@
+/*
+Don't underline hyperlinks
+*/
+a {
```
cthoyt, line 4: Please look into a more careful solution that uses additional selectors suggested in https://stackoverflow.com/a/69299851/5775947

erik-whiting:
Update. Note the use of additional selectors as requested:
```css
a:not(.btn, .nav-link, .navbar-brand, .dropdown-item) {
  text-decoration: none;
  color: #6699cc;
}
```
This was necessary because had I just applied `not(.btn)`, the new style gets applied to every single anchor element (except for those with a `btn` class). Note, without the `not(.btn, .nav-link, .navbar-brand, .dropdown-item)`, the new style did not apply the color attribute to the navbar items because color attributes are defined in `nav-link` and `dropdown-item` classes, but it did apply the underlining-when-hovered-over because that attribute is not defined by `nav-link` or `dropdown-item`.

However, style definitions with a `not` take precedence over class definition, so when you define a color for a hyperlink that is `not(.btn)`, literally every anchor element without a `btn` class will be that color, even if its own class defines a different color. This is why `not(.btn, .nav-link, .navbar-brand, .dropdown-item)` is necessary here, you have to say everything you don't want this style to apply to.

Personally, I preferred it without the extra selectors because I was getting the feeling a concern was maintenance overhead and these extra `not` selectors seem like more to maintain. Without the extra selectors, we get the desired effect (with perhaps the undesired effect of having button links underlined when you hover over them).

Let me know if y'all like it as is or if you want me to take out the `not` selectors.

cthoyt:
@erik-whiting that's a good point, I was hoping by using some selectors we could reduce the possibility of  strange styling popping up in unexpected places because we we accidentally overrode some bootstrap things. Thanks for giving some more info about the CSS - if you think it will work without them then I'm okay with removing them.

matentzn:
Please ping me again when ready to merge.

matentzn (approved):

matentzn:
Looks very good to me! THANK YOU @erik-whiting and @cthoyt!

PR #2170, opened 2022-10-28
Fix typo in preferred prefix T4FS

jamesaoverton:
@allysonlister Can you please double check this? Probably related to https://github.com/EBISPOT/OLS/issues/643


jamesaoverton:
@cthoyt should also be aware of this

cthoyt:
I would be happy to add a test that preferred prefixes are simply casing variants of the prefixes themselves, but https://obofoundry.org/ontology/dpo has some incompatible assumptions with this since it is `FBcv`

allysonlister:
Apologies for the delay, I was on annual leave. The modification fixes the problem, thank you!

matentzn (approved):

PR #2171, opened 2022-10-30
Add publication for OGMS

cthoyt:
This PR adds the OGMS publication.

@BAevermann can you please confirm that this is an appropriate publication? Do you have others to suggest? 

See also the epic on adding more publications to OBO Foundry in #1826.


BAevermann (approved):
This isn't a canonical ontology origin publication; however, OGMS does not have one. Given that, this could do. 

Thanks!

b.

matentzn (approved):

PR #2172, opened 2022-10-31
Ensure preferred prefixes are capitalization variants of ontology identifiers

cthoyt:
As a follow-up to #2170, this PR adds a test that enforces that preferred prefixes are capitalization variants of ontology identifiers.

There is exactly one exception to this rule, DPO (which is a subset of FBcv?) and duplicates another ontology's preferred prefix.

## Blocked By

- [x] #2170


matentzn (approved):
Great, thank  you!

jamesaoverton:
Thanks @cthoyt!

Issue #2167, opened 2022-10-27
Reduce redundant text in OBO ops landing and mission statment page


cthoyt:
The text on the following two pages is almost the same:

1. OBO Ops landing page: https://obofoundry.org/docs/OperationsCommittee.html
2. OBO Ops mission statement: https://obofoundry.org/docs/MissionStatement.html

I suggest taking the text out of the mission statement page and putting it on the landing page, then deleting the dedicated mission statement page. This would make the burden of reading through the site on the reader much smaller.

nlharris:
That sounds good to me! Should I do that (and the associated cleanup of menus or whatever is needed)?

cthoyt:
I would have just done it directly but I’ve got so much Luke warm feedback from that way of approaching things I figured I better just ask first. If you want to do it then great

cthoyt:
Tbh most of the pages about the ops committee are redundant so maybe you could even go further to collapse them together

nlharris:
Discussion from PR #2178:

@cthoyt: While you're compressing things, can you also just combine the Mailing List (https://obofoundry.org/docs/MailingLists.html) into the main page?
Main reason is the OBO Site is currently too big and disorganized to maintain. I am trying to fix this!

@nlharris: How about instead if I merge https://obofoundry.org/docs/CommitteeMeetings.html, https://obofoundry.org/docs/OBOCalendar.html, and https://obofoundry.org/docs/MailingLists.html into one page? (Not sure what to call it -- OBO Operations Operations??)
I understand the desire to tighten things up, but adding lots of details (such as mailing lists) to the main page doesn't necessarily make it easier for people to navigate.

@cthoyt: Sure that’s fine, call it logistics or something like that

PR #2180, opened 2022-11-02
new OBO Ops logistics page

nlharris:
merges content from https://obofoundry.org/docs/CommitteeMeetings.html, https://obofoundry.org/docs/OBOCalendar.html, and https://obofoundry.org/docs/MailingLists.html (and will therefore obsolete all of these pages; I will also remove links to them from the sidebar) addresses latest comment in #2167

### DIFF
--- a/docs/OBOOpsLogistics.md
+++ b/docs/OBOOpsLogistics.md
@@ -1,0 +1,43 @@
+---
+layout: doc
+id: OBOOpsLogistics
+title: OBO Operations Committee: Logistics
+---
+
+## Joining the OBO Operations Committee
+Please see the [SOP](https://obofoundry.org/docs/SOP.html#OPS_MEMBER)
+
+## Chairing an OBO Operations Committee meeting call
+Please see the [SOP](https://obofoundry.org/docs/SOP.html#OPS_CHAIR)
+
+## Meeting calendar
+
+The OBO Foundry calendar is publicly available to view at http://www.google.com/calendar/embed?src=cW8yM2Q3NDliajh0ZGwwb2VlNnRhdnBoc3NAZ3JvdXAuY2FsZW5kYXIuZ29vZ2xlLmNvbQ. Please see below for how to add it to your own calendar.
+Everybody can see it; however only a subset of people can edit it. Maintenance of the calendar is done on a per request basis, and only approved people have editing rights to prevent spamming.
+
+This calendar contains event such as
+(1) OBO meetings, for example the regular OBI developer call or the ad-hoc OGMS discussion on diagnosis
+(2) events of interest to the community: the NCBO webinars, or the ontolog sessions
+(3) meetings and workshops, such as ICBO, NCBO events etc. Deadlines would be acceptable (e.g. September 5th, deadline for tutorial proposal for some conference)
+
+Personal events are not appropriate for this calendar (e.g. Z on vacation) and adding those will result in editing rights being revoked.
+
+### Add the OBO Foundry calendar to your own
+
+The iCal URL of the OBO Foundry calendar is https://www.google.com/calendar/ical/qo23d749bj8tdl0oee6tavphss%40group.calendar.google.com/public/basic.ics
+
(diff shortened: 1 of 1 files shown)


docs/OBOOpsLogistics.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Warning regarding responding to invitations on the OBO calendar
+
+When replying to a calendar invitation **you need to accept/decline in your own copy of the event**. Declining an event in the shared calendar deletes it for everybody.
+
+## Mailing lists
+
+(All are restricted access and subscription, non-members may post (posts held for moderator approval))
+- [obo-operations-committee](https://groups.google.com/forum/?fromgroups#!forum/obo-operations-committee)
+- [Technical group](https://groups.google.com/forum/?fromgroups#!forum/obo-foundry-technical-working-group)
```
cthoyt, line 41: Can each of these working group-specific lists be put on the correspond working group page?

docs/OBOOpsLogistics.md
```diff
@@ -0,0 +1,43 @@
+---
+layout: doc
+id: OBOOpsLogistics
```
cthoyt, line 3: Does the ID do anything? I couldn’t find any places these are used.

nlharris: No idea! I copied the format from another page.

docs/OBOOpsLogistics.md
nlharris: Technical, editorial, and outreach working groups already have these mailing lists on their pages. I can add the OBO Ops one to https://obofoundry.org/docs/OperationsCommittee.html. Do you think they should then be removed from this page?

cthoyt: Yes! No need to duplicate

docs/OBOOpsLogistics.md
cthoyt: They don’t do anything. I can go through and remove them all so it’s one less thing people have to think about to maintain the site

matentzn: They dont somehow relate to the http URLs?

cthoyt: No. the URL is generated based on the filename, which is really terrible. A better way with jekyll is to assign a permalink (see https://jekyllrb.com/docs/permalinks/)

I was actually thinking about asking if we could do this, since it would allow us to rearrange more of the site to make sense

matentzn: ok, interesting! Is this though the right thing to focus our limited time on? I don't mind, whatever you think!

cthoyt: if you think there are other reasons nobody (besides Nomi and Darren) maintains the OBO site besides its immense complexity and lack of organization, I'm all ears

matentzn: lack of interest

nlharris: Nico is right. It's not like there are people out there thinking, "I am all primed to improve the OBO website! Oh, wait, no, it has these IDs I don't understand; never mind."

But if you want to spend your time simplifying it, Charlie, that's fine! I somewhat disagree, though, with your "don't repeat information in more than one place" philosophy here. I think sometimes it's useful to repeat information in multiple places that people might look (though of course that has to be weighed against the hassle of keeping it up to date in multiple places).

docs/OBOOpsLogistics.md
```diff
@@ -0,0 +1,42 @@
+---
+layout: doc
+title: OBO Operations Committee Logistics
```
nlharris, line 3: Note that this page needs to be added to the sidebar in place of these:

- [Meetings](https://obofoundry.org/docs/CommitteeMeetings.html)
- [Mailing Lists](https://obofoundry.org/docs/MailingLists.html)
- [OBO Calendar](https://obofoundry.org/docs/OBOCalendar.html)

I will do that in a separate PR

matentzn (approved):

PR #2186, opened 2022-11-08
Update fp-006-textual-definitions.md

nataled:
Replaces spermatocyte example and removes the (possible) implication that logical definitions are a requirement.


nataled:
@cmungall please check the example that replaced the spermatocyte example (used in the section touching upon mismatched text and logical definitions).

matentzn:
Oh, didnt realise you wanted specifically @cmungall to see this. Apart from the fact that he does not really dig the information modelling side of OBO, the general principle of the example is fine: you basically show how to align the `genus` of the `genus-differentiae` pattern, so its appropriate to convey the point..

nataled:
@matentzn He was requested because he objected to the previous example, and we wanted to ensure that the new example doesn't go astray.

Issue #2188, opened 2022-11-15
new logistics page does not work


matentzn:
https://obofoundry.org/docs/oboopslogistics.md

nlharris:
should be https://obofoundry.org/docs/oboopslogistics.html, but that doesn't exist either.
I'm adding an id line to the .md, which was thought not to matter but I'm guessing maybe it's needed.

PR #2192, opened 2022-11-15
added id

nlharris:
https://obofoundry.org/docs/oboopslogistics.html is not showing up. Maybe an id is needed for that .html to be created from the .md

### DIFF
--- a/docs/OBOOpsLogistics.md
+++ b/docs/OBOOpsLogistics.md
@@ -1,5 +1,6 @@
 ---
 layout: doc
+id: OBOOpsLogistics
 title: OBO Operations Committee Logistics
 ---
 


nlharris:
hopefully will fix #2188

matentzn (approved):

docs/OBOOpsLogistics.md
```diff
@@ -1,5 +1,6 @@
 ---
 layout: doc
+id: OBOOpsLogistics
```
matentzn, line 3: Didnt @cthoyt advice us that this was redundant?

nlharris: yes, he did. however, it's my best guess that it is needed for the html (which is currently missing) to be created. we'll see.

nlharris:
can we just merge it and see if it fixes the problem? if it doesn't, i can remove the id again, but i don't have any other ideas for fixing it. (it's not a capitalization or misspelling issue, as far as i can determine.)

docs/OBOOpsLogistics.md
cthoyt: You can already go to https://obofoundry.org/docs/OBOOpsLogistics.html on the live site. The ID is strictly not relevant for anything. The only issue was the typo in the table of contents.

nlharris: hmm, ok, i thought i'd tried going there and it didn't work. if that url works, then you're right, the id is not needed. i will close this PR.

nlharris: do i even have power to close this PR or does someone else need to do it?

nlharris: actually, the other file edit in this same PR is the one we want to keep. is there a way to just remove this edit from the PR? (sorry for my gh ignorance)

matentzn:
```suggestion
```

cthoyt (approved):

cthoyt:
@nlharris would you please update the description of this PR to include the magic message to close #2188? I don't know if it works when it's in the title

matentzn:
Thank you @nlharris and @cthoyt for addressing this issue 🚀

PR #2196, opened 2022-11-15
Remove extraneous metadata and unused pages

cthoyt:
This PR does two things:

1. Removes `id` from several documents. This field is does not hold any special meaning in Jekyll.
2. Removes three pages that are marked as deprecated, only have stub text, and are not linked to from any other page in the website


nlharris (approved):

matentzn:
Thank you @cthoyt :)

PR #2204, opened 2022-11-24
Update fideo.md

rabiaazzi:
The official repository of the FIDEO ontology is "https://gitub.u-bordeaux.fr/erias/fideo".


matentzn (approved):
It is easier for us as a community if everyone was using GitHub, but there is no rule against using something else! Thank you for your edits!

PR #2147, opened 2022-10-06
Create RoleNORManager.md

matentzn:

PR #2211, opened 2022-11-30
Fix yaml issue and add link to role-related SOPs

cthoyt:
Follow-up to #2147

### DIFF
--- a/docs/roles/RoleNORManager.md
+++ b/docs/roles/RoleNORManager.md
@@ -1,7 +1,7 @@
 ---
 layout: doc
 permalink: /roles/nor-manager
-title: OBO Role: New Ontology Request (NOR) Manager
+title: New Ontology Request (NOR) Manager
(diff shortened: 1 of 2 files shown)


docs/roles/RoleNORManager.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: doc
 permalink: /roles/nor-manager
-title: OBO Role: New Ontology Request (NOR) Manager
```
cthoyt, line 4: writing OBO in front of everything is redundant and a bit distracting, maybe better not to name roles OBO this and OBO that since it's implied that it's part of OBO

cthoyt: alternative is to put quotes around all of this to make it valid yaml

matentzn (approved):
Oh thank you for fixing this!

PR #2215, opened 2022-12-05
Update fbdv.md

dosumis:


deepakunni3 (approved):
Looks good to me. The added metadata are appropriate for this entry 👍

matentzn:
The outcome of the review is correct, but this is a tricky one that should be documented in a general, see my comments in #2236.

deepakunni3:
The provided usage is correct, links resolvable, and the example is appropriate for the fbdv ontology.

PR #2233, opened 2022-12-08
Add Erik role and Picture

erik-whiting:
Hello, just adding myself as requested


erik-whiting:
I do not understand why my commits always look like this

matentzn (approved):

PR #2235, opened 2022-12-08
Update maxo.md

LCCarmody:


erik-whiting:
@matentzn Correct me if I'm wrong, but this is Deepak's prerogative right? i.e., it's ok that there's not related issue to this and also it's not up to me to review? I'm going to go ahead and update the branch to be in sync with `master` and approve the CI run either way.

erik-whiting:
@deepakunni3 please note the CI failure: https://github.com/OBOFoundry/OBOFoundry.github.io/actions/runs/3667203381/jobs/6199508470
I have no idea what it means so I can't really offer the original dev any guidance, maybe you can

@LCCarmody I rebased your branch to be in sync with `master`, you will have to do this before working on it further:
```sh
$> git fetch --all
$> git branch backup-of-patch-1
$> git reset --hard origin/patch-1
```
^ Those commands create a backup of your local `patch-1` branch called `backup-of-patch-1` and then updates your local `patch-1` branch to the remote one here. This way your changes are up-to-date with the master branch, avoiding any merge conflicts once your PR is approved

matentzn:
> @matentzn Correct me if I'm wrong, but this is Deepak's prerogative right?

I made clarifying edits here: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2238/files

deepakunni3:
@erik-whiting Thanks for the heads up about the CI failure. Looks like it was due to the list of dependencies in maxo.md not being sorted alphabetically. 

Commit 68b160b should take care of this.

matentzn:
Next @deepakunni3 you should provide a positive review - and merge the change if you think its appropriate!

ontology/maxo.md
```diff
@@ -48,5 +48,3 @@ tags:
 tracker: https://github.com/monarch-initiative/MAxO/issues
 activity_status: active
 ---
```
deepakunni3, line 50: @LCCarmody It looks like the free text description was removed. This would result in a sparse page for MAXO at https://obofoundry.org/ontology/maxo.html

Ideally, we want to have a detailed description about the ontology in the free text.

LCCarmody: Would moving the 'description' to the free text (I think you mean the box below) suffice?

deepakunni3: Copying over the description from the metadata section to the main body of the markdown is definitely a good first step. Even better if you can provide more information about the ontology and its use (like the type of concepts defined, upper ontologies used, its application, projects that use the ontology, etc.).

matentzn: There should not be any redundancy between the description and the main body of the text, as both are displayed side by side. I would prefer if we keep this issue focused on the license change and not add additional complexity by adding a new request on top.

deepakunni3 (approved):
The updates introduced in this PR looks good and ready to be merged 👍

PR #2237, opened 2022-12-08
Strengthen metadata standards for roles

cthoyt:
This PR does the following:

1.  enforces that start dates are added for all people (adds placeholders for the many roles of Nico)
2. Renames `name` field to `github` for all people, not sure how this misannotation happened in a community of esteemed OBO curators :p 
3. Adds readable names to all people


erik-whiting (approved):
nice

Issue #1990, opened 2022-07-18
Some logos on ontology pages are broken: GO, SO [because of https]


cmungall:
https://obofoundry.org/ontology/go.html
https://obofoundry.org/ontology/so.html

I think the unifying factor here is that these logos are served from http not https, does this make sense @kltm?

kltm:
Works for me, I think?
That said, there are a lot of issues around mixed-schema web pages (https://support.mozilla.org/en-US/kb/mixed-content-blocking-firefox#w_what-is-mixed-content-and-what-are-the-risks) and different browsers are differently picky about this stuff.

matentzn:
My Chrome does not show these logos!

nlharris:
Those logos are broken for me in Chrome on Mac. @kltm not sure what you're suggesting as the fix.

kltm:
While it depends on the browser, the best overall way to fix would be to remove cross-schema issues and host from https.

nlharris:
Hmm. https://github.com/OBOFoundry/OBOFoundry.github.io/edit/master/ontology/go.md has
> depicted_by: http://geneontology.org/assets/go-logo-icon.mini.png

http://geneontology.org/assets/go-logo-icon.mini.png works, but https://geneontology.org/assets/go-logo-icon.mini.png does not.

Maybe I'm misunderstanding what you mean by "host from".

kltm:
For GO, we do not currently offer HTTPS from that location. Until we get more HTTPS in place (https://github.com/geneontology/go-site/issues/53#issuecomment-817047884), it would have to be hosted elsewhere.

nlharris:
I just ran into this broken-icon thing again and remembered that we already had a ticket for it. (Ok, I'm lying, I didn't remember, but I searched before starting to create a new ticket.)

Maybe we should just not attempt to show icons on ontology pages until this is fixed? Because it looks ugly, and makes the ontologies unfairly look abandoned.
<img width="496" alt="Screen Shot 2022-12-12 at 1 34 11 PM" src="https://user-images.githubusercontent.com/348704/207160814-bd8db68e-47a5-428e-83a4-63edf3970494.png">

matentzn:
@erik-whiting can you:

- [ ] Move GO and SO logos to https://github.com/OBOFoundry/OBOFoundry.github.io/tree/master/images
- [ ] Change the GO and SO metadata to link to these images like we do for CL

nlharris:
[...]

PR #2247, opened 2022-12-14
Use internally sourced logos for SO and GO

erik-whiting:
For https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1990

This PR puts the SO and GO logos into our `images` directory and displays them from there.


nlharris (approved):
The SO logo seems very teeny but that's probably ok

erik-whiting:
> The SO logo seems very teeny but that's probably ok

I grabbed this logo from the URL that it's currently being served from

matentzn (approved):

Issue #2048, opened 2022-08-25
Motion: Opening up OBO Operations Committee calls and meeting minutes


matentzn:
I hereby motion to

- open up the OBO Operations Call Meeting minutes in read-only mode to the public
- open up the OBO Operations Call itself for non-members to listen

At the last OBO Operations Call (23 August) I put forward this motion and it was generally received well. Before documenting the valid concerns, I provide details of the Implementation of the motion, as well as concrete items of motivation.

## Motivation

1. **_Recruitement._** The OBO Operations Committee has had a set of about a dozen (mostly) steady members of the last decade, but, short of a few direct recruitments, few new members. To maintain the current workload performed by our working groups (TWG, EWG) and the ontology review processes, we need new recruits, in particular, in the early career stages, to manage day to day activities such as improving metadata, website issues and outreach activities. I believe that opening up the meetings could provide a low-barrier entry for people that just want to see "what we are up to", rather than committing, from the onset, to all the duties assigned to full OBO Ops Members. The idea is that they listen, learn, spread the word, and, eventually, volunteer to help with our day to day business.
2. **_Transparancy_**. As an _Open Science_ organisation, we make a lot of important decisions during our calls, and many of our decisions are never documented publicly (as part of issues on GitHub etc), but simply recorded in the minutes. Arguments for and against admitting certain ontologies are not the only important examples - others include new member admissions, details about SOPs and metadata policies. While we try to make the decisions themselves public (not even that very well I think), we almost never publish the decision making process. I believe that an Open Science organisation, committed to openness, should be entirely transparent about all the decisions they do, and any community member should be able to obtain some understanding of what is being discussed.

## Implementation

- As past meetings were held under the assumption of confidentiality, I do not advocate to publish past meeting minutes. I would simply suggest to start a new agenda (we are overdue a new doc anyways), making it readable to the public and sharing it on the website.
- Obviously allowing community members to join meetings my induce communication overhead. I am open to implement a rule that prevents non-OBO Ops members to talk during a meeting (they can use the chat).
- Only OBO Ops members are allowed to vote on OBO Ops decisions.

## Concerns

- Members at the last call suggest that our meetings are already not exactly the most efficient, and adding more people could make it worse. I agree, therefore I would suggest to implement the above rule that prevents non-Members from speaking.

mellybelly:
[...]

PR #2253, opened 2023-01-09
add section on OFOC minutes being open to public

nlharris:
I'm adding this as a strawperson draft; please feel free to suggest edits!

### DIFF
--- a/docs/SOP.md
+++ b/docs/SOP.md
@@ -137,3 +137,9 @@ The current processes of nomination, acceptance and onboarding are described [he
 
 For a discussion on this SOP, see [here](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2043).
 
+## OBO Operations calls and meeting minutes are now open to the public
+In August 2022, in order to increase transparency and inclusivity and perhaps help recruit new members to the OBO Operations committee,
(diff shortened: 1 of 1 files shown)


nataled:
A few questions:
1) I'm unsure as to why this needs to be an SOP? It's really just information that (to me) belongs in some 'how do I participate' (or related) FAQ.

2) Do we normally include histories in the SOPs? Obviously we want to provide links, but why the other background information?

nlharris:
1. I should have noted, this was in response to the request in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2048#issuecomment-1328914773
2. In this case, I felt it was relevant to provide that (brief) history because I've seen so many cases where someone comes in later saying, "Wait, when/why did we agree to do this?". But if it doesn't seem needed, feel free to remove it.

nlharris:
Removed history of public meeting minutes decision and also removed statement that OBO Ops meetings are open to the public, as it was agreed during today's meeting that this is not the case. Now there's just one added line.

nataled (approved):

matentzn (approved):

nlharris:
Thanks. I'm not authorized to merge, so someone else will need to.

PR #2256, opened 2023-01-11
add text about ontology acceptance/rejection decision process

nlharris:
(as discussed in OBO Ops meeting on 2022-01-10)

also added a section about ontology rejection notification (with a placeholder asking whether a template should be added). if ontology rejection notification is not a thing, feel free to remove or edit that section as appropriate.

also added a mention of the NOR Manager (there may also be other places we want to mention them).

### DIFF
--- a/docs/SOP.md
+++ b/docs/SOP.md
@@ -18,19 +18,19 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry 
 
 ### New Ontology Requests (NOR)
 
-1. When receiving a new ontology request (NOR), the NOR Manager  should thank the submitter for their submission.
-1. The OBO dashboard administrator adds the new submission to the NOR dashboard, which is deployed at https://obofoundry.github.io/obo-nor.github.io/
-1. After the dashboard is run, the OBO dashboard administrator informs the submitter about the need to fix the issues revealed by the dashboard, noting this is not part of the review itself, just a precursor, and that upon completion, a liaison will be assigned.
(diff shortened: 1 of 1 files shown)


erik-whiting (approved):
Mad some minor changes to this PR (numbering, a couple of small typos, whitespace). I know there's an outstanding question in one of the issues but if y'all decide to go with the info in this PR instead, here's the code review approval 👍

nlharris:
This PR adds a new section about the NOR decision-making process (in the same region of the SOP as other stuff about NOR acceptance):

>## Ontology Acceptance/Rejection Decision
>Once the reviewer has decided whether a new ontology meets the requirements, they present their review and recommendation to the OBO Ops committee.
The committee can ask the reviewer for clarifications, and then the people on the call make the decision about whether to accept the ontology.
The decision is made by “consensus” on the call: i.e., no one objects strenuously.
No quorum (minimum number of attendees) on the call is required.

If someone wants to make a more general "OBO Foundry decision-making process" document, some of this section could be replaced by a link to that. But I think for now, this text is good enough to merge. Others may disagree.

docs/SOP.md
```diff
@@ -18,19 +18,19 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry
 
 ### New Ontology Requests (NOR)
 
-1. When receiving a new ontology request (NOR), the OBO dashboard administrator should thank the submitter for their submission.
```
cthoyt, line 21: @erik-whiting note that the repetition of `1.` is potentially nice since it reduces diff if there is ever an additional bullet added in the middle. Markdown is smart and knows to number appropriately

PR #2277, opened 2023-01-28
Add automated license correspondence check

### LINKED ISSUE #2276: Add automated check for licenses
Did someone carefully check the repository? It says there the data is licensed under CC BY 4.0-SA, which is not acceptable for OBO Foundry: https://github.com/Dalalghamdi/NGBO

_Originally posted by @cthoyt in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2214#issuecomment-1399241788_
            

What happened in NGBO is that the author had a different license in the metadata they submitted to the OBO Foundry than what the LICENSE file said inside their repository. Further, this wasn't carefully checked by reviewers. We should add an automated check for the future to make sure this doesn't happen again.

cthoyt:
Closes #2276

This PR adds a check for new ontologies that are currently in review to check that the license in the submitted metadata matches what's shown in GitHub. Implicitly, this now requires new ontologies have an appropriate LICENSE file.

### DIFF
--- a/src/obofoundry/utils.py
+++ b/src/obofoundry/utils.py
@@ -45,7 +45,7 @@ def query_wikidata(query: str):
 
 
 def get_new_data():
-    """Get records for ontologies that have additional checks
+    """Get records for ontologies that have additional checks.
 
     So far, this applies in the following scenarios:
 
--- a/src/obofoundry/utils.py
+++ b/src/obofoundry/utils.py
@@ -3,11 +3,12 @@
 import requests
 import yaml
 
-from obofoundry.constants import ONTOLOGY_DIRECTORY
+from obofoundry.constants import ONTOLOGY_DIRECTORY, ROOT
 
 __all__ = [
     "get_data",
     "query_wikidata",
+    "get_new_data",
 ]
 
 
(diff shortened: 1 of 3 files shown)


tests/test_integrity.py
```diff
@@ -2,6 +2,7 @@
 
 import json
 import unittest
+from functools import lru_cache
```
erik-whiting, line 5: did you end up using this somewhere? I don't see a `@lru_cache` annotation in the PR

src/obofoundry/utils.py
```diff
@@ -41,3 +42,20 @@ def query_wikidata(query: str):
     res.raise_for_status()
     res_json = res.json()
     return res_json["results"]["bindings"]
+
+
+def get_new_data():
+    """Get records for ontologies that have additional checks
```
erik-whiting, line 48: the Flake8 test wants you to put a period after "checks," can you please do that?

erik-whiting (changes requested):
This looks good, I like your approach of using the GitHub API. I noted two issues, both of which are super minor.

tests/test_integrity.py
cthoyt: I was going to cache GitHub API calls - but never got around to this. Somehow, flake8 didn't catch this! thanks for reading carefully, I removed it.

src/obofoundry/utils.py
cthoyt: addressed!

tests/test_integrity.py
cthoyt: In cc3b3620 I actually implemented this caching properly. This will be valuable for future additional tests that use GitHub data.

erik-whiting (approved):
Looks good to me, I see you tagged @deepakunni3 for review as well. Do you want to wait for him to take a look? If not, I think we're ok to merge this.

cthoyt:
@erik-whiting nope, I just tagged you both for posterity. Please merge when ready!

PR #2279, opened 2023-01-29
Automated check of passing NOR dashboard

cthoyt:
As a follow-up to #2277, this PR adds an automated check that new ontologies pass the NOR dashboard. Specifically, it checks that 1) the ontology is in the [NOR dashboard configuration](https://raw.githubusercontent.com/OBOFoundry/obo-nor.github.io/master/dashboard/dashboard-results.yml) and 2) that the summary status is `PASS` for all checks *except* for FP09 (usage)

### DIFF
--- a/.github/ISSUE_TEMPLATE/new-ontology.yml
+++ b/.github/ISSUE_TEMPLATE/new-ontology.yml
@@ -103,6 +103,14 @@ body:
     validations:
       required: true
   - type: input
+    id: contributing
+    attributes:
+      label: Contribution Guidelines
+      description: What is the URL of contribution guidelines? Ideally, this is a file name CONTRIBUTING or CONTRIBUTING.md in the root of your repository.
+      placeholder: e.g. https://github.com/geneontology/go-ontology/blob/master/CONTRIBUTING.md
+    validations:
+      required: true
+  - type: input
     id: download
     attributes:
       label: Ontology Download Link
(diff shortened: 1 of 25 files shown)


nlharris:
Thanks for doing this. My understanding is that new ontologies are allowed to FAIL the "usage" principle, though not any others. I suggested that the dashboard could change the "fail" for insufficient usage (pink) to a "warn" (yellow) so that we don't have to worry about making an exception for this one principle, but I did not prevail. So you may need to put in some logic about "all fields must be PASS except usage" to make your automated check match the current requirements.

cthoyt:
@nlharris i'm sorry you were unsuccessful with that, I agree totally... now this just means that the OBO Dashboard status is not useful...

matentzn:
The problem is that usage is required _full stop_, so we can't make it yellow. Its red. Any OBO ontology that wishes to pass the dashboard must pass that check!

The real question is if a usage must be recorded as a condition for registering a new ontology - here the vote was _no_. We could revisit this, but the threshold for what counts as a usage is pretty high, and its hard to see how more recent ontologies would be able to pass it (you need to provide a public link to a resources that resolves to a page mentioning an ID in your ontology). I can be swayed both ways. We could also relax for what counts as "evidence for usage" and make it required for NOR as well!

I have opened an issue about this here: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2281

cthoyt:
I've updated the look inside the results rather than use the top-level summary. The code is now allowing the usage check (FP09) to fail, while requiring all others do not.

erik-whiting (approved):

PR #2288, opened 2023-02-14
Added hard-coded pipe symbols to TOC in fp-011-locus-of-authority.md

nataled:
Hoping this will render properly. Can't be worse than what's public now!


cstoeckert (approved):
Markdown page in branch looks fine [here](https://github.com/OBOFoundry/OBOFoundry.github.io/blob/nataled-patch-8/principles/fp-011-locus-of-authority.md)

PR #2311, opened 2023-02-28
Use canonical OGMS publication

cmungall:
I think this is the canonical publication, the biomarkers one seemed secondary


deepakunni3:
@cmungall Thank you for updating the publication entry for OGMS.

CC'ing @BAevermann for review

ontology/ogms.md
```diff
@@ -18,8 +18,8 @@ preferredPrefix: OGMS
 products:
 - id: ogms.owl
 publications:
-- id: https://www.ncbi.nlm.nih.gov/pubmed/25991121
-  title: Biomarkers in the Ontology for General Medical Science
+- id: https://pubmed.ncbi.nlm.nih.gov/21347182/
```
cthoyt, line 21:
```suggestion
- id: https://www.ncbi.nlm.nih.gov/pubmed/21347182
```

cthoyt:
why not add this in additional, then use the `preferred: true` flag on the new one? is the old one completely irrelevant?

matentzn:
Good point Charlie

ontology/ogms.md
```diff
@@ -18,8 +18,8 @@ preferredPrefix: OGMS
 products:
 - id: ogms.owl
 publications:
-- id: https://www.ncbi.nlm.nih.gov/pubmed/25991121
```
cthoyt, line 21:
```suggestion
- id: https://www.ncbi.nlm.nih.gov/pubmed/25991121
  title: Biomarkers in the Ontology for General Medical Science
- id: https://www.ncbi.nlm.nih.gov/pubmed/21347182
  title: Toward an ontological treatment of disease and diagnosis
  preferred: true
```

cmungall:
Do we really want to add all relevant publications? For many ontologies
this will be huge. I think we should restrict this to a handful of papers
max that describe the entire ontology

cthoyt:
Yes, I am totally in favor of listing all publications that are about the resource itself (not necessarily all *relevant* publications, like "hey we used GO" is not in scope IMO)

If this gets out of hand, then we can use our best judgement, but here it's just 2

cmungall:
hmm, well in any case I think for this particular PR this reference is confusing to add since it refers to a proposed addition of biomarker terms to OGMS from 8 years ago, and none of these terms have since been added.

image

however, this should ultimately be decided by the OGMS developers which publications they feel more relevant to explicate their ontology

deepakunni3:
Looks like we are waiting on the contact person of OGMS to chime in on this PR.

CC'ing @BAevermann:

Would be great to get your inputs on the suggested changes.

Specifically, which paper would be more appropriate for OGMS:
- https://www.ncbi.nlm.nih.gov/pubmed/25991121
- https://www.ncbi.nlm.nih.gov/pubmed/21347182
- or both

And if you are not actively maintaining OGMS then please provide an alternative contact person.
@cthoyt has suggested @scheuerm .

deepakunni3 (approved):

PR #2321, opened 2023-03-12
Add missing ORCIDs for seveal deprecated/obsolete ontologies

cthoyt:
It seems to be the case that Thorsten Henrich's name was misspelled in many places, making it hard to identify this individual. This is now resolved.

I also was able to find the ORCID for Fahim Imam (related to obsolete NIF ontologies)

All of the updates in this PR refer to obsolete/orphaned ontologies and therefore I don't think needs confirmation from these people (though if anyone knows Fahim's GitHub handle, it would be nice to include)


cthoyt:
@deepakunni3 are there any comments/concerns about this?

PR #2343, opened 2023-04-17
Adds text regarding contributions guidelines for P20

nataled:
@cstoeckert please check the wording and verify that the text is appropriately placed for this principle. @cthoyt and @matentzn please verify that the added text captures the intent.

I started with P20 because the new text easily fits with the current text. @cstoeckert could you add some text to P8 as you suggested?

For P10 the new text will rather starkly stand out as addressing something different than the rest of the current text, so it will take more work to fit it in. I'm inclined to add it as an example, but this example is rather useless without the tracker, and that's covered in P20 (hence why I favored adding it there).

### DIFF
--- a/principles/fp-020-responsiveness.md
+++ b/principles/fp-020-responsiveness.md
@@ -15,7 +15,7 @@ Ontology development benefits from community input, which is strongly encouraged
 
 ## Recommendations and Requirements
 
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Fantastic, very nice @nataled!

cstoeckert (approved):

cstoeckert:
@nataled want me to add requested text as part of this PR or a new one?

nataled:
@cstoeckert a new one please. Thanks!

PR #2349, opened 2023-04-28
Update operations.yml

LK112019:
Could you please add me to the OBO Foundry site?
Thanks so much!

PR #2351, opened 2023-04-29
Lint operations YML file

cthoyt:
This PR makes no content changes, it just runs the lint command. Will reduce diff for other upcoming PRs such as #2349.

This makes a minor fix in the test to handle long names.


erik-whiting (approved):
🎈

PR #2357, opened 2023-05-02
Update TWG reporter

erik-whiting:
Switch Nico out for Ray as person giving TWG report


erik-whiting:
Identical PR submitted at the same time: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2358 decided to merge that one instead

PR #2363, opened 2023-05-09
Update fp-009-users.md to emphasize requirements for links

nataled:
Revised the document to draw attention to the need for links in usages declarations (this requirement was already present, but was spelled out in the examples). The examples as they were prior to revision were general and contained important information, so these were moved to the R & R section, while the Examples section was revised to give concrete examples.

### DIFF
--- a/principles/fp-009-users.md
+++ b/principles/fp-009-users.md
@@ -17,13 +17,37 @@ scientific area and does so in a usable and sustainable fashion.
 
 ## Recommendations and Requirements
 
-It is important to be able to illustrate usage outside of the immediate circle of ontology developers and stakeholders. Note that the ontology can still be listed on
-the OBO Foundry website while publicising your resource in appropriate channels and searching for users with needs you can meet.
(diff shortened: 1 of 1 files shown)


principles/fp-009-users.md
```diff
@@ -38,42 +62,56 @@ You may have multiple examples for each user, and mulitple users under the `usag
...
-  to warrant inclusion)
-
-- Documentation of requests from multiple users, e.g., via an issue
-  tracker (provide a link to the issue tracker online)
-
-- Use of the terms from the ontology to structure or annotate
-  experimental or derived data (e.g. GOA; provide details of how to
-  review the annotations)
+Use of PSI-MOD term IRI (in the form of a CURIE) in the PR ontology:
+```
```
cthoyt, line 66:
```suggestion
```yaml
```

would be great to use code highlighting in these examples :)

nataled: @cthoyt I'm afraid I don't know what you mean. Is it just to add 'yaml' ? What does that do?

cthoyt: yes, if you follow the triple ticks with the language name or data format name, the content will get beautifully highlighted :)

nataled: Done!

PR #2374, opened 2023-05-29
Add placeholder for missing fields in vario.md

deepakunni3:
This PR attempts at fixing a previously (prematurely) merged PR.

For Vario, the `repository` and `tracker` metadata fields were missing.
There doesn't seem to be a public repository where Vario is managed.

@maunov: Do you have a public repository URL where Vario is currently hosted?

@matentzn What is the course of action if there is no public repository URL for Vario?

### DIFF
--- a/ontology/vario.md
+++ b/ontology/vario.md
@@ -29,5 +29,7 @@ publications:
   title: 'Variation ontology: annotator guide'
 - id: https://www.ncbi.nlm.nih.gov/pubmed/25616435
   title: Types and effects of protein variations
+repository: ???
+tracker: ???
 activity_status: orphaned
 ---
--- a/ontology/vario.md
+++ b/ontology/vario.md
@@ -29,7 +29,7 @@ publications:
   title: 'Variation ontology: annotator guide'
 - id: https://www.ncbi.nlm.nih.gov/pubmed/25616435
   title: Types and effects of protein variations
-repository: ???
-tracker: ???
+repository: N/A
+tracker: http://variationontology.org/instructions.shtml
 activity_status: orphaned
 ---


matentzn:
Vario has a term request tool http://variationontology.org/instructions.shtml but this is not the same as having a tracker.

It's a priority to unbreak master branch now, so do that in whatever way necessary. I am surprised an "orphaned" ontology is checked for errors!

To fix this more properly, we need to contact @maunov again. In the worst case, we need to set the issue tracker to the request term URL, and the repository... I really don't know top of my head. We don't have a status for an "active ontology without an issue tracker"

deepakunni3:
Sorry about this. I should have been more careful.

I added your suggestion to the entry along with a `N/A` for the `repository` field. This not ideal but it at least unbreaks master.

I can make another PR for updating the repository and tracker fields and tag @maunov for feedback.

matentzn:
Sounds good :)

deepakunni3:
Would be great to get your review on this PR :)

matentzn (approved):
@cthoyt would shed some tears but let's merge it quickly and ask for forgiveness later..

cthoyt:
Yes, I am very upset about what's happened here for several reasons:

1. There should not be a technical way to merge a PR that is not passing QC. GitHub has a mechanism for this. This is not the first time this has happened and I still don't understand why we're not using protected branches properly. I'm happy to screen share with someone who has admin privileges on this repo to make it work properly.
2. Discussion should never be around subverting QC, but either for improving the broken content or updating the QC. There wasn't even a discussion about the failing tests due to 1)
3. There's no point in adding dummy values, this just means that the data is broken. This means anyone who writes scripts that consumes this data will have their scripts break. This degrades the trust in the OBO Foundry as a concept

Follow-up about the VariO experience: we NEED better ways of describing the status of repositories. This should include separate fields for "active/inactive" and "responsive/unresponsive" and we also need a flag to say when an ontology simply does not follow OBO Principles but we don't have a governance/reasonable social mechanism for booting repositories (and how bad does this have to be, for example, we aren't going to do this to HPO even though it explicitly doesn't follow OBO licensing principles, but VariO is a different level of bad)

right now, the tests assume that all non-inactive repositories have a tracker. This does not also apply to orphaned ontologies. It could be extended this way, but we never had this discussion...

deepakunni3:
I apologize for the poor merge and subsequent decisions on my part.
The intention to merge a failing PR was not out of malice.

I completely agree with your point @cthoyt about not being possible to merge a PR without QC passing.
I think somewhere I assumed that I would not be able to do this if the QC had not been all cleared (Which was pretty poor assumption on my part).
And I would like to push for the protected branches so that this does not happen again.

Once master was broken, my reaction was to fix the master branch which led to some questionable decisions.
Adding the dummy values was just to get master fixed which in hindsight was not the correct approach.

Adding a flag to state whether OBO ontologies follow OBO principles is definitely an important addition. And governance mechanism around this topic is vital.

The discussion on QC and extending the tests is definitely important and needs to happen. I created a ticket to start this discussion: https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2376

matentzn:
First and foremost - can someone explain to me how this PR:

https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2103/files

caused QC to fail? It seems to only make changes unrelated to repository etc, so why did it pass before, and fail after? `orphaned` ontologies can by definition not be considered `active`, so QC should not even run for this ontology.

The other issues we will discuss in the new open issues

PR #2384, opened 2023-06-07
Update actions

deepakunni3:
This PR does the following:
- tag the the metadata steward for review when creating a PR after 'Build derived files' action
- add conditional check to skip QC and tests if the actions are triggered from changes introduced by 'github-actions' bot

The latter should prevent actions from running immediately after 'Build derived files' action.

**Note:** The skip only applies when 'github-actions' bot commits files and creates a PR. All other user actions will still trigger the checks for QC and tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,5 +35,4 @@ jobs:
           title: 'Update OBO foundry metadata'
           body: |
             Recreates all OBO metadata.
-            Tagging @deepakunni3 for review.
           assignees: deepakunni3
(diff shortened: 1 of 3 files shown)


cthoyt:
@deepakunni3 is it necessary to have it tag you in addition to assigning you? like don't those both show up in your inbox the same?

WRT the other actions, this is fine.

deepakunni3:
Yes, but I **think** tagging me adds me as a reviewer to this PR. 
Assigning me just assigns the PR to me (without adding me as a reviewer).

deepakunni3:
@cthoyt Thanks for the sanity check. 
Yes, tagging in the body of the PR doesn't help.

matentzn (approved):
Interesting solution! I am fine with this!

Issue #2389, opened 2023-06-16
Bug in display of newsletter


matentzn:
@erik-whiting 

I merged https://github.com/OBOFoundry/OBOFoundry.github.io/commit/134993974a0abca719b5a830f4be9e5286b7ea9e

But it neither
- led to the menu showing the "Newsletter" section
- Nor any other viewable link of the newsletter

Can you figure out how to display this newsletter on the "blog" roll as @cthoyt called it?

erik-whiting:
@matentzn some notes from my initial investigation:
1. You called the menu item "Citation." That's why you don't see "Newsletter"
2. It doesn't look like anything in the `_posts` directory is linked anywhere in the repo. In fact, doing a global search of the repo for "/posts" only yields one result, the one you added.

I'm working on fixing this, but the point is: it doesn't seem like posts has ever even been used, so no one noticed it wasn't working. Or it was used, but discontinued so long ago, that it stopped working at some point.

erik-whiting:
Figured out the problem, see https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2390

cthoyt:
I can fix this later this weekend if Eric didn’t already

erik-whiting:
> I can fix this later this weekend if Erik didn’t already 

I fixed it, see https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2390

matentzn:
This definetly works: https://obofoundry.org/posts/2023-06-16-InauguralNewsletter.html

But I am assuming not in the way @cthoyt was envisioning the "jekyll blog roll".. In any case, it shows up for now, so I am happy :) Thanks a ton @erik-whiting

erik-whiting:
I didn't realize that's what you wanted, can you link me to the conversation where this was discussed so I can see what's being requested?

matentzn:
No worries you couldnt have seen this: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2388#issuecomment-1593113574

PR #2390, opened 2023-06-17
Fix menu item and link directory

erik-whiting:
fixes https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2389

I renamed the `_posts` directory to `posts`. The main reason I'm not worried about the impact of this change is because no files in the `_posts` directory were linked anywhere, so this seems like a safe change.

Also changed the added menu item to "Newsletter" rather than "Citation"


matentzn (changes requested):
If this is the correct solution, then just move the newsletter, the old blog posts should not really show up anymore.

matentzn (approved):

PR #2392, opened 2023-06-19
Add test that new ontologies can be converted into OBO Graph JSON with ROBOT

cthoyt:
This pull request adds a unit test that checks if an ontology can be converted from the required OWL format into the OBO Graph JSON format using `robot convert`. This presupposes that the OWL can be accessed using the standard PURL `http://purl.obolibrary.org/obo/<prefix>.owl`. This test is only applied to *new ontologies* and explicitly not to old ones.

## Why

Of the existing OBO Foundry ontologies, the ones that can't be readily converted to OBO Graph JSON have modeling issues like duplicate labels, malformed axioms, or other stuff that seems like it should be picked up by the OBO Dashboard, but currently isn't. This test implicitly will check that this isn't the case.

It's not a requirement that OBO Foundry ontologies are distributed in OBO Graph JSON, but many users want to consume ontologies in this pre-simplified format. This test makes sure that this is possible for new ontologies.

Further, it's possible to build applications for assessing the quality and working with ontologies in languages outside of Java if it's simple to convert out of the OWL format into a platform-agnostic format like JSON. My personal goal is to build on top of this PR to also include checks that CURIEs and IRIs used in the ontology are recognized by the Bioregistry, therefore pushing new ontologies to also more carefully consider standardization before being able to attain OBO Foundry status.


matentzn:
We should implement a "custom check" section in the obo dashboard, where we can implement a lot of these checks without requiring them to pass. I agree with you @balhoff this is not the right place for this test, and also not separation of concerns - but ok the other hand, obographs parse-ability is actually a good indicator for certain kinds of errors, and this can be used by the NOR dashboard reviewer to do some preliminary checking. The first thing I would request to be implemented would be "label uniqueness" - we could query ontobee looking for other terms in obo ontologies with the same label. 

If you agree @cthoyt i will work with Anita to add such a python check section in the obo dashboard.

cthoyt:
@matentzn can you instead focus efforts on developing a contributor guide for OBO Dashboard? There are so many checks I can imagine implementing (assuming I can write them in Python. I have zero interest in implementing anything in Java). I am sure many of them I will be more efficient to do on my own, but would have to spend the most time on communicating them, jockeying for which OBO Principle they fall under, etc. This is not how I want to spend my time, so having documentation and SOPs ready to go would be helpful.

Overall, the main question is: how can you reduce red tape for me to make code that improves the quality of new ontologies the best?

Assuming the NOR dashboard continues to work consistently and reliably for all new ontologies, then we can adjust the check we've already implemented that new ontologies have to pass the NOR dashboard to also include these other "custom" checks

I want to stress that these changes **only apply to new ontologies**. I think we have already agreed that the bar should be higher for new content, as we have consistently seen time after time in the last 2 years that we have indeed admitted a lot of garbage that we could have weeded out with more clever automation.

matentzn:
Removing "red tape" is not really possible, since this is an organisation, and we can't just make unilateral calls like this. 

You have two options: use your own dashboard system and implement whatever check you want, or make an issue and on the issue tracker saying:

"Request adding {obographs JSON writability} to principle {2}"

Once you have convinced the committee to add this to the principle text, then we can add this check to the dashboard.

But the process will have to be followed no matter what:

1. Issue first
2. Wait for feedback if ok
3. Make PR

Just skipping 1+2 will work for bug fixes, but not for feature requests. This is also frustrating for you, who took the time to write this check, just to be told we don't want it.

Just remember - if I wave your enthusiastic and highly appreciated contributions through without adhering to process, OFOC will lose trust in my ability to gatekeep and the organisation will either kick me out or others will start pulling out.

(These comments are independent of the concern of adding custom checks to the dashboard, which we should 100% do)

matentzn:
As per OFOC call 2023-06-27, there was no support to adding these kinds of resource intensives checks as part of OBO metadata. Its not out of the question to add this to the OBO Dashboard though, which will require a specific request here: https://github.com/OBOFoundry/OBO-Dashboard/issues.

Issue #2398, opened 2023-07-09
Urgent: Fix master branch


matentzn:
Something in the testing pipeline changed, which causes new Pull Requests to fail. The master branch is still green, but the truth is that its failing due to changes in the testing dependencies themselves.

We need to:

- Know why it is failing
- How we can prevent it failing moving forward
- Fix it

@erik-whiting I assigned you but I am not sure if you are @deepakunni3 are responsible for the ontology metadata _testing suite_.

deepakunni3:
I think this is in scope of the metadata registry steward.

Quick assessment from my side:

### Why tests are failing

The tests are failing because of Pydantic 2.0 (which was released last week).

When we define and use pydantic classes, the default behavior of pydantic has become more strict w.r.t. the typing of optional fields. Thus, you see validation errors when trying to work with pydantic classes that have optional fields but these optional fields weren't mark as `None` in the class itself. See https://docs.pydantic.dev/latest/usage/models/#required-fields for more information on this.

The quickest fix is to stick to pydantic 1.x

Alternatively, we can update the pydantic classes (and its usage) such that the validation errors are resolved.

### How to prevent it failing

I would recommend locking our dependencies in requirements.txt so that we can avoid scenarios where a newly released packages end up causing build and test failures.

deepakunni3:
Working on a fix for this... stay tuned

matentzn:
Excellent @deepakunni3 thank you! I am happy with your suggestion to freeze dependencies, but I could suggest to do something like https://github.com/INCATools/ontology-development-kit/blob/master/.github/workflows/constraints.yml which is an action that tries to automatically update constraints periodically.

deepakunni3:
I like the idea of using actions to update the requirements.txt 👍

nlharris:
Is this still urgent?

PR #2399, opened 2023-07-09
Fix for failing tests as a result of pydantic 2.0

deepakunni3:
This PR addresses the concerns raised in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2398

### DIFF
--- a/tests/test_memberships.py
+++ b/tests/test_memberships.py
@@ -21,8 +21,8 @@ class Affiliation(BaseModel):
     """Represents an affiliation."""
 
     name: str
-    ror: Optional[str]
-    wikidata: Optional[str]
+    ror: Optional[str] = None
+    wikidata: Optional[str] = None
 
 
 class Member(BaseModel):
--- a/tests/test_memberships.py
+++ b/tests/test_memberships.py
@@ -48,7 +48,7 @@ class TestMembershipData(unittest.TestCase):
 
     def test_data(self):
         """Test the working group data is clean."""
-        res = Group.parse_obj(yaml.safe_load(OPERATIONS_METADATA_PATH.read_text()))
+        res = Group.model_validate(yaml.safe_load(OPERATIONS_METADATA_PATH.read_text()))
         self.assertIsNotNone(res)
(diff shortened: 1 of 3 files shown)


tests/test_roles.py
```diff
@@ -48,7 +48,7 @@ def test_data(self):
             name = role_dict.get("name")
             with self.subTest(name=name):
                 self.assertIsNotNone(name)
-                obj = Role.parse_obj(role_dict)
+                obj = Role.model_validate(role_dict)
```
matentzn, line 51: I cant judge this change.

tests/test_memberships.py
```diff
@@ -48,7 +48,7 @@ class TestMembershipData(unittest.TestCase):
 
     def test_data(self):
         """Test the working group data is clean."""
-        res = Group.parse_obj(yaml.safe_load(OPERATIONS_METADATA_PATH.read_text()))
+        res = Group.model_validate(yaml.safe_load(OPERATIONS_METADATA_PATH.read_text()))
```
matentzn, line 51: I cant judge this change :)

matentzn (approved):
I cant judge one change in this PR, the rest looks non-controversial

tests/test_memberships.py
deepakunni3: Its just because `parse_obj` method was deprecated in Pydantic 2.0 and the recommended replacement is `model_validate`.
The functionality still remains the same - the method converts the incoming dictionary into the Group object.

tests/test_roles.py
deepakunni3: Its just because `parse_obj` method was deprecated in Pydantic 2.0 and the recommended replacement is `model_validate`.
The functionality still remains the same - the method converts the incoming dictionary into the Role object.

tests/test_memberships.py
matentzn: Thanks!

matentzn:
Nope I just don't like merging other peoples PRs, when I approve, you merge whenever

PR #2405, opened 2023-07-17
Check for triggering_actor instead of actor in github actions

deepakunni3:
This PR adds a check for triggering actor instead of github actor to prevent unnecessary runs when the derived files are created by github-actions bot.

### DIFF
--- a/.github/workflows/obo-test.yml
+++ b/.github/workflows/obo-test.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


deepakunni3:
Unfortunately there is no separate ticket for this.

This is an extension to a previous PR (https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2384) aiming at avoiding running QA and test actions each time the artifacts are derived.

matentzn:
All good! tHanks!

PR #2409, opened 2023-07-25
Adding IRIs verification in pre-registration checklist

pfabry:
Adding a checkbox in the checklist asking submitter that their ontology's IRIs follows the OBO Foundry Identifier Policy


matentzn (approved):
Excellent, thanks!

PR #2432, opened 2023-09-05
Update new-ontology.yml with labels & defs in English

### LINKED ISSUE #2304: Request clarification: Add English language requirements to NOR Review Process
During the call today (21st February 2023) we have raised the issue again of a non-English ontology submission. While it is a very unfortunate practical requirement, I think we should officially codify that there must at least an English language label present for each class in the ontology. If there are definitions, these should be provided in English as well.

There are some practical issues with validating multi-language ontologies. Lets not make this about that. I think if we want to seriously do our jobs reviewing an ontology, we should be able to understand it, and therefore I advocate the above.

nataled:
fixes #2304


matentzn (approved):
Great! Love it!

PR #2442, opened 2023-09-14
Update OntologiesReviewWorkflow.md to fix formatting

nataled:
Questions were supposed to each be on its own line


nlharris (approved):
Probably ok but may need to see on web page to determine for sure

nataled:
It worked, thanks!

PR #2452, opened 2023-09-26
Update fp-013-notification.md fix rogue html

nataled:


hoganwr:
The fix looks great to me! Ordering of close tags is corrected.

hoganwr (approved):
The fix looks great to me! Ordering of close tags is corrected.

PR #2458, opened 2023-10-08
Update omrse.md

hoganwr:
We are changing the name of the ontology, while keeping the namespace and acronym the same.


hoganwr:
Sorry for the multiple commits. We also updated the description and the GitHub URLs (moved repo from one organization to another in GitHub)

johnbeve:
@hoganwr @yongqunh Looks good to me for what it's worth, with the caveat that in the omrse.md file OMRSE still carries the "medicine" tag, though its scope is now much broader than medicine.

yongqunh:
@hoganwr, thanks for the request. 
I have reviewed this, and think that the name change makes sense. Such a name change is meant to make the OMRSE ontology have a broader scope, which will make it serve the broader community. What has done earlier with a focus on medice related social entity can be naturally expaned to cover more. Such a change is beneficial to the ontology and for the whole social entity research community as well. 

I also agree with @johnbeve that some corresponding changes, such as the "medicine" tag, need to be updated accordingly.

yongqunh (approved):
I have reviewed this, and think that the name change makes sense. Such a name change is meant to make the OMRSE ontology have a broader scope, which will make it serve the broader community. What has done earlier with a focus on medice related social entity can be naturally expaned to cover more. Such a change is beneficial to the ontology and for the whole social entity research community as well. 
Also noted that some corresponding changes (such as 'medicine' tag) may need to be updated accordingly.

deepakunni3:
@yongqunh Thank you for the review.

What tag would you recommend for OMRSE? 
Would "social" or "behavior" be more appropriate?

Also should the "domain" change as well?

yongqunh:
@deepakunni3 Yes. I think that "social" and/or "behavior" would be more appropriate.

yongqunh:
@deepakunni3 Thanks! Glad about the updating!

deepakunni3 (approved):

matentzn:
@deepakunni3 omrse should have a different `domain` since its scope is now generalised..

deepakunni3:
@matentzn Thanks for confirming. Yes, I couldn't find a domain in our current list of domains that fit OMRSE.
I think we will have to add something specific on the lines of: 'behavior' or 'social' or 'social behavior'

@hoganwr @yongqunh Thoughts?

PR #2465, opened 2023-10-18
Mark FBbi as inactive

cthoyt:
@wawong, the contact person for FBbi, has stated that the ontology is inactive and the people who are responsible for it have moved on. Reference: https://github.com/CRBS/Biological_Imaging_Methods_Ontology/issues/1#issuecomment-1769157042


deepakunni3 (approved):
Thank you @cthoyt !

PR #2467, opened 2023-10-19
Mark OGSF as inactive

cthoyt:
confirmed by @linikujp in https://github.com/linikujp/OGSF/issues/2#issuecomment-1771376195


deepakunni3 (approved):
Thank you @cthoyt !

PR #2480, opened 2023-11-28
Update eupath.md

zhengj2007:
Updated contact person email and set the ontology to inactive


cthoyt (approved):
All non-controversial, but requires @deepakunni3 to merge

deepakunni3 (approved):

Issue #2462, opened 2023-10-13
Should OBO context JSON-LD file contain a prefix `obo`?


cthoyt:
The OBO context JSON-LD file at http://purl.obolibrary.org/meta/obo_context.jsonld contains preferred prefixes for OBO Foundry ontologies with their corresponding URI prefixes. However, it does not contain the prefix `obo` (which would correspond to the top-level namespace http://purl.obolibrary.org/obo/).

This might be helpful for people using the `curies` package to compress other things living in the OBO namespace, such as GO_REF, UBPROP, and others.

matentzn:
@cthoyt can you give us an indication of how important this is to you? I am worried that about:

<img width="493" alt="image" src="https://github.com/OBOFoundry/OBOFoundry.github.io/assets/7070631/0c61b71d-56a6-4f2f-b25d-63d36591b88c">

(because of the problem that maybe not everyone knows they need to prefer the longest match)

I don't see yet a clear advantage for adding this here rather than just having this in bioregistry? The examples you give are all unsupported URL spaces.. (I am not trying to argue that the prefix isn't useful, just that it has to be added here)

cthoyt:
This is pretty low priority. I was just thinking about it based on a question from the `curies` package https://github.com/cthoyt/curies/issues/63 where some things get compressed to `obo:UBPROP_000006` but I realized when writing up the examples into unit tests that this wasn't actually possible if I start with the OBO Foundry converter.

I like your xkcd. It's another reminder that we should all be using standard code for handling prefix maps. We have a java and python implementation, I guess early 2024 I will be scheduled to write this in Javascript, R, and Rust if all goes well.

PR #2483, opened 2023-11-29
Improve JSON-LD context export

### LINKED ISSUE #2410: JSON-LD context does not expand as intended
https://purl.obolibrary.org/meta/obo_context.jsonld

We can't have trailing underscores on expansions (and have it behave as expected). Sorry, I don't make the rules

More context here:

- https://github.com/w3c/json-ld-syntax/issues/329

These would all have to be modified to be nested

```json
      "RO": {
         "@id": "http://purl.obolibrary.org/obo/RO_",
         "@prefix": true
      }
```

cthoyt:
Closes #2410 by using the `@id`/`@prefix` combination, which is required since JSON-LD technically doesn't support URI prefixes that end with an underscore `_`

cc @cmungall


matentzn:
@jamesaoverton do you know any specific reason not to do this? It was requested by @cmungall and it is the "correct" thing to do. My gut feeling is wave it through, roll back if something breaks, but if you already kNOw something will break we might as well fix it first.

balhoff (approved):
Thanks!

jamesaoverton:
This change is necessary, I guess, but quite annoying. We have a bunch of tools that predate the changes to JSON-LD that assume prefixes have to end with selected characters. We pushed back on that design decision but they ignored us.

ROBOT packages `obo_context.jsonld` and uses it for the default set of prefixes. I simply copied this file into ROBOT and that broke the unit test suite: <https://github.com/ontodev/robot/pull/1170>. I haven't looked any deeper, and I won't have time for at least a week. If someone else can step up, that would be great.

Since we have to make this change, and we'll have to fix our workflows anyway, I think that this PR should also add the `obo` prefix as requested in #2462.

jamesaoverton:
Sorry, I was wrong. Another 5 minutes of looking into this and I see that ROBOT does not just use the `obo_context.jsonld` but adds several common prefixes to it. When I restored those, the test suite passed again: https://github.com/ontodev/robot/pull/1170/commits/4f1af7e5030d6dbaff8a5bedfb54e6cedc574e54.

jamesaoverton:
Ok the upshot is that ROBOT will need a minor update to handle this change before the next release. Not a problem.

This reassures me that our workflows should not break as long as they are using JSON-LD libraries to work with the context file, and not doing it with custom code.

jamesaoverton (approved):
I tested with ROBOT and this looks good. Thanks!

Since we're already changing the context, this would be a good time to add the `obo` prefix, as requested in #2462.

cthoyt:
@jamesaoverton I am happy to add that in either this PR or a follow-up one. Maybe there are other reservations, though, based on the discussion in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2462

matentzn (approved):
Lets see what breaks!

PR #2495, opened 2023-12-19
DRAFT: Update SOP.md 'reactivating' section with timelines

nataled:
As per recommendation by Ops call 2023-11-14:
```
"Do we have a clearly documented process that governs how an ontology status is established? In particular, how these are marked unresponsive. We do have a process for an ontology to be marked responsive again.
There is a section in SOP on responsiveness: https://obofoundry.org/docs/SOP.html#REACTIVATION
This would have to be extended further
Add more clarification about change in activity status in the ‘Changing ontology metadata in the registry’ section"
```
It is unclear what "extended further" refers to. It seems to refer to adding 'unresponsive' to the list, but need verification that it has been ratified.


nlharris (approved):
These changes seem fine. I'm not sure where you're seeing "extended further" but that can be dealt with in another PR.

nataled:
The changes made were not the ones discussed (I think), so I want to verify with the group that they are okay. At the same time I'll update with the information that was actually requested.

nataled:
Closing since it looks like the changes are already public.

Issue #2499, opened 2023-12-27
Website updates: dashboard status indicator


matentzn:
in #2360 we have determined how to indicate OBO Dashboard status on the OBO Foundry front page table:

> Highlight passing ontologies favourably on the OBO Foundry front page: It was suggested that we can use both a color (maybe white for passing and gray for ontologies that failed) and add a badge to note that an ontology passed. Noone opposed the idea of having both.
Sort them to the top of each domain (default grouping by domain remains the same).
Nico will sort out the assignment of a suitable developer.

@erik-whiting has agreed to implement this.

Concretely I would suggest the following three interventions:

- [ ] Shade all rows on the front page https://obofoundry.org/ in a light, but visible, gray tone where the ontology is not passing the dashboard
- [ ] Sort ontologies to the top of each group where ontologies are passing the dashboard (sort order: first dashboard pass status, second alphabetical).
- [ ] See if we can somehow without creating an eye-sore add the OBO Foundry dashboard passing badge:

<img width="258" alt="image" src="https://github.com/OBOFoundry/OBOFoundry.github.io/assets/7070631/6cd75b1c-dedf-4687-8b4f-18f626216b36">

I think we may have to create a simplified version of this with just green and yellow (pass, or not) without the details on errors to fit. I think this can go underneath the license in the re-use column. @erik-whiting you can see on the ontology specific pages like https://obofoundry.org/ontology/bfo.html how to add these dynamically.

- [ ] I would like, optionally, to have a checkbox (off by default) "Hide Failing QC" which can be used to filter the view for only those ontologies that actually pass the test.
- [ ] Make the Dashboard badge we currently have on the ontology specific pages more prominent / central.

The central results are here:

https://github.com/OBOFoundry/obo-dash.github.io/blob/gh-pages/dashboard/dashboard-results.yml

matentzn:
@jsstevenson this is the corresponding issue.

nlharris:
Should we unassign Erik from this (and other issues)?

PR #2500, opened 2023-12-30
Dashboard status indicator

erik-whiting:
Working towards #2499 

@matentzn can you take a quick look at my first pass? Here's basically what I'm doing:

1. Parse the HTML of the OBO dashboard to find the pass/fail status of a specific ontology. (I know you said to translate the YML to JSON but as I mentioned, JavaScript doesn't work well with YML so translating it to JSON would present the same problem I had with just parsing the YML file, only with an extra step).
2. Use the parsed data to set a status indicator and success sort order value.
3. Use those values in the sortable table

With the _**current**_ code, it looks like this:

image

I don't actually want to show the "success sort order" column, it will be hidden eventually. My challenge right now is setting that to be the default sort value, but I cannot figure out how to do that since we are using a custom data table, I don't think there's a "default column to sort by" function. Do you have any tips for getting me started on that?

Also, let me know if the color doesn't work.

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -22,3 +22,11 @@ in the navigation menu
 .navbar a:hover {
   text-decoration: none;
 }
+
+/*
+Slightly darken failing ontologies
+*/
+
+.failing {
+  background-color: lightgray;
+}
--- a/assets/js/custom.js
+++ b/assets/js/custom.js
@@ -108,6 +108,9 @@ jQuery(document).ready(function() {
                             <th scope="col">
                                 <span>Social</span>
                             </th>
+                            <th scope="col">
+                                <span>Dashboard Status</span>
+                            </th>
                         </tr>
                     </thead>
                     <tbody>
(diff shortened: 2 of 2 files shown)


matentzn:
@erik-whiting Happy New Year, and thanks a ton for getting this going. 

Some thoughts:

-  Can you check if you can use this as your new data source instead of scraping the HTML? https://github.com/OBOFoundry/obo-dash.github.io/blob/gh-pages/dashboard/dashboard-results.json
- Instead of adding the checkbox for the dashboard in the "Dashboard status" column, can you try to see how it would look like to add the shield.io badge? see #2499 
- The grey looks great - some people may prefer a tiny bit lighter, but this is a detail we can postpone to the final review.
- The success order column should indeed be hidden, but I have no idea how to set "default sort orders"

matentzn:
@jsstevenson this is the draft PR, you can use or not as much of this as you like!

PR #2508, opened 2024-01-10
Update OntologyStatus.md to remove "proposed" and pretty the table

nataled:


nlharris (approved):

nataled:
@nlharris Well, the table thing didn't work. Looking into it now.

PR #2509, opened 2024-01-10
Update OntologyStatus.md the table thing again

nataled:


nlharris (approved):
i can't tell whether this will work better, but go ahead and merge and see...

PR #2510, opened 2024-01-10
Update OntologyStatus.md more table fun

nataled:


nlharris (approved):
the formatting stuff is worth a try. sorry it's such a pain!

PR #2513, opened 2024-01-12
add text about openness

### LINKED ISSUE #2048: Motion: Opening up OBO Operations Committee calls and meeting minutes

nlharris:
as suggested by Chris in https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2048#issuecomment-1863366238 fixes #2048

### DIFF
--- a/docs/OperationsCommittee.md
+++ b/docs/OperationsCommittee.md
@@ -5,6 +5,10 @@ title: Operations Committee
 
 [Members](Membership.html) of the OBO Foundry Operations Committee aim to improve the flow of operations and make things happen within the [OBO Foundry](http://obofoundry.org). Such operations include, but are not limited to, establishment of policies, review of resources, outreach and education.
 
+The OBO Operations commitee is committed to openness and accountability.
+In this spirit, we make the meeting notes from OBO Operation calls public (https://docs.google.com/document/d/1qhLFQL5IzTMUIBOtxJ5AqaEHcOCOQgNeXfvAb5O5P0A/edit).
+In addition, we strive to record decisions and action items as issues on the public [OBO GitHub tracker](https://github.com/OBOFoundry/OBOFoundry.github.io/issues) as much as possible.
+
 ## Organization of the OBO Foundry Operations Committee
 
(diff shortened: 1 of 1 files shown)


cthoyt (approved):

nlharris:
I'm not able to merge this

PR #2515, opened 2024-01-16
Add "unresponsive" to ontology status enum

### LINKED ISSUE #2255: Adding a flag "unresponsive" to OBO metadata
Problem: We (OFOC) have received several messages about ontologies that are actively maintained, but do not respond to issue tracker requests or pull requests _at all_ over the course of multiple months. This is an important piece of information for many ontology users, especially curators, that need to ensure that issues with the ontology and new term requests can be processed at all.

Proposal: The proposal is to introduce a new flag, `unresponsive: true` which documents that the fact that an ontology, while perhaps active, does not respond to issues, including bug reports and new term requests. It is valuable for prospective users to understand this before using an ontology. This flag could be used to display a small icon, like a red speech bubble icon or some such, on the main obo website. This flag would also inform the "responsiveness" check of the OBO dashboard, which currently only is checked by the mere presence of an issue tracker.

cthoyt:
Closes #2255

### DIFF
--- a/util/schema/registry_schema.json
+++ b/util/schema/registry_schema.json
@@ -7,12 +7,13 @@
     "activity_status": {
       "level": "error",
       "principle": "http://obofoundry.org/principles/fp-016-maintenance.html",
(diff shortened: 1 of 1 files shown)


nlharris (approved):

matentzn (approved):

deepakunni3 (approved):

deepakunni3:
Thanks @cthoyt ! Looks good 👍

PR #2523, opened 2024-02-06
Add clarifying remarks on running lexical matching to reviewer SOP

matentzn:
@nataled TWG will write instructions on how to run the tool and I will make a separate PR for that.


nataled:
Wondering if it would be better to have this done as part of the NOR Manager's responsibility? Also, should the lexical match program be specified here? (not needed if made NOR Manager's job)

matentzn:
@nataled good thinking - this will make running the pipelines more centralised and less error-prone, as different reviewers may have difficulties to run the code. I made the change, what do you think?

nataled (approved):

Issue #2459, opened 2023-10-09
Document how we should organise community contributions in ontology reviews


matentzn:
Contributions like https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2406#issuecomment-1751979404 are very valuable for our ontology reviews, but right now our SOPs have a gaping whole in them how to elicit and communicate these reviews. In my view, there should be a documentation section call "community input on ontology reviews" which contributors like Charlie can be pointed to, and a standard blurp that can be copy-pasted into the review that ensures the ontology submitters know which elements need to be fixed and which don't. The OBO Foundry assigned primary reviewer also needs a sentence in their SOP that clarifies that they need to pick and choose from community reviews which element are SHOULD, MUST and MUST NOT..

nataled:
EWG questions for @matentzn:
1) You mention the need for a new 'documentation section'. Which document?
2) How do we ensure that those who provide community input would read it?
3) Can you provide a draft standard blurb?

EWG questions for @pfabry:
1) It's unclear to me how these stages (technical, content) would be enforced.
2) You mention for content review there should be a separate channel for community input. Do you propose to have separate tracker items for each (official vs community)? If so, again, how would these be enforced?

Overall the EWG agrees with Paul's last bullet point and Nico's last sentence.

Regarding the SOP, we've identified a potential place for the new directive for the NOR reviewer (not Manager). We propose that it go between 5 and 6 below:

5. Once the new ontology passes the review, the NOR manager assigns the next available reviewer from the OBO operation members.
_**NEW: The assigned reviewer is considered the final arbiter of requirements, and in some cases might have to clarify which suggestions made by other reviewers are required, which are simply good to do but not required, and which should not be done.**_
6. Finally, when the ontology is fully accepted, the NOR manager remove the ontology from the OBO NOR dashboard.

Note: I'd like to think about the use of SHOULD, MUST, etc notation for these suggestions. So far we use them in principles, but I found during my own NOR review that more nuance was needed. In my case I used Consider,  Recommend, Strongly Recommend, and Imperative, defined below:

`CONSIDER means just what it says--think about it. Deciding against such suggestions will (probably) not adversely impact acceptance, but might impact users in terms of understanding and usage. RECOMMEND and STRONGLY RECOMMEND are changes that really should be done at some point (though not necessarily before acceptance). IMPERATIVE means that the ontology cannot be accepted until the change is made.`

[...]

PR #2529, opened 2024-02-28
Update RoleNORManager.md to include instructions regarding community …

nataled:
…suggestions

Added to the 'response to submitters' letter. Note: the original wording given here does not completely match recent communications with submitters, so there is a possibility that the additions given here must be reproduced elsewhere.

The additions were suggested [here](https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2459#issuecomment-1755910058) and approved by @pfabry in a followup comment.

### DIFF
--- a/docs/roles/RoleNORManager.md
+++ b/docs/roles/RoleNORManager.md
@@ -32,8 +32,8 @@ The New Ontology Request (NOR) Manager has these basic duties:
   Dear @GITHUBNAME,
   
   Thank you for your submission. The review will be executed as a two stage process. 
-  First, you will have to pass [OBO NOR Dashboard](https://obofoundry.org/obo-nor.github.io/dashboard/index.html). Pass means that no check _apart from `Users` may be red_.
-  After you have successfully passed the Dashboard you will be assigned an OBO Operations committee member to review the ontology.
+  - First, you will have to pass [OBO NOR Dashboard](https://obofoundry.org/obo-nor.github.io/dashboard/index.html). Pass means that no check _apart from `Users` may be red_.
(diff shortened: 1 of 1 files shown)


nlharris (approved):

nlharris:
@pfabry do you have time to take a look at this?

pfabry (approved):
I agree with the changes.

nataled:
@pfabry there is some question regarding the 'response to submitters' letter. I noticed that the original wording upon which my revisions were based does not completely match your recent communications with submitters. That raises the possibility that there is another version of the letter and, as a consequence, the additions given here must be reproduced elsewhere.

pfabry:
> @pfabry there is some question regarding the 'response to submitters' letter. I noticed that the original wording upon which my revisions were based does not completely match your recent communications with submitters. That raises the possibility that there is another version of the letter and, as a consequence, the additions given here must be reproduced elsewhere.

@nataled Just to be sure we are on the same page, I use [this](http://obofoundry.org/roles/nor-manager) template for the response to submitter. I understand that this is the template that will be modified by the present PR. Is the concern about the mention of the `versioning` flag?

nataled:
Hi @pfabry,

Yes, that's one of the differences between the role SOP and what has been used to communicate with submitters in practice. There are other differences as well. The differences were fairly minor, but it was hard to tell if there were two places the letter was kept, or if you've been modifying each time. Now that I know there is only a single place, I now need to know if the version in the role SOP should be updated with the version used in practice. If you think it should, I'll go ahead and make the revisions.

pfabry:
Hi @nataled  
I just checked in the OBO operation minutes and there is a decision to "Not make resolvable versionIRIs mandatory" for the NOR Dashboard dating back to January 9. I think this could indeed be added to the response to submitter template. Thank you!

nataled:
Will merge this and create a new PR to make other changes.

PR #2531, opened 2024-02-29
update mp.mp

sbello:
added depicted_by line pointing to new MP logo

### DIFF
--- a/ontology/mp.md
+++ b/ontology/mp.md
@@ -17,6 +17,7 @@ contact:
   github: sbello
   label: Sue Bello
   orcid: 0000-0003-4606-0597
+depicted_by: https://github.com/mgijax/mammalian-phenotype-ontology/blob/main/logo/2024_MP_logo_RGB_ICON_color.png
 description: Standard terms for annotating mammalian phenotypic data.
 domain: phenotype
 homepage: http://www.informatics.jax.org/searches/MP_form.shtml


matentzn (approved):
Hopefully our template scales the logo appropriately :)

sbello:
That was my concern as well. If it doesn't I can request a smaller scale version. I just need to know the dimensions.

sbello:
Instead of showing the logo I just see the word logo on the MP page (http://obofoundry.org/ontology/mp.html), so something is not working. Do I need to use a smaller version of the file? @matentzn Do you know what the dimensions of the image should be?

deepakunni3:
Ah, it looks like the URL used is pointing to the GitHub page with the PNG file instead of pointing to the actual PNG.

When referring to content like images, OWL files, zip, etc. you should make use of the raw GitHub URL.

`https://github.com/mgijax/mammalian-phenotype-ontology/blob/main/logo/2024_MP_logo_RGB_ICON_color.png` points to the page where the logo PNG file is located.

Whereas `https://raw.githubusercontent.com/mgijax/mammalian-phenotype-ontology/main/logo/2024_MP_logo_RGB_ICON_color.png` points to the actual PNG

@sbello I made a small PR here to fix this issue: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2538

PR #2551, opened 2024-03-19
Update operations.yml to remove inactive EWG members

nataled:


_data/operations.yml
```diff
@@ -46,7 +46,6 @@ members:
   country: USA
   github: wdduncan
   groups:
-  - editorial
```
nlharris, line 49: What does it mean to have someone on the Ops committee with no assigned groups? Don't all active Ops members have to belong to one or more groups?

nlharris (approved):

_data/operations.yml
nataled: That's my understanding but my memory is fuzzy on that. Heck, there are many members listed that never join meetings. Anyway, this is to address the EWG issue only. Anything beyond that is out of my scope and will need to be brought up at an Ops meeting.

nlharris: I'll add it to the agenda

anitacaron: This is why the PR is failing. If a person doesn't have a group assigned to him or her, you have to remove the `groups:` key, too.

nataled: Thank you @anitacaron !

nataled: @anitacaron Either I did something wrong, or there is something else causing the failure. I fixed the issue you mentioned but things are still failing. Perhaps I should just start over.

_data/operations.yml
```diff
@@ -66,8 +66,8 @@ members:
   country: USA
   github: bpeters42
   groups:
-  - editorial
-  link: https://www.lji.org/labs/peters/
+  - outreach
+ link: https://www.lji.org/labs/peters/
```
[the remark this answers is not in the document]
nataled: -_- so these files are sensitive to spaces except when I want them to be. Got it! 😄

Anyway, I just re-did the whole thing from the start, and will delete this branch when I figure out how.

PR #2555, opened 2024-03-22
feat: dashboard status indicator

### LINKED ISSUE #2499: Website updates: dashboard status indicator

jsstevenson:
close #2499 

taking over from @erik-whiting


matentzn:
This looks _wonderful_ @jsstevenson !! Fantastic!

1. Can you make a PR on https://github.com/OBOFoundry/OBO-Dashboard/blob/0af8ac6da43e6607c9a1eecb52a820d00438abca/util/lib.py#L543 to fix the dashboard badge to something reasonable
2. For now, can you remove the "OBO Dashboard" bard of the badge label to be just "QC"?
3. I prefer both badges in the re-use column for now but this is not my call. Can you generate both the screenshots again with the updated badge labels, add one screenshot entirely without a badge, and I will create a vote for OFOC to decide.

THANKS!

matentzn:
> In this case, I think it'd be nice to do some slimming on them to come closer to the width of the license badges.

Ok, agreed! Whatever you can do to make it look a bit more homogenous. Having the rows higher is not a problem!

jsstevenson:
Finally circling back -- implemented requests from 4/16 call, made some style tweaks, handle some broken links:

Screenshot 2024-05-27 at 3 45 17 PM

matentzn:
Awesome, I love it! We will run it by OFOC for the last time today and then - is it ready to be merged?

jsstevenson:
@matentzn almost -- I just need to plug in a hard link to the dashboard results JSON file (for some reason I couldn't find it yesterday, but checked today and it is... [right there](https://github.com/OBOFoundry/obo-dash.github.io/blob/gh-pages/dashboard/dashboard-results.json))

jsstevenson:
Two other notes (+ some example screenshots): badges can also say "Processing error: build dashboard" or "Failed to compute base" [presumably in cases where there's some error in the dashboard build process]. I have these badges linking to the base dashboard rather than the individual ontology dashboard page (since it doesn't exist when the build fails). That said, these badges end up widening the column significantly -- maybe they could be pared down on the obofoundry/obo-dashboard side in another issue?

Screenshot 2024-05-28 at 8 42 00 AM
Screenshot 2024-05-28 at 8 42 17 AM

And a final to-do for me: obsolete/inactive badges should be removed entirely (does that sound okay?)

Screenshot 2024-05-28 at 8 48 27 AM

matentzn:
I would say:

1. a single default qc badge we can store somewhere which says: `NA` for ontologies that dont have qc. being run
2. In OBO-Dashboard we should drop the "WARN" from the badge if an "ERROR" exists to make the badge shorter. Similarly the processing error text should be simply changed to FAILURE

GREAT work!

jsstevenson:
- Add a static badge (defined in `assets/json/`) for obsolete ontologies/anything that doesn't show up in the dashboard. I may end up needing to hotfix the exact URL since it the file won't be in a linkable location until this PR is merged but I think I got it right.
- Link output to dashboard JSON output

jsstevenson:
Screenshot 2024-06-02 at 2 09 45 PM
Screenshot 2024-06-02 at 2 10 02 PM
Screenshot 2024-06-02 at 2 11 21 PM

anitacaron (approved):

souzadevinicius (approved):
LGTM

matentzn (approved):
I trust @souzadevinicius and @anitacaron to review this more, so I would say: lets take it for a ride!

matentzn:
@jsstevenson as per convention, you get to merge!

PR #2571, opened 2024-04-09
Update of NOR Manager SOP

pfabry:


cmungall:
can you normalize on newlines so it's easier to see changes?

nlharris:
> can you normalize on newlines so it's easier to see changes?

Oh wow, yeah, what happened there that the entire document is shown as being changed? You think that's different flavors of newlines, Chris?

nlharris:
@pfabry since the way you did the edits make it look to GitHub like every single line has changed, can you (1) avoid doing editing that way in the future, and (2) explain what you actually did change so we can comment on that? Thanks!

matentzn:
I have recreated this PR here: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2631/files

PR #2578, opened 2024-04-17
Minimal update to Uberon metadata.

gouttegd:
This PR updates the metadata for Uberon to:

* update the checkout instruction (which still referred to the old SourceForge repository);
* add missing dependencies;
* remove the links to the individual import modules (`dependencies -> subset`), which no longer exist since Uberon adopted the ODK’s “merged imports” strategy;
* remove the `ext` product, which no longer exists (we still provide a `ext.owl` file to avoid breaking existing links, but it points to the canonical `uberon.owl` product);
* remove the MA and ZFA bridges from the product; it doesn't seem sensical to mention only two bridges -- either we list them all (all 44 of them) or we list none of them;
* add collected-metazoan as an official release product;
* update some documentation links.

### DIFF
--- a/ontology/uberon.md
+++ b/ontology/uberon.md
@@ -19,11 +19,10 @@ browsers:
   label: KnowledgeSpace
   url: https://knowledge-space.org/index.php/pages/view/UBERON:0000061
 build:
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Thank you @gouttegd looks good

PR #2595, opened 2024-05-07
sharedname.org no longer exists; link to web archive version instead

### LINKED ISSUE #2565: id-policy doc has a link to sharedname.org which has been domain squatted
https://obofoundry.org/id-policy

@alanruttenberg do you have any alternative links?

Shall we also mention bioregistry here cc @cthoyt

nlharris:
fix #2565


matentzn (approved):
Great!

PR #2597, opened 2024-05-14
Add Inactive NOR section to SOP

matentzn:


docs/SOP.md
```diff
@@ -62,6 +62,13 @@ Part of the NOR review process includes helping submitters navigate through sugg
 
 In summary, the designated NOR Reviewer is considered the final arbiter of requirements, and in some cases might have to clarify which suggestions made by other reviewers are required (MUST), which are simply good to do but not required (SHOULD), and which should not be done (MUST NOT).
 
+#### Inactive New Ontology Requests
+
+In case an open New Ontology Request has not been responded to by the NOR submit for more than 2 months, it is considered inactive. In this case the NOR reviewer should add the following message:
```
nlharris, line 67:
```suggestion
If an open New Ontology Request has not been responded to by the NOR submitter for more than 2 months, it is considered inactive. In this case the NOR reviewer should add the following message:
```

nlharris (approved):

nataled:
In general I too approve of the addition. However, there are a number of changes I'd like to make, including to other parts of the SOP. Rather than try to navigate additions upon additions, I'm merging and will submit a separate PR.

nataled:
One issue of concern is that the criteria for this 'inactive' status is precisely the same as 'rejected', but it was put in a different section of the SOP so that similarity wasn't readily visible.

PR #2598, opened 2024-05-14
Update SOP.md with various edits and rearrangments

nataled:


nlharris (approved):
Thanks, this is much improved. I made a few minor suggestions.

PR #2607, opened 2024-06-05
Add photo

janelomax:


matentzn (approved):
This if fine by me, in the future you may want to change this to something a bit more profile like?

https://obofoundry.org/docs/Membership.html

Issue #2611, opened 2024-06-20
Minor tweaks to last two columns in ontology table on home page


nlharris:
The new home page ontology table looks amazing! 🙌

I have some suggestions for the last two columns (discussed in Slack with @nataled):
- [ ] The second-to-last column is labeled "Re-Use", but it contains both the dashboard results and the license. Keep the license, remove the dashboard results, and relabel the column "License".
- [ ] Add a new column with the dashboard result badges, with the column header "QC"
- [ ] Remove the not-very-useful "Social" column (see https://github.com/OBOFoundry/OBOFoundry.github.io/issues/2408 where this was already requested)

@jsstevenson

nlharris:
Currently it looks like this:
![Screenshot 2024-06-20 at 11 45 17 AM](https://github.com/OBOFoundry/OBOFoundry.github.io/assets/348704/b219a50b-ec05-43bc-b9b5-436f70f877d3)

jsstevenson:
>  The second-to-last column is labeled "Re-Use", but it contains both the dashboard results and the license. Keep the license, remove the dashboard results, and relabel the column "License".

Might be less of an issue with the pending work over on the dashboard to slim down to badge width, but the original intent here was to avoid packing in too many columns. Agree that the current configuration is imperfect, though.

matentzn:
I personally dont agree with suggestions two (add column makes the table too wide) and three (removing the social column should be done independently of this question here). What about renaming the "Re-use" to "Interoperability" or "OBO Principles"? Both the Dashboard checks and the license pertain to that..

nlharris:
> add column makes the table too wide

Add one, remove one, relabel one! But, ok, fair point that removing the social column should be a separate decision.

> What about renaming the "Re-use" to "Interoperability" or "OBO Principles"?

PR #2618, opened 2024-07-08
Change column header

nlharris:
The second-to-last column of the ontology table is currently labeled "Re-Use", but it contains both the dashboard results and the license. I was in favor of separating those into separate columns, but there was concern that that would make the table too wide. The proposed compromise was to change the column header, as I have done. address #2611


jsstevenson (approved):

ddooley (approved):
Great!

PR #2592, opened 2024-05-03
Update po.md

cooperl09:
Fixing the Planteome logo

PR #2637, opened 2024-09-23
Link directly to image file for Planteome

balhoff:
See https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2592#issuecomment-2369656305


nlharris (approved):
looks like that should work, but i don't know if it will show up a reasonable size (i guess that is controlled by the obofoundry code)

deepakunni3:
@nlharris Yes, the image should render with a reasonable size :)

PR #2640, opened 2024-10-10
Update Slack link

nlharris:
to the one that actually lets you join w/o an invitation (hopefully this won't attract spam)

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -137,7 +137,7 @@
                                 Mail List</a>
                         </li>
                         <li>
-                            <a class="dropdown-item" href="https://obo-communitygroup.slack.com/archives/C01DP18L5GW">OBO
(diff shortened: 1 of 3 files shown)


nlharris:
anyone want to review this?

matentzn (commented):
If this is a permanent join link, won't this make the slack space vulnerable to Spam? 

As an aside; I didn't know you could create permanent invite links! I thought they expire after 30 days!

nlharris:
> If this is a permanent join link, won't this make the slack space vulnerable to Spam?

That is a risk. However, we have had the permanent join link on our website already for a year or so; this PR is just updating the link in the other various places it appears. So far, I have not been aware of any spammers joining the Slack. (If they do, we can ban them.)

> As an aside; I didn't know you could create permanent invite links! I thought they expire after 30 days!

matentzn (approved):
In this case, let's go do it.

PR #2655, opened 2024-11-25
fix: fix bioportal link generation and remove redundant links

jsstevenson:
Two issues 
* many Jekyll deployment tools don't support [custom filters](https://jekyllrb.com/docs/plugins/filters/). This is unfortunate because direct Liquid templating can otherwise get ugly fast -- but in our case it's fine, we only have a handful of exceptions and that list shouldn't get bigger
* several ontologies provided their own BioPortal links (oops, thanks @nataled )

This PR removes redundant metadata from the latter and performs a basic if/else "mapping" for edge cases directly in the layout rather than shipping it out to a custom filter.

### DIFF
--- a/_layouts/ontology_detail.html
+++ b/_layouts/ontology_detail.html
@@ -72,7 +72,7 @@ page.preferredPrefix | upcase %} {% if upperPrefix != page.preferredPrefix %}
         </a>
         {% assign skip_bioportal = false %}
         {% for browser in page.browsers %}
-          {% if browser.label.upcase == "BIOPORTAL" %}
+          {% if browser.label == "BioPortal" %}
             {% assign skip_bioportal = true %}
           {% endif %}
         {% endfor %}


jsstevenson:
Two other notes 

* If there's a notion of wanting to let the ontology submitters control their own metadata files (which makes sense to me), I could also undo the edits made to those files in this PR and have the template supply a BioPortal link only if the ontology page doesn't provide it already. I edited them directly to delete those links for the sake of removing redundancy, but I'd be happy to restore those changes if necessary.
* Some ontologies also provide BioPortal links in their free text descriptions. I felt a little less okay about altering those, so I left them as-is.

matentzn (commented):
I think its ok, but maybe better to not make edits to the specific metadata files without debate, as this decreases trust. 

I will wave this through immediately with you just drop the md file changes..

matentzn:
Is there a way to avoid changes to the metadata, but instead removing the code that parses the button from the md file?

_layouts/ontology_detail.html
```diff
@@ -70,9 +70,44 @@ <h1>
...
+            {% assign bioportal_id = 'FB-CV' %}
+          {% elsif bioportal_id == 'MOD' %}
+            {% assign bioportal_id = 'PSIMOD' %}
+          {% elsif bioportal_id == 'PSO' %}
+            {% assign bioportal_id = 'PLANTSO' %}
+          {% endif %}
+          <a href="https://bioportal.bioontology.org/ontologies/{{ bioportal_id }}" class="btn btn-outline-primary" target="_blank" rel="noopener">
+            BioPortal
+          </a>
+        {% endunless %}
```
matentzn, line 110: Wait the way you do it, if a bioportal link is manually set, wont they now see _no button_ at all?

jsstevenson: Don't believe so -- 

Lines 73 through 78 set a local `skip_bioportal` variable. If it's false, then lines 80-110 render a BioPortal button, but if it's true, then that's skipped. Regardless, 111-115  render custom buttons for each set of values defined in the ontology's front matter.

https://github.com/OBOFoundry/OBOFoundry.github.io/blob/c6e650d3012c8cbc0baca2a49ba7e9299f88ff4c/_layouts/ontology_detail.html#L111-L115

jsstevenson: eg here's my local render of PRO

Screenshot 2024-11-25 at 1 22 17 PM

matentzn (approved):
Nice! Thanks @jsstevenson!

PR #2659, opened 2024-12-02
Update pr.md products & descriptions

nataled:


ontology/pr.md
```diff
@@ -32,10 +32,16 @@ preferredPrefix: PR
...
+  description: PRO after reasoning has been applied, OWL format. Add '.gz' for compressed.
 - id: pr.obo
   title: pro_reasoned.obo
   description: PRO after reasoning has been applied, OBO format.
+- id: pr-asserted.owl
+  title: pro_nonreasoned.owl
+  description: PRO without reasoning applied, OWL format. Add '.gz' for compressed.
+- id: pr-asserted.obo
+  title: pro_nonreasoned.obo
+  description: PRO without reasoning applied, OBO format.
```
matentzn, line 44: What is the use case for asserted axioms? out of curiosity!

nataled: 'Asserted' here refers to the definitions, not axioms. In other words, for example, we assert a single parent, but other parents come via reasoning.

rays22 (approved):

deepakunni3 (approved):

PR #2665, opened 2024-12-16
Add JSON products to CL's description.

gouttegd:
This PR updates CL's description to add the JSON version of the main, -basic, and -base products.

It also removes the declared dependency to CHEBI, which is incorrect (CL does not import CHEBI), and adds instead a declared dependency OMO, which CL does import.

### DIFF
--- a/ontology/cl.md
+++ b/ontology/cl.md
@@ -11,11 +11,11 @@ contact:
 dependencies:
 - id: go
 - id: ncbitaxon
+- id: omo
 - id: pato
 - id: pr
 - id: ro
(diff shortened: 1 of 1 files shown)


deepakunni3:
@addiehl As the main point of contact of CL, could you please review and approve this PR?

matentzn (approved):
As a member of the wider CL tech team I feel fairly confident with these edits. No one knows how to spell obographs (https://github.com/geneontology/obographs/issues/84) but what I see here makes full sense.

Thanks @gouttegd

PR #2692, opened 2025-03-28
Update OBA usage example and publication

rays22:
This commit intends to
1. Update an example term from the Ontology of Biological Attributes that is used by the NHGRI-EBI GWAS Catalog.
2. Update the publication referencing the NHGRI-EBI GWAS Catalog.


matentzn (approved):

matentzn:
Thank you @rays22!

PR #2704, opened 2025-04-25
fix link to publications page to cite OBO Foundry

### LINKED ISSUE #2703: Missing page in FAQ > [How do I cite the OBO Foundry?](https://obofoundry.org/registry/publications.html)
Hi everybody,

I wanted to visit the page <https://obofoundry.org/registry/publications.html>, which is linked in the FAQ, but it is 404.

This is where it is referenced: https://github.com/OBOFoundry/OBOFoundry.github.io/blob/570eddd47010cd2bbb9f374750e405c6e2e9df04/faq/index.md?plain=1#L15

In the line below that, the referenced file is also not available. Should be online at <https://obofoundry.org/docs/Citation.html>, explaining "How do I cite a specific ontology?".

I guess, this is the right location now? <https://github.com/OBOFoundry/OBOFoundry.github.io/tree/570eddd47010cd2bbb9f374750e405c6e2e9df04/citation>

Thanks for fixing!

:wave: Susanne

nlharris:
fixes #2703

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -12,7 +12,7 @@ title: OBO Foundry Frequently Asked Questions
 
 ### Citation
 
-- [How do I cite the OBO Foundry?](/registry/publications.html)
+- [How do I cite the OBO Foundry?](/citation/RelatedPublications.html)
 - [How do I cite a specific ontology?](/docs/Citation.html)
 
 ### Terms


rays22 (approved):

jamesaoverton:
The test failure is unrelated to this change. It's an HTTP 429 Too Many Requests error, and probably just a temporary problem with GitHub. But the underlying reason is that the test is very inefficient. I created an issue: #2708.

nlharris:
When will it allow the checks to be re-run?

PR #2705, opened 2025-04-28
Update member status

rays22:
This commit intends to
1. Remove @rays22 as member of the OBO Operations Committee.
2. Add @rays22 to OBO Operations Alumni.


matentzn (approved):
Thank you @rays22!

Issue #2708, opened 2025-04-29
Test for existence of PURL config makes too many requests


jamesaoverton:
This test is failing with an HTTP 429 Too Many Reqeusts error:

https://github.com/OBOFoundry/OBOFoundry.github.io/actions/runs/14736689406/job/41364454604#step:5:54

Since the code is old, and it only started failing now, I expect it's a transient error. But this `test_has_purl_config()` function is very inefficient: <https://github.com/OBOFoundry/OBOFoundry.github.io/blob/master/tests/test_integrity.py#L217>. There has to be a better way to do this check without 200+ HTTP requests.

jsstevenson:
👍 I think I'd like to see (at least) two changes

1) we're basically testing for the existence of several specifically-named files in a subdirectory in the git tree. This could probably be accomplished by a single GitHub API call: https://stackoverflow.com/a/61656698

2) rather than failing as soon as a single discrepancy is encountered, it'd be more helpful to get a complete list of failures. IE collect a list of missing files and fail if it is of len > 0

deepakunni3:
Resolved via #2715

PR #2715, opened 2025-05-06
Use git trees to minimize the number of API calls in tests 

deepakunni3:
Use Git trees to minimize the number of API calls when checking for existence of PURL configuration.

Previously the check resulted in hundreds of API calls to https://github.com/OBOFoundry/purl.obolibrary.org

This PR simplifies the test into a single API call. Many thanks to @jamesaoverton and @jsstevenson for identifying the issue and giving the idea of using [Git trees](https://docs.github.com/en/rest/git/trees).

See #2708 for more context.


jamesaoverton (approved):
Thanks!

PR #2723, opened 2025-05-27
Update fp-019-term-stability.md

nataled:
Still trying to fix format issue.

### DIFF
--- a/principles/fp-019-term-stability.md
+++ b/principles/fp-019-term-stability.md
@@ -53,6 +53,7 @@ It is not necessary (and not advisable) to delete the textual definition.
    -  OBO: Use the `replaced_by:` tag with the value set to the CURIE of the relevant term
 2. Indicate any inexact term replacements:
    -  OWL: Use the `oboInOwl:consider` annotation property with the value set to the full IRI(s) of the relevant term(s)
+
 ```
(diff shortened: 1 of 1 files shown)


nataled:
@jamesaoverton if the HTML looks like the 'preview', then I do not like it. It removes a newline that, when missing, makes it more difficult to read.

principles/fp-019-term-stability.md
```diff
@@ -52,21 +52,15 @@ It is not necessary (and not advisable) to delete the textual definition.
...
-   -  OWL: Use the `oboInOwl:consider` annotation property with the value set to the full IRI(s) of the relevant term(s)
-```
-   <oboInOwl:consider rdf:resource="http://purl.obolibrary.org/obo/OBI_0001544">
-``` 
-
-   -  OBO: Use the `consider:` tag with the value set to the CURIE(s) of the relevant term(s)
-```
-   consider: OBI:0001544
-```
+   -  OWL: Use the `oboInOwl:consider` annotation property with the value set to the full IRI(s) of the relevant term(s): `<oboInOwl:consider rdf:resource="http://purl.obolibrary.org/obo/OBI_0001544">`
```
nataled, line 55: I'd like to first try the change I originally made. This change will make it harder to read.

PR #2725, opened 2025-05-30
reorder fields in a more logical way

nlharris:
why did these records start with affiliation, with name buried somewhere in the middle? that makes it much tricker to correctly make edits (adding or removing someone).

i reordered the fields (with the help of a Claude-generated script) to put name first and affiliation list, in hopes of making this file more maintainable both in the short term (i need to move some members to alumni) and longer term.

lmk if anything looks wonky; i only spot-checked.

### DIFF
--- a/_data/operations.yml
+++ b/_data/operations.yml
@@ -1,374 +1,374 @@
 members:
-- affiliation:
-    name: University at Buffalo, Buffalo
-    ror: 01y64my43
-  country: USA
-  github: alanruttenberg
-  groups:
-  - outreach
-  link: http://sciencecommons.org/about/whoweare/ruttenberg/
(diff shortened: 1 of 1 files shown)


nlharris:
i'm not sure why the tests failed - was there some sort of syntax error? (do extra spaces at the ends of lines matter?) or were the tests checking that the fields were ordered in the old way?

matentzn:
We have worked a long time to standardise the metadata records to make give as little and attack surface as possible (bugs etc). So we chose to just apply alpha numerical sort ordering for everything. I think its best we keep it that way, because implementing another sort order in the QC will be very cumbersome, and the YAML files are NOT for human consumption

PR #2726, opened 2025-05-30
Add description of how we vet requests to join obo-discuss

### LINKED ISSUE #2112: Add OBO Discuss mailing list management to SOP
In the context of the OBO Town Hall meeting today at ICBO, I made some notes about how the OBO Discuss mailing list is managed. Here is my view of what happens in practice today:

OBO Discuss is a Google Group: <https://groups.google.com/g/obo-discuss>. Anyone with an email address can ask to join OBO Discuss, but we have had trouble with spammers in the past, so the Google Group is configured to require new members to be approved by one of the mailing list owners. So I (and some other Operations members) filter the Pending Members list every few days. Checking the Pending Members was also part of the Operations chairing instructions, to remind us to do it every two weeks, at a minimum, but that's a long time.

When someone asks to join, they **must** include an email address and they **may** provide a reason that they want to join. If they provide a reason that's not gibberish, I just click "Accept". If they don't, I search Google for their email address. If I get good results for a person (often from a university or company), I click "Accept". If I don't get good results, I email the person and explain that everyone is welcome but we filter for spammers. I usually get a reply and then click "Accept".

In practice, 99% of join requests are accepted. I think I've rejecting one or two applications in the course of several years. 

This is an OK system, I guess, but it could be better. In any case, I think this should be part of the SOP document.

There are also other configuration options for Google Groups. We could automatically accept people, and ban them in the case of spam. I think that would be more problematic, in practice.

nlharris:
as described by James in #2112. I changed the 'I' to 'we', but it would be better if we replaced 'we' with the name of the role of the person who does the steps described (who apparently is currently James).

fixes #2112

### DIFF
--- a/docs/SOP.md
+++ b/docs/SOP.md
@@ -12,6 +12,7 @@ This document contains standard operating procedures (SOPs) for the OBO Foundry 
 - [Notification of NOR Decision](#NOD)
 - [Changing ontology metadata in the registry](#META)
 - [Reviving obsolete, orphaned, or inactive ontologies](#REACTIVATION)
+- [Adding new members to obo-discuss](#OBO_DISCUSS)
 - [Becoming a member of the OBO Operations Committee](#OPS_MEMBER)
 - [Chairing an OBO Operations Committee meeting call](#OPS_CHAIR)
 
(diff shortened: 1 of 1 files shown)


jamesaoverton (approved):

sebastianduesing:
@jamesaoverton I'll volunteer to be another person doing that job—I don't currently have many responsibilities within OBO Ops, so I'm happy to step up for this.

nlharris:
> I guess my small suggestions are:
> - set up a private Google Sheet to track new member requests and add it to the SOP
> - get more people to do this job, which would hopefully minimize how long it takes to approve a new member

Sounds good to me! Thanks for your work on this, and thank you, @sebastianduesing for volunteering to help!

nlharris:
I guess we decided to go ahead and merge this and maybe make future improvements (e.g. my suggestion above to name 'this job')?

jamesaoverton:
Yes.

I did not see a way to add any other message to users about the 'reason' field.

PR #2729, opened 2025-06-03
Create Development&StandardizationGuidelines.md

nataled:


hoganwr (approved):

docs/Development&StandardizationGuidelines.md
matentzn, line 1: not sure if it was wise to have a `&` symbol in the file name..

nlharris: agree, i would change that

PR #2731, opened 2025-06-03
Remove former OBO Ops members

nlharris:
(Moved to alumni.yml in separate PR)

I did this by hand; hope I didn't mess anything up!


nlharris:
The lint failed for a reason unrelated to my changes:
> Run actions/setup-python@v2
> Version 3.8 was not found in the local cache

nlharris:
Hmm, but #2732 passed, so maybe there is something wrong with this commit? The error messages are not informative.

PR #2732, opened 2025-06-03
Add former members

nlharris:
(Removed from operations.yml in separate PR #2731)

I did this by hand; hope I didn't mess anything up!


zhengj2007:
OK. I will do when your pull requests of operations.yml are merged. Thanks @nlharris

PR #2740, opened 2025-06-17
Patch 10

LK112019:
@nlharris the Bioportal image has been replaced


nlharris (commented):
This patch seems to have a lot of other changes besides the image - not sure what will change if this is merged.

nlharris:
I will close this PR unmerged and make a new one that just has the intended change to an image URL.

nlharris:
The newsletter already has a uPheno figure, so I don't think any of the changes in this PR are needed.

PR #2755, opened 2025-07-22
Remove docs/CompletedReviews.md

### LINKED ISSUE #2747: docs/CompletedReviews doesn't actually list any completed reviews
[docs/CompletedReviews](docs/CompletedReviews.md) ([web link](https://obofoundry.org/docs/CompletedReviews.html)), which purports to be a table of all OBO ontologies that have gone through the current review process, doesn't actually list any ontologies. We should probably either fill it out or get rid of the page. Depending which option others prefer, I can do either (with a slight preference for getting rid of it).

sebastianduesing:
Closes #2747. In the page removal policy discussion in 2025-07-22 OBO Ops call, it was agreed that this page could be removed. There are no references/links to it in other pages on the site, so nothing else needed to be modified.


jsstevenson (approved):
@sebastianduesing it looks like this originally failed because an HTTP request that fires as part of the test suite had failed. I re-ran the test and it passed, so I think it was probably a one-off issue.

sebastianduesing:
Interesting. Thanks, James!

PR #2765, opened 2025-09-02
Update disdriv.md

lschriml:
Updated home page to
https://disease-ontology.org/
from
http://disease-ontology.org/
as this did not resolve for Dashboard


deepakunni3 (changes requested):
I believe @nlharris is referring to the `http` link on Line 8, which also needs to be updated.

nlharris (approved):

PR #2784, opened 2025-10-19
Change ChEBI logo URL

CarMoreno:
Updated the ChEBI logo URL


deepakunni3:
I think it should render properly. See an example below with the SVG:
Screenshot 2025-10-20 at 09 22 31

jsstevenson:
It's actually a question of whether the browser supports it (anything from this decade will https://caniuse.com/svg)

nlharris:
Ok then, I approve!

PR #2802, opened 2025-11-20
Update contact information for Cell Ontology

dosumis:


matentzn (changes requested):
Is this agreed with @addiehl?

PR #2812, opened 2025-12-10
fix: ensure yaml examples are valid and conform to format expectations

### LINKED ISSUE #2811: Example YAML for principle 9 does not pass automated tests
For PR #2810 I copied one of the examples from <https://obofoundry.org/principles/fp-009-users.html#examples> and updated it. The automated tests failed because the indentation was wrong, a colon was used inside one of the descriptions, and the keys are not alphabetical. I had to check out the branch locally, set up a working environment, run `make test` and debug the output. What should have been a 2 minute "drive-by" commit took 30 minutes.

jsstevenson:
close #2811 

Ensure that YAML examples
* are valid YAML
* quote things that need to be quoted (I may have covered a few unnecessary cases, I don't remember the exact YAML rules here)
* provide keys in order (this is required for tests, I also don't remember how exhaustive the checks are so I just did it for every key)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1171,6 +1171,16 @@ ontologies:
   repository: https://github.com/OBOFoundry/COB
   title: Core Ontology for Biology and Biomedicine
   tracker: https://github.com/OBOFoundry/COB/issues
+  usages:
+  - description: Ontologies using COB terms
+    examples:
+    - description: List of ontologies using at least one COB term (See section entitled
+        "Which ontologies use it?")
+      url: http://dashboard.obofoundry.org/dashboard/cob/dashboard.html
+    - description: List of ontologies using the term 'organism' (See section entitled
+        "Ontologies that use the Class")
+      url: https://ontobee.org/ontology/COB?iri=http://purl.obolibrary.org/obo/COB_0000022
+    user: (multiple)
 - activity_status: active
   contact:
     email: entiminae@gmail.com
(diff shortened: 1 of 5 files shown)


jsstevenson:
@nataled sorry! I should've self-assigned the GH issue, I'm always quite lazy about that

> I do notice that the indentation you're using is different from the examples I'm looking at. Please double check.

Could you be more specific? If you're referring to something like

```yaml
- description: Protein Ontology
  examples:
    - description: OBO Format stanza showing use of PSI-MOD term in logical definition
      url: "https://proconsortium.org/app/export/obo/PR:000027653/"
  user: "http://purl.obolibrary.org/obo/pr"
```

vs

```yaml
- description: Protein Ontology
  examples:
  - description: OBO Format stanza showing use of PSI-MOD term in logical definition
    url: "https://proconsortium.org/app/export/obo/PR:000027653/"
  user: "http://purl.obolibrary.org/obo/pr"
```

these are interchangeable in YAML -- generally, the important thing is to be indenting by chunks of 2, it's sort of arbitrary stylistic preference whether you do either of the above



> I wonder if the incorrect version is a holdover from a different file type. We're referring to the file being edited as a YAML file, and the file itself says 'yaml' within it (though it is not visible externally), but the actual file is a .md file

Yeah, it's sort of... both? Jekyll (which is used to generate the static website) uses [yaml front matter](https://jekyllrb.com/docs/front-matter/) embedded within markdown files to declare page metadata in a computable form. I agree that it's not a "YAML file", strictly speaking -- it's just a YAML subsection. There are probably ways to effectively communicate this without burdening a new contributor too much.

PR #2825, opened 2026-01-15
Add uo.json

SmartMonkey-git:
A json version of the ontology is available at: https://raw.githubusercontent.com/bio-ontology-research-group/unit-ontology/master/uo.json

And tools like [Ontolius](https://github.com/P2GX/ontolius) depend on it


matentzn:
@deepakunni3 will do it when he gets to it!

PR #2876, opened 2026-03-30
Add action to ensure SOP for ontology activity related changes

matentzn:
This adds a review requesting changes whenever the activity status of an ontology is changed. I dont want to encode the entire SOP here, this is just a fail safe that can be easily dismissed by a TWG maintainer if neeed.

This is to avoid faux-pas like https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2874 (mea culpa) in the future.


deepakunni3 (approved):
Great addition @matentzn

PR #2891, opened 2026-04-07
Update fp-012-naming-conventions.md

nataled:


nlharris (approved):
looks good (much improvement over the existing version). my one suggestion is very minor.

PR #2892, opened 2026-04-07
Remove builds from all ontologies' metadata

cthoyt:
Closes #2882

This PR removes the `build` entry from all ontologies' metadata. I implemented a generic script `remove_field` that can be used to remove _any_ field. The work for this PR is done by invoking it like:

```console
$ python -m obofoundry.remove_field build
```

This PR also removes outdated documentation about the build field and references to it. I replaced a link to the build page in a 2016 blog post to an archive.org version of the page.


matentzn (commented):
Two questions:

1. Do we really need to keep the file src/obofoundry/remove_field.py? If we do, I think a note should be added to the contributors guiide - how to remove a field from the schema programmatically
2. Did you check that the build field is mentioned nowhere else, like validation and linting scripts?

cthoyt:
> 1. Do we really need to keep the file src/obofoundry/remove_field.py? If we do, I think a note should be added to the contributors guiide - how to remove a field from the schema programmatically

I can imagine using it in the future (i.e., for the `integration_server` field and other things returned by `util/check_schema.py`) I also like to add these scripts to version control for auditing purposes of the PR.

> 2. Did you check that the build field is mentioned nowhere else, like validation and linting scripts?

I ran the following command in the root of the repository to find all mentions of "build".

```console
$ grep -I -rn "build" . --color=auto --exclude-dir={.tox,build,.git,.idea,.ipynb_checkpoints,docs/build,dist,.mypy_cache}
```

here's what I found:

1. Documentation in readme about the defunct BBOP server (now removed)
2. A makefile target for the defunct BBOP server (removed)

otherwise, it's not mentioned anywhere else. I wasn't able to rebuild the TTL and NT files, but they should get rebuilt after this is accepted. I'm not sure if/how I need to manually rebuild the `registry/context.jsonld` file.

matentzn (approved):
@cthoyt this is _great_ thank you. Lets give the community 2 days to object, else this can be merged on Friday.

deepakunni3:
Thank you @cthoyt for the PR. 
And thank you @matentzn for reviewing and merging

PR #2894, opened 2026-04-08
Automate downloading Jena RIOT in Makefile

cthoyt:
This PR adds targets to the Makefile to automatically download Apache Jena RIOT, but only if it's not already configured and on the class path.

Without this, someone who clones the repository and runs `make` will get errors about `riot` not being on the classpath if they don't already have their environment configured like a semantic engineer.

Now, these kinds of people (me, lol) don't have to think about how to install dependencies like this!

This PR also adds to the `make clean` command to delete the compressed and uncompressed folders.

Testing help requested: can someone who has RIOT already configured on their path run this and make sure 1) RIOT doesn't get downloaded again and 2) the existing commands still work as intended


matentzn (approved):
I am fine with this change. Can be merged if no objections before Monday 25.

PR #2899, opened 2026-04-17
Remove several fields used less than five times

cthoyt:
As a follow-up to #2892, this PR removes the following fields, each of which was used less than five times:

- canonical (not needed because of convention)
- documentation (duplicated homepage)
- integration_server (irrelevant)
- label (duplicate of `title` field)
- releases (trivally computable from `repository` field)
- wasDerivedFrom (only in NCBITaxon)
- wikidata_template (only in UBERON)
- slack (only in UBERON)

This was done by running the script:

```console
$ python -m obofoundry.remove_field canonical documentation integration_server label releases wasDerivedFrom wikidata_template slack
```

If you want to see which fields are not frequently used, run:

```console
$ python utils/check_schema.py --max-cutoff 20
```

| key                |   count | ontologies                                                    |
|--------------------|---------|---------------------------------------------------------------|
| twitter            |      10 | go, hp, uberon, chebi, doid, ddpheno, po, cl, disdriv, ddanat |
| in_foundry         |      11 | obcs, bco, epo, zfs, omit, stato, bcgo, flu, fix, uo, lipro   |
| in_foundry_order   |      11 | pbpko, go, pr, obi, chebi, doid, pato, po, bfo, zfa, xao      |
| funded_by          |       4 | nomen, uberon, eco, kisao                                     |
| pull_request_added |       4 | clyh, t4fs, clao, ngbo                                        |
| issue_requested    |       3 | occo, t4fs, ngbo                                              |

PR #2901, opened 2026-04-20
Remove unused fields `in_foundry` and `in_foundry_order`

cthoyt:
Suggested in https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2899#issuecomment-4282678765

Note that the ontology_table.html template had some pretty weird bugs... the `<tr>` elements appear to get processed by the Javascript during runtime

this PR also finishes adding modern inline metadata for running scripts


matentzn (approved):
I fully agree with this but please lets get a @nataled review here before merging

nataled (approved):

nataled:
I approved, though my review is restricted to the ontology files, not the scripts.

matentzn:
@nataled I did the scripts - I only cared about the ontology files! Thanks

Thanks @cthoyt for the simplicifcation!

PR #2903, opened 2026-04-28
Update site with new Slack Signup form

sagehrke:
Updated the index.html and header.html with new slack signup form. This form was created after there were incidents of impersonators joining the OBO Community Slack (someone impersonated @matentzn). The form will allow us to review requests before sending a dedicated link for them to join the Slack. This workflow should help reduce unwanted impersonations/spammers.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -137,7 +137,7 @@
                                 Mail List</a>
                         </li>
                         <li>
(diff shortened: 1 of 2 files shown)


matentzn (approved):
This is fantastic! Thank you @sagehrke

nlharris:
Thanks, Sarah! I happen to be at my computer right now, so I'll merge!

PR #2909, opened 2026-05-20
Add several missing ORCiD identifiers

cthoyt:
# Reviews:

- [ ] @mariaheza -  Please check to confirm you are ok with the PR
- [ ] @burkesquires -  Please check to confirm you are ok with the PR
- [ ] @Onard -  Please check to confirm you are ok with the PR
- [ ] @mbrochhausen -  Please check to confirm you are ok with the PR
- [ ] @topalis -  Please check to confirm you are ok with the PR
- [ ] @DSchober -  Please check to confirm you are ok with the PR
- [ ] @mberriman -  Please check to confirm you are ok with the PR
- [ ] @slarson -  Please check to confirm you are ok with the PR
- [ ] @coulet -  Please check to confirm you are ok with the PR

The following only via email:

- [ ] Daniel Sonenshine -  Please check to confirm you are ok with the PR
- [ ] Satya S. Sahoo -  Please check to confirm you are ok with the PR  ⚠️ EMAIL bounced
- [ ] Christos (Kitsos) Louis -  Please check to confirm you are ok with the PR
- [ ] Julie Thompson -  Please check to confirm you are ok with the PR ⚠️ EMAIL bounced

### DIFF
--- a/ontology/tahe.md
+++ b/ontology/tahe.md
@@ -5,6 +5,7 @@ title: Terminology of Anatomy of Human Embryology
 contact:
   email: pierre.sprumont@unifr.ch
   label: Pierre Sprumont
+  orcid: 0000-0003-4933-045X
 domain: anatomy and development
 homepage: null
 is_obsolete: true
--- a/ontology/tahe.md
+++ b/ontology/tahe.md
@@ -5,7 +5,6 @@ title: Terminology of Anatomy of Human Embryology
 contact:
   email: pierre.sprumont@unifr.ch
   label: Pierre Sprumont
-  orcid: 0000-0003-4933-045X
 domain: anatomy and development
 homepage: null
 is_obsolete: true
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     container: obolibrary/odkfull:v1.2.30
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v7
         with:
           persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
           fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
--- a/.github/workflows/obo-test.yml
+++ b/.github/workflows/obo-test.yml
@@ -13,12 +13,12 @@ jobs:
     runs-on: ubuntu-latest
     if: github.triggering_actor != 'github-actions[bot]'
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v7
 
       - name: "Install uv"
         uses: "astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57" # v8.0.0
 
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v6
         with:
           python-version: '3.9'
       - name: Install requirements
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -11,9 +11,9 @@ jobs:
       matrix:
         python-version: [ "3.8", "3.10" ]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v7
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
       - name: Install dependencies
--- a/_config.yml
+++ b/_config.yml
@@ -2301,6 +2301,14 @@ ontologies:
     label: Viridiplantae
   title: Flora Phenotype Ontology
   tracker: https://github.com/flora-phenotype-ontology/flopoontology/issues
+  usages:
+  - description: FLOPO integrates qualitative plant trait data extracted from digitized
+      Floras across species and higher taxa of flowering plants.
+    examples:
+    - description: Browsing FLOPO and querying it with automated reasoning via the
+        AberOWL ontology repository.
+      url: http://aber-owl.net/ontology/FLOPO
+    user: https://www.kaust.edu.sa
 - activity_status: active
   contact:
     email: polcaes@gmail.com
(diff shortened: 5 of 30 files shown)


ontology/tahe.md
```diff
@@ -5,6 +5,7 @@ title: Terminology of Anatomy of Human Embryology
 contact:
   email: pierre.sprumont@unifr.ch
   label: Pierre Sprumont
+  orcid: 0000-0003-4933-045X
```
matentzn: Cant verify this.. How did you?

matentzn: In fact, since this person kept their orcid private, I think its not a good idea to add it here. Please remove.

matentzn (changes requested):
Almost looks good, one ORCID to remove please.

ontology/tahe.md
cthoyt: i did this late last night, I think I found a paper that annotated the ORCiD to confirm it's the same one. i'll remove it in this PR and can send another one when I dig up the evidence

cthoyt: addressed in 0153e3a6

matentzn (approved):
I am fine with this but please let's get a @nataled review to be sure this is procedurally kosher

cthoyt:
precedent for adding ORCiDs to inactive records without any fuss: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2321

nataled:
To my knowledge, we don't have an *official* procedure to cover this situation (meaning: however it was handled before was never codified). However, in February of 2022, the Operations Committee did make it a point to send a message to all ontology contact people about adding their ORCID, and each person could opt out. On that basis, I would say that's as close as we have to procedural precedent: We ask.

That being said, if these ontologies were part of the Foundry in early 2022 and there is no ORCID, then it likely means they've already opted out and they should NOT be added. The question I'd have is "do we have any indication that the responsible people already opted out?" If we do, then definitely don't add. If we don't, then ask.

Do we have a date for admittance to the Foundry for these ontologies?

matentzn:
>  if these ontologies were part of the Foundry in early 2022 and there is no ORCID, then it likely means they've already opted out and they should NOT be added.

@nataled did we receive any "opt-outs"? these need to be documented I think. I dont think we can assume they opted out - I myself have not heard of anyone opting out, and I would not even know through which channel that would be communicated.

All of these ontologies are much older, many inactive.

nataled:
Okay, good, we at least know that they were around back in 2022 and COULD have opted out if they wished. By the rule established in that email, failure to opt out means that the ORCID would have been added. That's the basis for the assumption they likely opted out, but there are other possibilities (see below).

I could not find any opt out cases, but the only search I did was to see if anyone responded to that particular email. People might have opted out via Slack if an announcement was made that way, or some other way. I would hope we have the opt out evidence. The email was sent by @nicolevasilevsky so maybe she has a list somewhere.

The fact that there is no ORCID for these can be due to multiple possibilities:
1) The relevant people opted out already;
2) The relevant people did not opt out, but did not have an ORCID at the time we did the mass add;

In the case of (1), hopefully we have a log of it and can put it to rest with a 'do not add'. If we don't have a log, we then have two paths. The first is to still assume they opted out and put it to rest with a 'do not add'. The second is to ask. The only downside to asking is them saying "I already answered this" and getting annoyed that we are asking again. I don't think that's such a terrible thing.

In the case of (2), we have to consider that they didn't opt out because it simply didn't apply to them when asked. Now that it would apply, we would need to ask.

So, overall, I think the default path forward is to ask. Only if we have concrete evidence they opted out already do we not bother because we already have the answer.

matentzn:
Its also so annoying to elicit permissions from this precise group because they dont even have a github handled registered. Sorry @cthoyt - your call but we have a clear rule here:

https://obofoundry.org/docs/SOP.html#META

1. First we have to tag with github handle (we cant in this case)

and then:

> If the change pertains to one of the following metadata fields, an OBO Foundry operations committee member MUST send an email to the listed contact with a link to the pull request: activity_status, contact.

matentzn:
Am I correct in assuming we dont have github handles for these people @cthoyt?

nlharris:
@cthoyt just making sure you saw the last question!

cthoyt:
I went through all additions to look up their GitHub accounts. Some weren't possible to find / don't exist. The rest are annotated now.

matentzn:
I updated the main comment of the PR to elicit reviews from everyone affected. I think we should put a timer on it though.

@nataled can you help me figure out if and how we can do this according to our current SOPs?

nataled:
@matentzn I think the email you sent (asking people to approve) is sufficient, as is the time period indicated. However, note that the responses will be opt-IN instead of opt out, which means that an unchecked box could mean either "I don't approve" or "I haven't seen this". Given that, the safest path forward would be to add the ones that opted in, and leave out those that didn't respond. I know that's not satisfying, but consider as well that the "I haven't seen this" case could mean that the contact information we have on record is no longer monitored/accessible by the person indicated.

matentzn:
Thanks @nataled - hmmmmkaaaaaay!

I guess if I wanted to challenge this I would have to come to an OFOC meeting and argue for changing the rules. You are clearly write on the interpretation. Lets see how it goes.

nataled:
@matentzn you were definitely in the discussions back then, so I assume that you were okay with the process (or got outvoted). What part would you wish to challenge?

matentzn:
> @matentzn you were definitely in the discussions back then, so I assume that you were okay with the process (or got outvoted). What part would you wish to challenge?

Yes I remember, and remember to also endorse the current SOP. Sorry what I meant to say: If I know _changed_ my mind and wanted to exempt contact information that can be obtained from the public record from opt-in (just opt-out), I'd have to request and SOP change.

nataled:
I did my best to represent Nico's new viewpoint (which I take to be that ORCIDs should be added by default without needed explicit consent) during today's Ops meeting. 

Pointed out:
1) Our NOR form already requires ORCIDs
2) NIH now requires them
3) Journals also request (require?) them

The vote was to require them (there were no objections). As for adding them without verification, there was just one concern: There are cases where multiple people have the same name, so we need to do our best to ensure that the ORCID we add is actually the correct one; that would be the main reason to reach out to the person, to confirm that it’s the right ORCID.

Thus, for cases where we CAN'T VERIFY using other information (for example, the ORCID entry shows the same email as the contact, or associated publications make it clear), we should still ask the person just to make sure. If that fails, we just take our chances (the sense was that it might not be such a big deal).

matentzn:
> I did my best to represent Nico's new viewpoint (which I take to be that ORCIDs should be added by default without needed explicit consent) during today's Ops meeting.

Wow I did not expect that, THANKS @nataled, that is super cool!

In this case, we will do this:

1. Do due diligence that the orcids @cthoyt has produced are 99% belonging to the person in our OBO records (I did this once and I do it again)
2. Keep the four weeks to let them intervene
3. Merge by 20th July if no one objects

Next time, we dont do that - we just go with the "orcid is required" line; do due dilligence; tag the authors for the record and merge.

@nataled will this be reflected in verbatim in the SOP text?

nataled:
I'm presuming that this is a one-time retrofit. Is that correct? The ORCID is already required for new ontologies (as reflected in the NOR form), so this shouldn't be an issue in the future.

matentzn:
> I'm presuming that this is a one-time retrofit. Is that correct? The ORCID is already required for new ontologies (as reflected in the NOR form), so this shouldn't be an issue in the future.

I think to some extent yes, but there might be a point where we need to record another item for identification which is required. So a sentence like:

OBO Foundry reserves the right to populate _required_ data elements from public sources without consent (OPT-OUT).

deepakunni3:
@matentzn It has been 4 weeks since your comment suggesting that we wait for 4 weeks to give people time to object: https://github.com/OBOFoundry/OBOFoundry.github.io/pull/2909#issuecomment-4783513448.

Is this PR okay to merge? Or are we waiting for additional checks or to send out notifications?

matentzn:
Thanks for the ping @deepakunni3 - merged!

PR #2912, opened 2026-05-23
Update ogms.md

zhengj2007:
Add usage

### DIFF
--- a/ontology/ogms.md
+++ b/ontology/ogms.md
@@ -33,10 +33,10 @@ tracker: https://github.com/OGMS/ogms/issues
 usages:
 - description: Multiple ontologies using OGMS terms, such as the Ontology for Biomedical Investigations (OBI), Relation Ontology (RO), Infectious Disease Ontology (IDO), Representation of Social Entities (OMRSE), and Mental Disease Ontology (MFOMD). OGMS has been extended to create various OBO ontologies, including the IDO and its extensions, as well as the OMRSE, MFOMD, Oral Health and Disease ontology (OHD), and Cardiovascular Disease Ontology (CVDO).
   examples:
-    - description: List of ontologies using at least one OGMS term
-      url: http://dashboard.obofoundry.org/dashboard/ogms/dashboard.html
-    - description: List of ontologies using the term "disease"
-      url: https://ontobee.org/ontology/OGMS?iri=http://purl.obolibrary.org/obo/OGMS_0000031
+  - description: List of ontologies using at least one OGMS term
+    url: http://dashboard.obofoundry.org/dashboard/ogms/dashboard.html
+  - description: List of ontologies using the term "disease"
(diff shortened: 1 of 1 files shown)


matentzn (dismissed):
This is not the intention of the usage field - if you want to show usage, you should show how the ontology is used by a resource, like a database, and provide a verifiable link to that resource. A simple owl import does not in and all by itself constitute usage. There are some fringe cases where you can make an argument like that (RO, COB, OMO, etc). Is there now online database using OGMS?

nataled:
I'm going to assume that the description is a roundabout way of saying that OGMS terms have been imported into the indicated ontologies, which would make the example valid. However, the key component 'url' is missing.

To fix: Replace 'type' (which doesn't appear in any example for [P9](https://obofoundry.org/principles/fp-009-users.html)) with 'url'. Then provide the IRI for an OGMS term used in one of the indicated ontologies. I'm not sure if multiple 'url' tags are allowed here but, if so, take advantage of that to provide links for each of the named ontologies. Alternatively, split the usages by ontology.

zhengj2007:
I intended to show that multiple ontologies import and utilize OGMS. I followed the BFO usage format as a template. 

I have replaced type with url.  Could you please review it again? @matentzn

deepakunni3:
@zhengj2007 Going by the Examples listed for [Principle 9](https://obofoundry.org/principles/fp-009-users.html), it would be better to represent the usage as follows:

```yaml
- description: OGMS has been extended to create various OBO ontologies, including the Infectious Disease Ontology (IDO) and its extensions, as well as the Ontology for Modeling and Representation of Social Entities (OMRSE), Mental Disease Ontology (MFOMD), Oral Health and Disease ontology (OHD), and Cardiovascular Disease Ontology (CVDO).
  examples:
    - description: List of ontologies using at least one OGMS term (See section entitled "Info: Which ontologies use it?")
      url: http://dashboard.obofoundry.org/dashboard/ogms/dashboard.html
    - description: List of ontologies using the term "symptom" (See section entitled "Ontologies that use the Class")
      url: https://ontobee.org/ontology/OGMS?iri=http://purl.obolibrary.org/obo/OGMS_0000020
  type: owl_import
  user: (multiple)
```

**Note:** I chose the term `OGMS:0000020 symptom` as an example. If there is more representative term then please feel free to use that instead.

zhengj2007:
@deepakunni3 Thank you for reviewing the pull request and recommending the changes. I have updated the usage and used 'disease' as an example.

deepakunni3 (approved):
Thank you for making the necessary changes 👍

zhengj2007:
@deepakunni3 Thanks for approving the request. The merging is still blocked. Shall I request @matentzn to review it again? Thanks!

deepakunni3:
@zhengj2007 I requested a review from @matentzn 
If I don't hear from him by tomorrow then I'll merge the PR

PR #2933, opened 2026-07-31
uo: update license to CC BY 4.0 and add usages

leechuck:
Two changes to the UO (Units of Measurement Ontology) registry entry. I maintain UO.

### License: CC BY 3.0 → CC BY 4.0

UO's licensing was inconsistent across its own repository: `LICENSE.md` was CC BY 4.0, `README.md` and the ODK config said CC BY 2.0, and this registry entry said CC BY 3.0. We have settled on **CC BY 4.0**, matching `LICENSE.md`, and made every in-repo statement agree.

The ontology header now declares `dcterms:license https://creativecommons.org/licenses/by/4.0/` (it previously declared no license at all, which is why the Open check was erroring). Since `fp_001` compares the ontology license against the registry license and errors on a mismatch, this entry has to move to 4.0 as well.

Companion PR on the ontology side: bio-ontology-research-group/unit-ontology#98

### Adding `usages` for FP09

Plurality of Users has been erroring with "Missing usages". The three users listed all demonstrably reference UO terms — verified by counting `UO:` references in each ontology's current release:

| User | UO references |
|---|---|
| HUPO-PSI mass spectrometry CV (`ms`) | 377 |
| Ontology for Biomedical Investigations (`obi`) | 12 |
| STATO (`stato`) | 6 |

All three also appear in the OBO Foundry's own "Which ontologies use it?" list on the UO dashboard.

### Validation

`util/validate-metadata.py` passes on the modified entry with zero errors and zero warnings, and produces a metadata grid byte-identical to the one from the unmodified entry.

Running the OBO Dashboard code locally against UO's next release with this registry entry takes FP01 Open and FP09 Plurality of Users from ERROR to PASS.

### DIFF
--- a/ontology/uo.md
+++ b/ontology/uo.md
@@ -11,8 +11,8 @@ description: Metrical units for use in conjunction with PATO
 domain: phenotype
 homepage: https://github.com/bio-ontology-research-group/unit-ontology
 license:
-  label: CC BY 3.0
-  url: http://creativecommons.org/licenses/by/3.0/
+  label: CC BY 4.0
+  url: https://creativecommons.org/licenses/by/4.0/
 preferredPrefix: UO
 products:
 - id: uo.owl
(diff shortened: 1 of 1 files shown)


ontology/uo.md
```diff
@@ -23,6 +23,19 @@ publications:
   title: 'The Units Ontology: a tool for integrating units of measurement in science'
 repository: https://github.com/bio-ontology-research-group/unit-ontology
 tracker: https://github.com/bio-ontology-research-group/unit-ontology/issues
+usages:
```
deepakunni3, line 26: @leechuck Since the goal is to highlight that UO has been used by many other OBO ontologies, another way of expressing this usage information would be as follows:

```suggestion
usages:
- description: UO has been used in various OBO ontologies, including the Ontology for Biomedical Investigations (OBI), The Statistical Methods Ontology (STATO), and HUGO-PSI Mass spectrometry ontology (MS).
  examples:
  - description: List of ontologies using at least one UO term (See section entitled "Info: Which ontologies use it?")
    url: http://dashboard.obofoundry.org/dashboard/uo/dashboard.html
  - description: List of ontologies using the term "length unit" (See section entitled "Ontologies that use the Class")
    url: https://ontobee.org/ontology/UO?iri=http://purl.obolibrary.org/obo/UO_0000001
  type: owl_import
  user: (multiple)
```

PR #2940, opened 2026-08-19
Update GitHub Actions

cthoyt:
This PR does the following:

1. Runs `pinact run --update` which pins each dependency in GitHub Actions to promote reproducibility
2. Adds minor spacing to make GitHub Actions configurations more readable
3. Update to running on modern Python 3.14

> [!WARNING]
> to accept this PR, the branch protection rules need to get updated, since there's no longer going to be a py310-tagged run

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,7 +35,7 @@ jobs:
           git commit -m "🚀 Automatically rebuilt derived files" -a
 
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
+        uses: peter-evans/create-pull-request@18f7dc018cc2cd597073088f7c7591b9d1c02672 # v3.14.0
         with:
           commit-message: Update OBO foundry metadata
           title: 'Update OBO foundry metadata'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,25 +15,27 @@ jobs:
     runs-on: ubuntu-latest
     container: obolibrary/odkfull:v1.2.30
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
           fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
 
       - name: "Install uv"
-        uses: "astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57" # v8.0.0
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
 
       - name: Create local changes
         run: |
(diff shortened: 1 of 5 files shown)


jsstevenson (approved):

jsstevenson:
I think branch rule management is restricted to [owners](https://github.com/orgs/OBOFoundry/people?query=role%3Aowner)... unfortunately I don't have permissions to alter that section of the repo settings

.github/workflows/build.yml
```diff
@@ -15,25 +15,27 @@ jobs:
...
+
       - name: Commit files
         run: |
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
           git commit -m "🚀 Automatically rebuilt derived files" -a
+
       - name: Create Pull Request
-        uses: peter-evans/create-pull-request@v3
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
```
matentzn: We can merge this, but this is not just a pin. This is a mega bump. If we merge this I expect that you will fix  subsequent issues with this - for example, I seem to remember we might need an explicit permissions block since v5. Your call!

.github/workflows/build.yml
```diff
@@ -15,25 +15,27 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v7
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
           fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
 
       - name: "Install uv"
-        uses: "astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57" # v8.0.0
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
```
matentzn, line 24: why do we have that if you use pip install in the next step?

.github/workflows/qa.yml
```diff
@@ -7,18 +7,18 @@ jobs:
...
-      - uses: actions/checkout@v7
-      - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up Python
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: ${{ matrix.python-version }}
+          python-version: 3.14
```
matentzn: Sould be a string, not a float, else a bump like 3.20 will fail. I know, cosmetics, buts still - good form

cthoyt: Fixed in 1c5017cf

.github/workflows/build.yml
cthoyt: reverted back to v3 series in 63e9ee57

.github/workflows/build.yml
cthoyt: it doesn't appear it's used directly used. I could remove it in this PR, but I think we would want to use it to modernize some of these scripts in the near future

PR #2942, opened 2026-08-19
Fix bug in sort ontologies script

cthoyt:
These conditions previously always evaluated to true because of a misunderstanding of Python syntax


jsstevenson (approved):

deepakunni3 (approved):
Great catch @cthoyt !

deepakunni3:
@matentzn Would you like to merge this or should I?

matentzn:
You go ahead! As long as some eyes are on these things no problem!

